PR #2, opened 2017-03-14
Fixed error in realloc in objects

dvirsky:
This bug does not allow (and will cause crash on) objects larger than 1M elements.


itamarhaber:
Thanks, merging :)

PR #7, opened 2017-03-29
Prevent float point serialization losing precision, fixes #6

### LINKED ISSUE #6: Float Point Precision
AFAIK, float point in redis is stored as 64 bit ieee float, but when I try to retrieve a float point value stored in a JSON field, it seems to lose precision, or is automatically rounded:

```
127.0.0.1:6380[7]> json.set world . 1.33333333
OK
127.0.0.1:6380[7]> json.get world .
"1.33333"
127.0.0.1:6380[7]> set hello 1.33333333
OK
127.0.0.1:6380[7]> get hello
"1.33333333"
```

Is this expected behavior?

evanhuang8:
Fixes the issue in serialization that falsely uses the `%g` flag and causes double precision float points to lose precision.

### DIFF
--- a/src/json_object.c
+++ b/src/json_object.c
@@ -339,7 +339,7 @@ inline static void _JSONSerialize_BeginValue(Node *n, void *ctx) {
                 else if (fabs(n->value.numval) < 1.0e-6 || fabs(n->value.numval) > 1.0e9)
                     b->buf = sdscatprintf(b->buf, "%e", n->value.numval);
                 else
-                    b->buf = sdscatprintf(b->buf, "%g", n->value.numval);
+                    b->buf = sdscatprintf(b->buf, "%lf", n->value.numval);
                 break;
             case N_STRING:
                 _JSONSerialize_StringValue(n, b);
--- a/src/json_object.c
+++ b/src/json_object.c
@@ -339,7 +339,7 @@ inline static void _JSONSerialize_BeginValue(Node *n, void *ctx) {
                 else if (fabs(n->value.numval) < 1.0e-6 || fabs(n->value.numval) > 1.0e9)
                     b->buf = sdscatprintf(b->buf, "%e", n->value.numval);
                 else
-                    b->buf = sdscatprintf(b->buf, "%lf", n->value.numval);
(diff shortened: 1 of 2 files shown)


itamarhaber:
Thank yo.

Currently this change makes one of the tests fail - https://github.com/RedisLabsModules/rejson/blob/master/test/pytest/test.py#L113 - for a test value of 3.14, that gets serialized as 3.1400000000000001

evanhuang8:
Right, this is a classic float point issue.

To preserve the precision, I have it output at least 17 significant digits using `%.17f`, which should ensure the correctness for 64bit IEEE float points.

In raw string however, it has the 1 at the end.

In python, if you parse float from that string, it actually outputs the correct value:

```python
Python 2.7.13 (default, Dec 18 2016, 07:03:39)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> float("3.1400000000000001")
3.14
>>> float("3.1400000000000001") == 3.14
True
>>>
```

evanhuang8:
I am not sure whether we want to modify the test to use a float conversion then check, otherwise if we purely compare strings, it will come off like that

itamarhaber:
The test sucked ;) Please use this instead, or something better:

```python
    def testSetRootWithJSONValuesShouldSucceed(self):
        """Test that the root of a JSON key can be set with any valid JSON"""
        with self.redis() as r:
            for v in ['string', 1, -2, 3.14, None, True, False, [], {}]:
                r.delete('test')
                j = json.dumps(v)
                self.assertOk(r.execute_command('JSON.SET', 'test', '.', j), v)
                self.assertExists(r, 'test')
                s = json.loads(r.execute_command('JSON.GET', 'test'))
                if type(v) is dict:
                    self.assertDictEqual(v, s, v)
                elif type(v) is list:
                    self.assertListEqual(v, s, v)
                else:
                    self.assertEqual(v, s, v)
```

itamarhaber:
So awesome that I can push to your PR ;)

Thanks for the fix!

PR #15, opened 2017-05-01
Fixes serialization to AOF

### LINKED ISSUE #14: Loading from AOF persistency silently breaks because of unescaped double quotes
Thus leading to values not being loaded

itamarhaber:
Removes enclosing redundant double quotes.

This resolves #14, but still leaves two uncovered topics:

1. AOF tests are missing
2. Need to capture and report the currently-silent AOF loading errors


itamarhaber:
WRT 1: https://github.com/RedisLabs/rmtest/issues/2

PR #19, opened 2017-05-10
Update README.md

mbusath:


itamarhaber:
@mbusath thanks for the contrib and welcome!

Issue #35, opened 2017-09-11
unicode support in rejson


wgjak47:
I'm trying to store a json with some chinese character in redis with rejson like this：
```
127.0.0.1:6379> JSON.SET test . '{"key":"测试"}'
OK
127.0.0.1:6379> JSON.GET test .key
"\u00e6\u00b5\u008b\u00e8\u00af\u0095"
127.0.0.1:6379> SET test2 "测试"
OK
127.0.0.1:6379> GET test2 
测试
127.0.0.1:6379> 
```
As you can see the value "测试" convert to "\u00e6\u00b5\u008b\u00e8\u00af\u0095" when using JSON.GET... It looks like rejson didn't support unicode yet...

dvirsky:
@wgjak47 the data should be encoded as UTF-8 IIRC.

wgjak47:
```
python
>>> '\u00e6\u00b5\u008b\u00e8\u00af\u0095'
'æµ\x8bè¯\x95'
>>> '\u00e6\u00b5\u008b\u00e8\u00af\u0095'.encode()
b'\xc3\xa6\xc2\xb5\xc2\x8b\xc3\xa8\xc2\xaf\xc2\x95'
>>> '\u00e6\u00b5\u008b\u00e8\u00af\u0095'.encode('ISO-8859-1')
b'\xe6\xb5\x8b\xe8\xaf\x95'
>>> _.decode('utf-8')
'测试'
```

It looks like the data is encoded with 'ISO-8859' not utf-8...

And I have tried to encoded the data with utf-8, but it doesn't work:
```
>>> result = client.execute_command("JSON.GET", "test", ".", encoding="ISO-8895-1")
>>> result
b'{"key":"\\u00e6\\u00b5\\u008b\\u00e8\\u00af\\u0095"}'
>>> result = client.execute_command("JSON.SET", "test", ".", a, encoding="ISO-8895-1")
>>> result = client.execute_command("JSON.GET", "test", ".", encoding="ISO-8895-1")
>>> result
b'{"key":"\\u00e6\\u00b5\\u008b\\u00e8\\u00af\\u0095"}'
>>> result.decode()
'{"key":"\\u00e6\\u00b5\\u008b\\u00e8\\u00af\\u0095"}'
```

Any example?

thienit5:
I have the same problem, can you show me how to decode?

dvirsky:
@wgjak47 @thienit5 
```py
[...]

PR #37, opened 2017-10-07
add simple unicode support to show unicode string proprely when use JOSN.GET

wgjak47:
#35
Show unicode  properly instead of hex. For example:
```
127.0.0.1:6379>  JSON.SET a . '{"test":"测试"}'
OK
127.0.0.1:6379> JSON.GET a
{"test":"测试"}
```
instead of
```
127.0.0.1:6379> JSON.SET test . '{"test":"测试"}'
OK
127.0.0.1:6379> JSON.GET test .
{test:"\u00e6\u00b5\u008b\u00e8\u00af\u0095"}
```
And I also add a simple unicode test case...

### DIFF
--- a/src/json_object.c
+++ b/src/json_object.c
@@ -345,7 +345,7 @@ inline static void _JSONSerialize_StringValue(Node *n, void *ctx) {
                 if ((unsigned char)*p > 31 && isprint(*p))
                     b->buf = sdscatprintf(b->buf, "%c", *p);
                 // for unicode
-                else if (_get_unicode(p, &step, len) == 0)
+                else if (_get_unicode(p, &step, len))
                   b->buf = sdscatlen(b->buf, p, step);
                 else
                     b->buf = sdscatprintf(b->buf, "\\u%04x", (unsigned char)*p);
--- a/src/json_object.c
+++ b/src/json_object.c
@@ -265,6 +265,49 @@ typedef struct {
     if (b->indent)               \
         for (int i = 0; i < b->depth; i++) b->buf = sdscatsds(b->buf, b->indentstr);
 
+inline static int _get_unicode(const char *p, int *step, int length) {
+  int unicode_char_length = 0;
+  unsigned char c = (unsigned char) *p;
+  // 0x04000000 - 0x7FFFFFFF:
+  if ((c >> 1) == 0x7E)
+    unicode_char_length = 6;
+  // 0x00200000 - 0x03FFFFFF:
+  else if ((c >> 2) == 0x3E)
+    unicode_char_length = 5;
+  // 0x00010000 - 0x001FFFFF:
+  else if ((c >> 3) == 0x1E)
+    unicode_char_length = 4;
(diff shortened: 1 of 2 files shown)


wgjak47:
strange... Docker build failed on ` pip install redis rmtest ramp-packer` in Travis CI...
I have no problem with building image on my own computer...

mnunberg:
Technically speaking, there is never any need to encode JSON as \uXXXX, when in reality you can just encode the characters as-is. JSON does actually not support encoding random binary data, so if the data is not valid UTF-8, it shouldn't be 'escaped', but should be handled at the user level. A proper JSON parser should explode when passing "binary" data in any event:

```
>>> json.loads('["\\0fff\\u0000"]')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python/2.7.12_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/__init__.py", line 339, in loads
    return _default_decoder.decode(s)
  File "/usr/local/Cellar/python/2.7.12_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 364, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/Cellar/python/2.7.12_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 380, in raw_decode
    obj, end = self.scan_once(s, idx)
ValueError: Invalid \escape: line 1 column 3 (char 2)
>>> 
```
If possible, I'd be more in favor of just never returning '\uXXXX' unless requested anyway

mnunberg:
However the code is still useful in cases where someone wants to see console output, or in environments where 7-bit safety is required, but I'm wondering if that shouldn't just be a separate command.

wgjak47:
So may be a new option like [FORMAT "raw"/"hex"](default is hex) is a better idea...

dvirsky:
@mnunberg @wgjak47 I think an optional  hex escaping argument is the way to go. something like `UNIESCAPE` or `ESCAPE`. The question is the default - with or without escaping? I would say that while most users will expect no escaping by default, since it's already escaping at the moment, let's not break anyone's app. so `NOESCAPE` means returning raw unicode?

mnunberg:
raw unicode, or whatever else may be inside the json (valid or invalid). This should also be properly documented.

mnunberg:
I'll have a PR up for this in a bit. `NOESCAPE` will return the JSON string as-is, doing minimal conversion and escaping - only what's required in the spec to make it be valid JSON. It will not insert \u-XXXX escapes. The `NOESCAPE` is a bit of a misnomer since it will still escape things like `\\` and `"`", and others specified in https://tools.ietf.org/html/rfc7159#section-7

dvirsky:
@wgjak47 I've merged @mnunberg's PR, so I'm closing this one. Thanks for all the help with the issue!

PR #63, opened 2018-04-02
ADDED: Dockerfile for Alpine.

HypertextPP:
If the Redis image has an :alpine tag, so should this module.

### DIFF
--- a/Dockerfile_Alpine
+++ b/Dockerfile_Alpine
@@ -1,0 +1,28 @@
+FROM redis:alpine as builder
+
+ENV LIBDIR /usr/lib/redis/modules
+ENV DEPS "python py-setuptools py-pip wget unzip make g++"
+# Set up a build environment
+RUN set -ex;\
+    deps="$DEPS";\
+    apk update; \
+	apk add $deps;\
+    pip install rmtest; 
+
+# Build the source
+ADD . /REJSON
+WORKDIR /REJSON
+RUN set -ex;\
+    make clean; \
+    deps="$DEPS"; \
+    make all -j 4;
+    
+# Package the runner
+FROM redis:alpine
+ENV LIBDIR /usr/lib/redis/modules
+WORKDIR /data
+RUN set -ex;\
+    mkdir -p "$LIBDIR";
+COPY --from=builder /REJSON/src/rejson.so  "$LIBDIR"
+
+CMD ["redis-server", "--loadmodule", "/usr/lib/redis/modules/rejson.so"]


mnunberg:
@itamarhaber @dvirsky - should we do this across other modules?

itamarhaber:
Makes sense to me

BorntraegerMarc:
Can this PR be merged? :)

BorntraegerMarc:
@itamarhaber should we add this across other modules before merging this PR? I don't see much sense in waiting to merge this PR...

mnunberg:
I apologize for being slow to get around to this.

I'm looking at this PR, and it seems a bit heavy handed to make a new dockerfile for Alpine. A better approach (IMO) would be to have a setup script that would call the appropriate commands depending on the distro (i.e. check `lsb_release`, or `uname`, etc.). If we one day decide to release an RHEL-based Redis image, then we'd need to have a Dockerfile_RHEL, and this can get annoying.

There is a Dockerfile_Dist, but that's really because of legacy reasons, and is actually the *real* dockerfile. I should actually change that.

HypertextPP:
@mnunberg Would it not make sense for the image to be based on Alpine rather than Ubuntu anyway? I personally see no gain in running a containerised Redis under anything other than Alpine. Debian and RHEL are pretty extensive just for a key-value store.

HypertextPP:
For me, one of the main purposes of Docker (and similar) is to slim everything down as much as possible and make development more productive. Having Redis sat on Ubuntu does the opposite purely because of how large it is and how much more resource it demands compared to Alpine.

mnunberg:
@HypertextPP It all depends on how Redis relates to your system, and whether Redis is just servicing the odd key here and there, and being deployed because of its simple setup, or whether your application uses Redis heavily, and requires a large degree of concurrency and will be under constant load.

We benchmark and test our Modules (and Redis itself) under traditional platforms. There's no reason why Alpine *shouldn't* work though. Maybe we can find a minimal image that still uses glibc, though this is more a discussion for Redis itself.

I also see my suggestion of simply detecting the environment won't work either.

BorntraegerMarc:
In my opinion it's a pretty common thing to do in docker to distribute multiple images based on the needs of the user. Take https://hub.docker.com/r/library/mongo/tags/ for example: has images for ubuntu, windows, alpine

BorntraegerMarc:
@HypertextPP did you by any chance get around to actually publish the alpine version of rejson to the docker hub?

HypertextPP:
@BorntraegerMarc No. I was waiting for the alpine version to be pushed to the 'official' repository otherwise people (including me!) just get confused.

BorntraegerMarc:
Makes sense. Soooo sorry for bothering again @mnunberg but I don't really understand if there is something still to do to merge this PR. And if yes -> How can I help?

BorntraegerMarc:
@HypertextPP Indeed it would be helpful :)

HypertextPP:
@BorntraegerMarc [Here ya go](https://hub.docker.com/r/bwatton/alpine-rejson/). Enjoy.

bwdotdev:
I think this is adequately stale enough to warrant closing.

PR #78, opened 2018-07-03
Added rejonson module to the client libraries and added client types

KromDaniel:
After talking with @itamarhaber 
I added client support module I wrote for go lang

I also specified the go-clients the client type


itamarhaber:
Thanks - merged and pushing the changes (could take a few hours for CDN to update though).

PR #79, opened 2018-07-04
path is not mandatory on JSON.DEL

gkorland:


itamarhaber:
Thanks @gkorland - merging and pushing.

PR #128, opened 2019-11-09
Removed random link to and update license scheme..

tombatron:
Looks like a URL to the NReJSON library made its way into the body of the quick start. 

Updated the client table with the license scheme of NReJSON.


gkorland:
Thanks!

PR #158, opened 2020-04-20
Corrected library name for MacOS

simonprickett:
Corrected the library name for MacOS, which is now `librejson.dylib`.


gavrie:
@simonprickett thanks for pointing this out!
We fixed it as part of a wider renaming change in #161 .

PR #177, opened 2020-05-05
add `JSON.INDEX DEL <index>`

gkorland:


src/commands/index.rs
```diff
@@ -67,6 +67,13 @@ fn add_field(index_name: &str, field_name: &str, path: &str) -> RedisResult {
     }
 }
 
+fn del_index(index_name: &str) -> RedisResult {
```
gavrie, line 70: Let's be consistent with the naming, and call the function `del_schema`. Way back we used the 'schema' term in this modules since the name 'index' was too overloaded.

test/pytest/test.py
```diff
@@ -745,6 +745,34 @@ def testIssue_74(self):
...
+            r.flushdb()
+
+            def do(*args):
+                self.assertOk(r.execute_command(*args))
+
+            do('JSON.INDEX', 'ADD', 'index', 'first', '$.first')
+            do('JSON.INDEX', 'ADD', 'index', 'second', '$.second')
+            do('JSON.INDEX', 'ADD', 'index2', 'second', '$.third')
+
+            # Error should be throw since this field already exists in the index
```
gavrie, line 762: `throw` => `thrown`

src/commands/index.rs
gkorland: done

test/pytest/test.py
gkorland: done

PR #179, opened 2020-05-06
fix #151 Don't return deleted docs on qget

### LINKED ISSUE #151: FLUSHDB doesn't affect index
Index survives past `FLUSHDB` - this will be undesirable. 

Consider this:
```
$ launchctl start io.redis.redis-server
$ redis-cli
> keys *
(empty list or set)
> json.set user1 $ '{"last":"Joe", "first":"Mc"}' INDEX person
OK
> json.index add person last $.last
OK
> JSON.QGET person Jo*
"{\"user1\":[{\"last\":\"Joe\",\"first\":\"Mc\"}]}"
> flushdb
OK
> JSON.QGET person Jo*
"{\"user1\":[]}"
```

This will be a very unexpected outcome for most Redis users. `FLUSHDB` (or `FLUSHALL`) should remove all data and this type of behaviour could reveal information that could be sensitive. 

I assume that the indexes are behind held outside of keyspace, which I get to some degree, but this will be very strange to explain.

gkorland:
Copy from https://github.com/RedisJSON/RedisJSON2/pull/76/

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,4 +14,4 @@ serde_json = "1.0"
 libc = "0.2"
 jsonpath_lib = { git="https://github.com/gkorland/jsonpath.git", branch="patch-1" }
 redis-module = { version="0.7.0", features = ["experimental-api"]}
-redisearch_api = "0.3.0"
+redisearch_api =  { git="https://github.com/RediSearch/redisearch-api-rs", branch="del_doc"}
--- a/src/commands/index.rs
+++ b/src/commands/index.rs
@@ -4,7 +4,7 @@ use serde_json::{Map, Value};
 
 use redis_module::{Context, NextArg, RedisError, RedisResult, RedisValue, REDIS_OK};
 
-use redisearch_api::{Document, FieldType};
+use redisearch_api::{Document, FieldType, TagOptions};
 
 use crate::error::Error;
 use crate::redisjson::{Format, RedisJSON};
@@ -61,7 +61,9 @@ fn add_field(index_name: &str, field_name: &str, path: &str) -> RedisResult {
     if schema.fields.contains_key(field_name) {
         Err("Field already exists".into())
     } else {
-        schema.index.create_field(field_name, 1.0, None);
+        schema
+            .index
+            .create_field(field_name, 1.0, TagOptions::default());
         schema.fields.insert(field_name.to_owned(), path.to_owned());
         REDIS_OK
     }
--- a/src/commands/index.rs
+++ b/src/commands/index.rs
@@ -61,7 +61,7 @@ fn add_field(index_name: &str, field_name: &str, path: &str) -> RedisResult {
     if schema.fields.contains_key(field_name) {
         Err("Field already exists".into())
     } else {
-        schema.index.create_field(field_name, 1.0, None, false);
+        schema.index.create_field(field_name, 1.0, None);
         schema.fields.insert(field_name.to_owned(), path.to_owned());
         REDIS_OK
     }
--- a/src/commands/index.rs
+++ b/src/commands/index.rs
@@ -61,7 +61,7 @@ fn add_field(index_name: &str, field_name: &str, path: &str) -> RedisResult {
     if schema.fields.contains_key(field_name) {
         Err("Field already exists".into())
     } else {
-        schema.index.create_field(field_name);
+        schema.index.create_field(field_name, 1.0, None, false);
         schema.fields.insert(field_name.to_owned(), path.to_owned());
         REDIS_OK
     }
@@ -79,6 +79,13 @@ pub fn add_document(key: &str, index_name: &str, doc: &RedisJSON) -> RedisResult
     if let Some(schema) = map.get(index_name) {
         let rsdoc = create_document(key, schema, doc)?;
         schema.index.add_document(&rsdoc)?;
+    }
+    REDIS_OK
+}
+
+pub fn remove_document(key: &str, index_name: &str) -> RedisResult {
+    if let Some(schema) = schema_map::as_ref().get(index_name) {
+        schema.index.del_document(&key)?;
     }
     REDIS_OK
 }
@@ -170,9 +177,9 @@ fn scan_and_index(ctx: &Context, schema: &Schema, cursor: u64) -> Result<u64, Re
                             .get_value::<RedisJSON>(&REDIS_JSON_TYPE)
                             .and_then(|doc| {
                                 if let Some(data) = doc {
-                                    if let Some(index) = &data.index {
-                                        if schema.name == *index {
-                                            add_document(key, index, data)?;
+                                    if let Some(index) = &data.value_index {
+                                        if schema.name == index.index {
+                                            add_document(key, &index.index, data)?;
                                         }
                                     }
                                     Ok(())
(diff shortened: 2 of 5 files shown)


src/commands/index.rs
```diff
@@ -83,6 +83,13 @@ pub fn add_document(key: &str, index_name: &str, doc: &RedisJSON) -> RedisResult
     REDIS_OK
 }
 
+pub fn remove_document(key: &str, index_name: &str) -> RedisResult {
+    if let Some(schema) = schema_map::as_ref().get(index_name) {
```
gavrie, line 87: Please use the explicit `let map = schema_map::as_ref();` as in the other methods for consistency (and I also think it's more readable).

src/commands/index.rs
```diff
@@ -170,9 +177,9 @@ fn scan_and_index(ctx: &Context, schema: &Schema, cursor: u64) -> Result<u64, Re
                             .get_value::<RedisJSON>(&REDIS_JSON_TYPE)
                             .and_then(|doc| {
                                 if let Some(data) = doc {
-                                    if let Some(index) = &data.index {
-                                        if schema.name == *index {
-                                            add_document(key, index, data)?;
+                                    if let Some(index) = &data.value_index {
+                                        if schema.name == index.index {
+                                            add_document(key, &index.index, data)?;
```
gavrie, line 182: Please rename the variable inside the pattern match for consistency:

```
if let Some(value_index) = &data.value_index {
    if schema.name == value_index.index {
        add_document(key, &value_index.index, data)?;
```

src/redisjson.rs
```diff
@@ -62,10 +63,16 @@ impl Path {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct ValueIndex {
+    pub key: String,
+    pub index: String,
+}
```
gavrie, line 70: I suggest naming this field `index_name` and not `index`; there's too many things `index` can refer to already.

src/lib.rs
```diff
@@ -98,7 +98,7 @@ fn json_set(ctx: &Context, args: Vec<String>) -> RedisResult {
 
     let mut format = Format::JSON;
     let mut set_option = SetOptions::None;
-    let mut index = None;
+    let mut value_index = None;
 
     loop {
```
gavrie, line 103: Not related to your change, but `cargo clippy` rightly suggests to simplify this `loop` with `while let`:

```
    while let Some(s) = args.next() {
        match s.to_uppercase().as_str() {
```

Cargo.toml
```diff
@@ -14,4 +14,4 @@ serde_json = "1.0"
 libc = "0.2"
 jsonpath_lib = { git="https://github.com/gkorland/jsonpath.git", branch="patch-1" }
 redis-module = { version="0.7.0", features = ["experimental-api"]}
-redisearch_api = "0.3.0"
+redisearch_api =  { git="https://github.com/RediSearch/redisearch-api-rs", branch="del_doc"}
```
gavrie, line 17: After finishing your `redisearch_api` PR (and releasing) this should be updated to a new release.

src/commands/index.rs
```diff
@@ -61,7 +61,7 @@ fn add_field(index_name: &str, field_name: &str, path: &str) -> RedisResult {
     if schema.fields.contains_key(field_name) {
         Err("Field already exists".into())
     } else {
-        schema.index.create_field(field_name);
+        schema.index.create_field(field_name, 1.0, None, false);
```
gavrie, line 64: We should fix the ambiguous arguments (per your redisearch_api PR).

src/commands/index.rs
```diff
@@ -209,27 +216,34 @@ where
...
-                                        .map_err(|e| e.into()) // Convert Error to RedisError
-                                        .map(|values| {
-                                            values.into_iter().map(|val| val.clone()).collect()
-                                        })
+                        let redis_key = ctx.open_key(&key);
+                        if redis_key.is_null() {
+                            // remove doc that doesn't exist from index
+                            remove_document(&key, &index_name)?;
+                            Ok(acc)
+                        } else {
```
gavrie, line 228: Please explain when this can happen, if we remove it from the index when deleting. Is this for backwards compatibility?

src/redisjson.rs
```diff
@@ -402,16 +412,22 @@ pub mod type_methods {
 
     #[allow(non_snake_case, unused)]
     pub unsafe extern "C" fn free(value: *mut c_void) {
-        Box::from_raw(value as *mut RedisJSON);
+        let json = value as *mut RedisJSON;
+        if let Some(index) = &(*json).value_index {
+            index::remove_document(&index.key, &index.index);
+        }
+
+        Box::from_raw(json);
```
gavrie, line 420: This code can be cleaned up a bit by boxing first instead of using the raw pointer (and adding a comment):
```
        let json = value as *mut RedisJSON;

        // Take ownership of the data from Redis (causing it to be dropped when we return)
        let json = Box::from_raw(json);

        if let Some(index) = &json.value_index {
            index::remove_document(&index.key, &index.index);
        }
```

src/commands/index.rs
gkorland: :+1:

src/redisjson.rs
gkorland: :+1:

src/lib.rs
gkorland: :+1:

src/commands/index.rs
gkorland: :+1:

src/redisjson.rs
gkorland: :+1:

src/commands/index.rs
gkorland: it was the original lazy solution, I think it can be removed now.

src/commands/index.rs
gkorland: :+1:

src/lib.rs
```diff
@@ -125,8 +124,8 @@ fn json_set(ctx: &Context, args: Vec<String>) -> RedisResult {
     match (current, set_option) {
         (Some(ref mut doc), ref op) => {
             if doc.set_value(&value, &path, op, format)? {
-                if let Some(index) = index {
-                    index::add_document(&key, &index, &doc)?;
+                if let Some(index) = value_index {
```
gavrie, line 127: For consistency this should be `if let Some(value_index) = value_index {`

src/lib.rs
```diff
@@ -136,16 +135,16 @@ fn json_set(ctx: &Context, args: Vec<String>) -> RedisResult {
         }
         (None, SetOptions::AlreadyExists) => Ok(RedisValue::Null),
         (None, _) => {
-            let doc = RedisJSON::from_str(&value, &index, format)?;
+            let doc = RedisJSON::from_str(&value, &value_index, format)?;
             if path == "$" {
                 redis_key.set_value(&REDIS_JSON_TYPE, doc)?;
 
-                if let Some(index) = index {
+                if let Some(index) = value_index {
```
gavrie, line 142: As above: `if let Some(value_index) = value_index {`

src/lib.rs
gkorland: :+1:

src/lib.rs
gkorland: :+1:

PR #193, opened 2020-05-27
escape line feeds in json path erros so they can be used in a redis e…

yoav-steinberg:
…rror response.


yoav-steinberg:
Note that without this we fail `testIssue_74` when using latest jsonpath_lib which correctly handles dollar sign misuse in paths.

PR #194, opened 2020-05-27
Fix #149 bracket notation

### LINKED ISSUE #149: Bracket notation no longer supported?
When I try the following is Redis Commander:
`JSON.SET "DATA" '["ID7"]' "{}"`

...I get:
`"Err: wrong static path"`

Has bracket notation support been dropped in 2?

yoav-steinberg:


test/pytest/test.py
```diff
@@ -197,7 +197,22 @@ def testSetBehaviorModifyingSubcommands(self):
...
             # with self.assertRaises(redis.exceptions.ResponseError) as cm:
             #     r.execute_command('JSON.SET', 'test', '.foo[1]', 'null', 'XX')
-    
+
+    def testSetWithBracketNotation(self):
+        with self.redis() as r:
+            r.client_setname(self._testMethodName)
+            r.flushdb()
+
+            # test against the root
```
gavrie, line 206: Would it make sense to add some negative tests here and check that they generate the relevant error messages?

yoav-steinberg: Added relevant tests. Note that there's one case which is handled in the code but is only theoretical so I didn't add a test for it:
`Err: path must end with object key to set` isn't relevant because "root sets" are handled by the calling function (`set_value`). I wanted to handle this anyway because of the generic implementation I did inside `StaticPathParser::check`. The more complete way to handle this is to used build in rust unit testing, which currently isn't part of this project and I didn't think this is enough of an excuse for adding it.

gavrie: Ok, makes sense (though it's always a good time to add unit tests 😄)

PR #200, opened 2020-06-04
fixbug: after the command of JSON.ARRPOP and JSON.STRAPPEND, the lru …

fayadexinqing:
…cache should be invalid
The command of JSON.ARRPOP and JSON.STRAPPEND are write command, when the cache is on, after these commands is executed, the cache of key should be invalid.


fayadexinqing:
@itamarhaber I have signed the CLA. please take a look.

Issue #154, opened 2019-09-15
Add `JSON.QSET <index> <query> <path> <json> [NX | XX]`


gkorland:

yoav-steinberg:
Regarding `<path>` option here there are two ways to go:
1. Support only static paths like in `JSON.SET`, this makes it compatible with `JSON.SET` and supports simple code reuse.
1. Support any jsonpath, this means more complex updates are possible and we can't reuse code.
Does this make sense? Which should we aim for?

gkorland:
I think for the first stage we should go with 1, and when we define the semantic for json.set also extend here

yoav-steinberg:
> I think for the first stage we should go with 1, and when we define the semantic for json.set also extend here

Very well. This makes things simpler.

Another question: `JSON.SET` returns `OK` in case we set a value and _null_ in case no value was set. When implementing `JSON.QSET` we can have either the same logic or return an integer indicating the number of keys which were affected. If we go for the integer then what should we return if no keys are affected null or 0?
My thought is that keeping in line with other redis commands and also the most practical response is an integer with the number of affected keys and 0 representing no keys affected.

gkorland:
> My thought is that keeping in line with other redis commands and also the most practical response is an integer with the number of affected keys and 0 representing no keys affected.

Agree

PR #203, opened 2020-06-10
Add support for JSON.QSET see #154

yoav-steinberg:


src/error.rs
```diff
@@ -1,38 +1,50 @@
...
+use std::fmt;
 
 #[derive(Debug)]
-pub struct Error {
-    pub msg: String,
+pub enum Error {
+    GeneralError(String),
+    PathNotAnObject,
+    WrongStaticPath,
 }
```
gavrie, line 9: Maybe it's time to switch to [thiserror](https://docs.rs/thiserror/1.0.19/thiserror/) (the de factor standard for structured errors) and remove some of the boilerplate in this file?

src/redisjson.rs
```diff
@@ -11,6 +11,7 @@ use crate::formatter::RedisJsonFormatter;
 use crate::nodevisitor::{StaticPathElement, StaticPathParser, VisitStatus};
 use crate::REDIS_JSON_TYPE_VERSION;
 
+use crate::error::Error::{GeneralError, PathNotAnObject, WrongStaticPath};
```
gavrie, line 14: I think it would be more readable just to `use crate::error::Error` here and then explicitly write `Err(Error::GeneralError)` etc. below.

src/error.rs
```diff
@@ -20,15 +32,9 @@ impl From<&str> for Error {
...
-impl From<serde_json::Error> for Error {
-    fn from(e: serde_json::Error) -> Self {
-        Error::GeneralError(e.to_string())
-    }
-}
-
-impl From<JsonPathError> for Error {
-    fn from(e: JsonPathError) -> Self {
-        Error::GeneralError(format!("JSON path error: {:?}", e).replace("\n", "\\n"))
+impl From<jsonpath_lib::JsonPathError> for Error {
```
gavrie, line 35: I was curious why you didn't use `#[from]` for this as well, and understood (by trying) that the reason is that `jsonpath_lib::JsonPathError` doesn't `impl std::error::Error`. Is that right?

If so, please add a comment that explains this for the next person who wonders.

Also: As long as the manual `From` is needed, I think it's cleaner to leave the `\n` escaping here instead of stuffing it into the `#[error] above.

gkorland: > jsonpath_lib::JsonPathError doesn't impl std::error::Error.
@yoav-steinberg let's open a PR on jsonpath_lib to change it

yoav-steinberg: Issued pr: https://github.com/freestrings/jsonpath/pull/52

PR #230, opened 2020-10-12
Fix #223 Return null on missing path

### LINKED ISSUE #223: JSON.MGET - wrong logic behaviour
ive pulled latest commit(8959470a24ce83e778c4e6b82448fb1e18006997)
and build on Mac OS Catalina:

ive found that the behaviour of MGET is a bit weird.

A try to get missing path from existing keys raises an error , but on the other hand- if no keys at all - MGET return null’s result

Missing path when at least 1 key exists:

JSON.SET akey1 . '{"obj1":{"obj2":{"val":1}}}'
JSON.SET akey2 . '{"obj1":{"obj3":{"val":1}}}'
JSON.MGET akey1 akey2 no_such_key

Result:
"ERR path does not exist"

Missing path when no keys exists:

JSON.MGET akey1 akey2 no_such_key

Result:
 1) null
 2)  null

Thats doesn’t match the documentation at all:

Description ¶

Returns the values at path from multiple key s. Non-existing keys and non-existing paths are reported as null.

gkorland:
Fix #223


src/lib.rs
```diff
@@ -218,18 +218,29 @@ fn json_mget(ctx: &Context, args: Vec<String>) -> RedisResult {
...
-                let result = ctx
-                    .open_key(key)
+                ctx.open_key(key)
                     .get_value::<RedisJSON>(&REDIS_JSON_TYPE)?
                     .map(|doc| doc.to_string(&path, Format::JSON))
-                    .transpose()?;
-
-                Ok(result.into())
+                    .transpose()
+                    .map_or_else(|_| Ok(RedisValue::Null), |v| Ok(v.into()))
```
oshadmi, line 225: It seems that current tests need to be updated accordingly.
Currently a failed key (`foo`) is tested in `testMgetCommand` in RedisJSON/tests/pytest/test.py

Now a failed path does not raise an exception, so this needs to be modified?
`r.expect('JSON.MGET', 'doc:0', 'doc:1', '42isnotapath').raiseError()`

And should probably add another test for combined successful and failed paths, such as:
```
    # Test that MGET fails on path errors
    r.cmd('DEL', 'test')
    r.assertOk(r.execute_command('JSON.SET', 'test', '.', '{"bull":4.2}'))
    raw = r.execute_command('JSON.MGET', 'doc:0', 'test', 'doc:1', '.bool')
    r.assertEqual(len(raw), 3)
    r.assertTrue(json.loads(raw[0]))
    r.assertEqual(raw[1], None)
    r.assertTrue(json.loads(raw[2]))

```

oshadmi (approved):
👍🏼

PR #244, opened 2020-12-03
Redesign command docs

ashtul:
* Depth changed to exclude subfields for commands
* Commands broken into sections
* Commands order changed

Look after the change -

RedisJSON commands


emmanuelkeller (approved):
Much cleaner ! The new order makes sense

PR #282, opened 2021-03-15
Change first key index to 2

### LINKED ISSUE #281: JSON.DEBUG wronge key position
```
127.0.0.1:12000> JSON.SET jsonkey . '{"a": "a"}'
OK
127.0.0.1:12000> JSON.GET jsonkey
"{\"a\":\"a\"}"
127.0.0.1:12000> JSON.DEBUG MEMORY jsonkey
(nil)
127.0.0.1:12000> JSON.DEBUG memory jsonkey
(integer) 82
```

The bug in line https://github.com/RedisJSON/RedisJSON/blob/6a641c11897ad0debac487ec16cc3bc37ff38937/src/rejson.c#L1919

The key position should be 2 and not 1

gkorland:
fix #281

### DIFF
--- a/src/rejson.c
+++ b/src/rejson.c
@@ -1917,7 +1917,7 @@ int Module_CreateCommands(RedisModuleCtx *ctx) {
         return REDISMODULE_ERR;
 
     if (RedisModule_CreateCommand(ctx, "json.debug", JSONDebug_RedisCommand, "readonly getkeys-api",
-                                  1, 1, 1) == REDISMODULE_ERR)
+                                  2, 2, 1) == REDISMODULE_ERR)
         return REDISMODULE_ERR;
 
     if (RedisModule_CreateCommand(ctx, "json.type", JSONType_RedisCommand, "readonly", 1, 1, 1) ==
--- a/test/pytest/test.py
+++ b/test/pytest/test.py
@@ -719,6 +719,10 @@ class ReJSONTestCase(BaseReJSONTest):
         self.assertEqual(1512060373.222988, float(res))
         # self.assertEqual('1512060373.222988', res)
 
+    def testDebugMemory(self):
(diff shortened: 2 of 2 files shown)


src/rejson.c
```diff
@@ -1917,7 +1917,7 @@ int Module_CreateCommands(RedisModuleCtx *ctx) {
         return REDISMODULE_ERR;
 
     if (RedisModule_CreateCommand(ctx, "json.debug", JSONDebug_RedisCommand, "readonly getkeys-api",
-                                  1, 1, 1) == REDISMODULE_ERR)
+                                  2, 2, 1) == REDISMODULE_ERR)
```
itamarhaber, line 1920: I'm not sure that this is the right fix it.

I think the problem is at https://github.com/RedisJSON/RedisJSON/blob/1.0/src/rejson.c#L258, which should become:

```C
        RedisModuleKey *key = RedisModule_OpenKey(ctx, argv[3], REDISMODULE_READ);
```

test/pytest/test.py
```diff
@@ -719,6 +719,10 @@ def testDoubleParse(self):
         self.assertEqual(1512060373.222988, float(res))
         # self.assertEqual('1512060373.222988', res)
 
+    def testDebugMemory(self):
+        self.cmd('JSON.SET', 'memoryTest', '.', '{"key1":"val", "key2":[1,2,3], "key3":{"key4":1,"key5":false}}')
+        res = self.cmd('JSON.DEBUG', 'MEMORY', 'memoryTest')
+        self.assertEqual(439.0, float(res))
```
itamarhaber, line 725: Tests are always good, but tying this to the implementation is bad. I would just check > 0 (and maybe < 1000).

rafie (approved):

PR #288, opened 2021-03-23
remove Search dep

gkorland:


src/redisjson.rs
```diff
@@ -437,23 +420,10 @@ pub mod type_methods {
...
             0 => RedisJSON {
                 data: backward::json_rdb_load(rdb),
-                value_index: None, // TODO handle load from rdb
             },
             2 => {
                 let data = raw::load_string(rdb);
-                let schema = if raw::load_unsigned(rdb) > 0 {
-                    Some(ValueIndex {
-                        key: raw::load_string(rdb),
-                        index_name: raw::load_string(rdb),
```
oshadmi, line 447: A bit far fetched, but is it possible that users extended the module and are writing additional module data to the db, and we should just consume (and ignore) our old schema data to avoid breaking user extension?

oshadmi (approved):
LGTM 👍🏼

src/lib.rs
```diff
@@ -102,7 +98,6 @@ fn json_set(ctx: &Context, args: Vec<String>) -> RedisResult {
 
     let mut format = Format::JSON;
     let mut set_option = SetOptions::None;
-    let mut value_index = None;
```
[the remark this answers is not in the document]
gkorland: :+1:

src/redisjson.rs
gkorland: Do you mean the redisearch 2.0 schema support that we just removed?

PR #306, opened 2021-04-19
Add JSON.CLEAR command

oshadmi:
Currently only handling the first `path` argument


oshadmi:
test:testClear - has tests for when multi-paths are handled - currently only the first is handled.
		❌  (FAIL):	'42' == [42, 44, {'a': 1, 'b': 2}, ['to', 'be', 'cleared', 4]]	test.py:413
		❌  (FAIL):	'42' == [42, 44, {}, []]	test.py:435
test:testSetInvalidPathShouldFail - also fails without this PR. Will need to be reviewed later.

oshadmi:
test:testSetInvalidPathShouldFail - it turns out this test does not fails on master. Reviewing it.

src/lib.rs
```diff
@@ -705,6 +713,33 @@ fn json_obj_len(ctx: &Context, args: Vec<String>) -> RedisResult {
...
+/// JSON.CLEAR <key> [path ...]
+///
+fn json_clear(ctx: &Context, args: Vec<String>) -> RedisResult {
+    let mut args = args.into_iter().skip(1);
+    let key = args.next_string()?;
+    let arg = get_optional_arg(&mut args, "$");
+    let mut paths: Vec<Path> = vec![];
+    let path = backwards_compat_path(arg);
+    paths.push(Path::new(path));
+    while let Ok(mut arg) = args.next_string() {
```
gkorland, line 726: @oshadmi why do you need multi paths support? with the support of full JSONPath I don't see the need.

oshadmi: @gkorland Do you mean that since multi paths could be expressed using a single JSONPath with wildcard, union iterator, recursive descent or slices?
It could still be useful to allow specifying several paths, either when they are plain/explicit, or when a complex path, with wildcards, etc., might yield unwanted paths, e.g., `$..attr` is too "permissive" when user only wants 2 specific paths out of many others, so `$.sub1.attr` and `$.sub2.attr` could be easily specified, while finding the exact single complex path could be harder.

gkorland: yes single "path" can be a union of multi paths.
And it's consistent with:  ```JSON.DEL <key> [path]```

oshadmi: @gkorland 
The current [doc](https://oss.redislabs.com/redisjson/commands/#jsonget) specifies one or more `path` args **only** for `JSON.GET` (using three dots)
```
JSON.GET <key>
         [INDENT indentation-string]
         [NEWLINE line-break-string]
         [SPACE space-string]
         [NOESCAPE]
         [path ...]
```
But for [JSON.DEL](https://oss.redislabs.com/redisjson/commands/#jsondel) only a single path is specified:
`JSON.DEL <key> [path]`

For `JSON.CLEAR` we aimed to eventually support multi-paths, right?
So it would eventually be documented with three dots as:
`JSON.CLEAR <key> [path ...]`
With the return value being the number of paths cleared, similar to `JSON.DEL`.

Additional question (also related to single path which could use wildcards, union iterator, etc. to match several elements):
What should be done if one or more paths do not exist or do not resolve to an array or an object:
1) Should the command be atomic, and avoid doing just part of the job? Should it then return the error messages reporting 
 the wrong paths, or just do nothing or return nil?
Or
2) Should the command report on erroneous path and resume with the rest of the paths, clearing only existing and matching arrays/objects and skipping the rest? (if the command is not atomic, the return value would be an array of numbers and/or error messages?)

src/lib.rs
```diff
@@ -705,6 +713,33 @@ fn json_obj_len(ctx: &Context, args: Vec<String>) -> RedisResult {
...
 
+///
+/// JSON.CLEAR <key> [path ...]
+///
+fn json_clear(ctx: &Context, args: Vec<String>) -> RedisResult {
+    let mut args = args.into_iter().skip(1);
+    let key = args.next_string()?;
+    let arg = get_optional_arg(&mut args, "$");
+    let mut paths: Vec<Path> = vec![];
+    let path = backwards_compat_path(arg);
```
gavrie, line 724: Why the need for `backwards_compat_path`? It seems like `Path::new` already has this logic; it's just duplication.

In any case, the code here is not very idiomatic Rust:
- Mutable variables are used, making the code harder to reason about
- The logic for the first arg and the rest is duplicated.

This whole block can be replaced by:
```
    let key = args.next_string()?;

    let paths = args
        .map(backwards_compat_path)
        .map(Path::new)
        .collect::<Vec<_>>();

    let paths = if paths.is_empty() {
        vec![Path::new("$".to_string())]
    } else {
        paths
    };
```

No need for the `get_optional_arg` and repeated logic. And the `.map(backwards_compat_path)` can be removed, unless I'm missing something.

gavrie (changes requested):

gavrie (approved):

PR #319, opened 2021-04-22
Add fake aux load

### LINKED ISSUE #318: Recent commit breaks loading of previous RDB backups
Hi everyone,

We discovered that the latest commits in master no longer support loading RDB backups from earlier commits, and get the following error:

```
8:M 21 Apr 2021 14:24:45.394 # The RDB file contains module AUX data, but the module 'ReJSON-RL' doesn't seem to support it.
```

The issue can be reproduced with the following procedure (tested using Redis 6.2.1):

1. Enable RDB backups in `redis.conf`, using `save 1 1` to ensure keys get backed up immediately.
2. Run Redis with commit `c205258` of RedisJSON.
3. Set a key: `redis-cli JSON.SET test . '{"foo": "bar", "ans": 42}'`
4. Stop Redis.
5. Run Redis with the next commit, `6202520`, of RedisJSON.

Redis will exit with the above error message.

Is there a way to make ReJSON compatible with earlier versions/commits?

Regards,
Alex.

gkorland:
fix #318

### DIFF
--- a/src/redisjson.rs
+++ b/src/redisjson.rs
@@ -8,10 +8,11 @@ use crate::backward;
 use crate::error::Error;
 use crate::formatter::RedisJsonFormatter;
 use crate::nodevisitor::{StaticPathElement, StaticPathParser, VisitStatus};
+use crate::REDIS_JSON_TYPE_VERSION;
 
 use bson::decode_document;
 use jsonpath_lib::SelectorMut;
-use redis_module::raw;
+use redis_module::raw::{self, Status};
 use serde::Serialize;
 use serde_json::{Map, Value};
 use std::io::Cursor;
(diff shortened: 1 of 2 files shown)


src/redisjson.rs
```diff
@@ -443,4 +444,30 @@ pub mod type_methods {
         let json = &*(value as *mut RedisJSON);
         raw::save_string(rdb, &json.data.to_string());
     }
+
```
MeirShpilraien, line 447: Maybe add a comment why its needed, otherwise someday someone will remove it again.

MeirShpilraien (approved):

gavrie (approved):

PR #336, opened 2021-05-01
added generic json path implementation

MeirShpilraien:


src/select/cmp.rs
```diff
@@ -0,0 +1,361 @@
+use crate::select::select_value::SelectValue;
+use array_tool::vec::{Intersect, Union};
```
gavrie, line 2: I'm not sure it's a good idea to add the `array_tool` crate just for this. It also looks old and unmaintained.
The intersection/union can be done with the standard `HashSet` type:
https://doc.rust-lang.org/std/collections/hash_set/index.html

src/redisjson.rs
```diff
@@ -7,11 +7,13 @@
...
 use crate::error::Error;
 use crate::formatter::RedisJsonFormatter;
+use crate::json_node::{replace_with, JsonValueUpdater};
 use crate::nodevisitor::{StaticPathElement, StaticPathParser, VisitStatus};
+use crate::select::{Selector, SelectorMut};
 use crate::REDIS_JSON_TYPE_VERSION;
 
 use bson::decode_document;
-use jsonpath_lib::SelectorMut;
+// use index::schema_map; todo: is this needed? it breaks the build!!!
```
gavrie, line 16: Looks like a leftover from the old search integration; should be removed.

src/lib.rs
```diff
@@ -1,6 +1,9 @@
 #[macro_use]
 extern crate redis_module;
 
+#[macro_use]
+extern crate log;
```
gavrie, line 5: This syntax is deprecated. Instead of it, just `use log::debug;` in the relevant file(s).

And while you're at it, we can remove the above `#[macro_use] extern crate redis_module;` and replace it with `use redis_module::{redis_command, redis_module};`.

src/json_node.rs
```diff
@@ -0,0 +1,240 @@
+use crate::select::select_value::{SelectValue, SelectValueType, ValueUpdater};
+use crate::select::JsonPathError;
+use crate::select::SelectorMut;
```
gavrie, line 3: It's hard to track exactly what you changed as compared to the original `jsonpath_lib`. I did a manual diff, but that's inconvenient compared to Git.

Instead, it would be better to keep the crate separate by forking the original and using the fork here (this is what we originally did). That way it will be easy to see what has changed, and also to track future upstream changes.
It will also be cleaner, keeping the code separate. And last but not least, the build will be faster.

Cargo.toml
```diff
@@ -9,6 +9,8 @@ crate-type = ["cdylib"]
 name = "rejson"
 
 [dependencies]
+log = "0.4"
+array_tool = "1.0.3"
```
gavrie, line 13: See below comment regarding `array_tool`.

src/select/select_value.rs
```diff
@@ -0,0 +1,35 @@
...
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum SelectValueType {
+    Undef,
+    Null,
+    Bool,
+    Long,
+    Double,
+    String,
+    Array,
+    Dict,
```
gavrie, line 12: Why `Dict` instead of `Object`? Seems inconsistent.

src/select/select_value.rs
```diff
@@ -0,0 +1,35 @@
+use crate::select::JsonPathError;
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum SelectValueType {
+    Undef,
```
gavrie, line 5: The `Undef` value is not necessary, as explained in another comment.

src/json_node.rs
```diff
@@ -0,0 +1,240 @@
...
+            Value::Number(n) => {
+                if n.is_i64() {
+                    SelectValueType::Long
+                } else if n.is_u64() {
+                    SelectValueType::Long
+                } else if n.is_f64() {
+                    SelectValueType::Double
+                } else {
+                    SelectValueType::Undef
+                }
```
gavrie, line 24: What is the purpose of `Undef`? Can it ever happen that the value of the number is neither of `i64/u64/f64`? It seems like it cannot, so it would be cleaner to simply `panic!()` here and get rid of the `Undef` value altogether.

src/select/select_value.rs
```diff
@@ -0,0 +1,35 @@
...
+pub trait SelectValue:
+    std::fmt::Debug + std::cmp::Eq + std::cmp::PartialEq + std::default::Default + std::clone::Clone
+{
+    fn get_type(&self) -> SelectValueType;
+    fn contains_key(&self, key: &str) -> bool;
+    fn values<'a>(&'a self) -> Option<Vec<&'a Self>>;
+    fn keys(&self) -> Option<Vec<String>>;
+    fn len(&self) -> Option<usize>;
+    fn get_key<'a>(&'a self, key: &str) -> Option<&'a Self>;
+    fn get_index<'a>(&'a self, index: usize) -> Option<&'a Self>;
```
gavrie, line 24: No need to explicitly specify the lifetime for `self` in all of these, the compiler determines it automatically.

src/select/select_value.rs
```diff
@@ -0,0 +1,35 @@
...
+    Bool,
+    Long,
+    Double,
+    String,
+    Array,
+    Dict,
+}
+
+pub trait SelectValue:
+    std::fmt::Debug + std::cmp::Eq + std::cmp::PartialEq + std::default::Default + std::clone::Clone
```
gavrie, line 16: The idiomatic way is to `use std::...` and then use the short names of the traits.

src/select/expr_term.rs
```diff
@@ -0,0 +1,290 @@
...
+use crate::select::{to_f64, FilterKey};
+
+#[derive(Debug, PartialEq)]
+pub(super) enum ExprTerm<'a, T>
+where
+    T: SelectValue,
+{
+    String(String),
+    Long(i64),
+    Double(f64),
```
gavrie, line 12: The original has just `Number`, what was the reason for splitting into `Long`/`Double`?

src/json_node.rs
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    fn contains_key(&self, key: &str) -> bool {
+        match self {
+            Value::Object(o) => o.contains_key(key),
+            _ => false,
+        }
+    }
+
+    fn values<'a>(&'a self) -> Option<Vec<&'a Self>> {
```
gavrie, line 36:
```suggestion
    fn values(&self) -> Option<Vec<&Self>> {
```

src/json_node.rs
```diff
@@ -0,0 +1,240 @@
...
+        match self {
+            Value::Array(arr) => {
+                let mut res = Vec::new();
+                for e in arr {
+                    res.push(e);
+                }
+                Some(res)
+            }
```
gavrie, line 44: The idiomatic way:
```suggestion
            Value::Array(arr) => Some(arr.iter().collect()),
```

src/json_node.rs
```diff
@@ -0,0 +1,240 @@
...
+            Value::Object(o) => {
+                let mut res = Vec::new();
+                for e in o.values() {
+                    res.push(e);
+                }
+                Some(res)
+            }
```
gavrie, line 51:
```suggestion
            Value::Object(obj) => Some(obj.values().collect()),
```

src/json_node.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    fn keys(&self) -> Option<Vec<String>> {
+        match self {
+            Value::Object(o) => {
+                let mut res = Vec::new();
+                for k in o.keys() {
+                    res.push(k.to_string());
+                }
+                Some(res)
+            }
```
gavrie, line 64:
```suggestion
            Value::Object(o) => Some(o.keys().cloned().collect()),
```

gavrie (changes requested):

src/json_node.rs
MeirShpilraien: I agree, will make a PR to our fork and you can review it there ..

src/select/expr_term.rs
MeirShpilraien: Number itself was split (IIRC) so I thought it redundant to create another enum for number that by itself will be split...

src/select/select_value.rs
MeirShpilraien: Are you sure? Its not compiling without it. We can look at it together on the interactive review...

src/json_node.rs
MeirShpilraien: See comment above

src/manager.rs
```diff
@@ -0,0 +1,532 @@
...
+pub enum UpdateInfo {
+    SUI(SetUpdateInfo),
+    AUI(AddUpdateInfo),
+}
+
+pub trait ReadHolder<V: SelectValue> {
+    fn get_value(&self) -> Result<Option<&V>, RedisError>;
+}
+
+pub trait WriteHolder<E: Clone, V: SelectValue> {
```
oshadmi, line 40: Used `O: Clone` in other locations (`E` is usually associated with Error)

src/commands.rs
```diff
@@ -0,0 +1,1173 @@
...
+        for p in paths {
+            res = Some(redis_key.str_append(p, json.clone())?);
+        }
+        redis_key.apply_changes(ctx, "json.strappend")?;
+        Ok(res.unwrap().into())
+    } else {
+        Err(RedisError::String(format!(
+            "Path '{}' does not exist",
+            path
+        )))
```
oshadmi, line 802: Could reach here also if path does exist, but its type is not `String`.

src/commands.rs
```diff
@@ -0,0 +1,1173 @@
...
+
+    let root = redis_key
+        .get_value()?
+        .ok_or_else(RedisError::nonexistent_key)?;
+
+    let paths = find_paths(&path, root, |v| v.get_type() == SelectValueType::String)?;
+    if paths.len() > 0 {
+        let mut res = None;
+        for p in paths {
+            res = Some(redis_key.str_append(p, json.clone())?);
```
oshadmi, line 794: - Will not report on paths leading to non-String values (different behavior than before)

- Might modify only a sub-set of `paths` and fail in the middle - not "atomic" (to be addressed by another PR?)

* Only reporting the first encountered error (using the `?` operator)

MeirShpilraien: * Yes but we need to decide this behavior, I believe we can decide it when redefine the API on 2.0
* Looking at the API it can happened but looking at the impl it can not, again we should give an api that can expose such errors
* Yes but again this is an api limitation

src/manager.rs
```diff
@@ -0,0 +1,532 @@
...
+            }
+        } else {
+            Err(RedisError::Str("bad input number"))
+        }
+    }
+
+    fn set_root(&mut self, v: Option<Value>) -> Result<(), RedisError> {
+        match v {
+            Some(inner) => {
+                let v = self.key.get_value::<RedisJSON>(&REDIS_JSON_TYPE)?;
```
oshadmi, line 207: Calling `get_value` could be redundant here, such as when coming from `command_json_set` calling `WriteHolder::set_value` calling `set_root`, where the value was already obtained before in `command_json_set` by
`let current = redis_key.get_value()?;`

src/commands.rs
```diff
@@ -0,0 +1,1173 @@
...
+    let current = redis_key.get_value()?;
+
+    let val = manager.from_str(&value, format)?;
+
+    match (current, set_option) {
+        (Some(ref mut doc), ref op) => {
+            if path == JSON_ROOT_PATH {
+                if *op != SetOptions::NotExists {
+                    redis_key.set_value(Vec::new(), val)?;
+                    redis_key.apply_changes(ctx, "json.set")?;
```
oshadmi, line 520: Should `apply_changes` be called only when the return value from `redis_key.set_value` is `Ok(true)`?
(although currently it seems to be possible to get only `Ok(false)` or `Err`)

src/manager.rs
```diff
@@ -0,0 +1,532 @@
...
+    }
+
+    fn set_value(&mut self, path: Vec<String>, mut v: Value) -> Result<bool, RedisError> {
+        let mut updated = false;
+        if path.is_empty() {
+            // update the root
+            self.set_root(Some(v))?;
+            updated = true;
+        } else {
+            self.update(&path, self.get_value().unwrap().unwrap(), |_v| {
```
oshadmi, line 253: `KeyHolderWrite::get_value` (and `KeyHolderRead::get_value`) may return `Ok(None)` so double `unwrap()` may panic? (also used in more locations)

src/manager.rs
```diff
@@ -0,0 +1,532 @@
...
+impl KeyHolderWrite {
+    fn update<F: FnMut(Value) -> Result<Option<Value>, Error>>(
+        &self,
+        path: &Vec<String>,
+        root: &mut Value,
+        mut func: F,
+    ) -> Result<(), Error> {
+        let mut target = root;
+
+        let last_index = path.len().saturating_sub(1);
```
oshadmi, line 102: Saw `saturating_sub` is nightly-only experimental API.
Might fail for users which do not have nightly rust?

src/manager.rs
MeirShpilraien: Yes but if you reach here its not possible that you will get none and so panic is OK here (if its happened its a bug and we want to know about it)

src/manager.rs
MeirShpilraien: I do not use a nightly build, I guess it uses this: https://doc.rust-lang.org/std/primitive.u32.html#method.saturating_sub

src/commands.rs
MeirShpilraien: Well, its updating the root so it can not failed (return false). We can assert it, WDYT?

oshadmi: There are more calls to `apply_changes`, some consider the result of the "write" operation (set_value, del, etc.),
But some don't, or only consider the last result of a `for` iteration, such as in `command_json_num_op` , `command_json_bool_toggle` , `command_json_str_append` , `command_json_arr_append` , `command_json_arr_insert` , `command_json_arr_trim` , `command_json_clear` 
However, this is also related to atomicity and the behavior needs to be decided as commented [here](https://github.com/RedisJSON/RedisJSON/pull/336#discussion_r659807305)
(also for `command_json_arr_pop`)

MeirShpilraien: Yes, the assumption is that the last result should be the same as the first result. but I agree this all story need to be handled together with the API changed

src/commands.rs
```diff
@@ -0,0 +1,1173 @@
...
+                    })
+                    .collect())
+            }
+        } else if let StaticPathElement::ArrayIndex(_) = last {
+            // if we reach here with array path we must be out of range
+            // otherwise the path would be valid to be set and we would not
+            // have reached here!!
+            Err("array index out of range".into())
+        } else {
+            Err("path not an object".into())
```
oshadmi, line 265: path not an object or array?

src/commands.rs
```diff
@@ -0,0 +1,1173 @@
...
+                .select_with_paths(|_| true)?;
+            if res.len() > 0 {
+                return Ok(res
+                    .drain(..)
+                    .map(|v| UpdateInfo::SUI(SetUpdateInfo { path: v }))
+                    .collect());
+            }
+        }
+        if SetOptions::AlreadyExists != *option {
+            self.find_add_paths(path)
```
oshadmi, line 288: If option is `NotExists` we will reach here - may `find_add_paths` find and return an existing value?

oshadmi (commented):
More questions...😊

src/commands.rs
MeirShpilraien: Yes this only gives you what need to be done, but the WriteHolder implementation will not let you set an existing value in a dictionary so it will be catch after.

oshadmi (approved):
👏🏼 Amazing job ❗

oshadmi (dismissed):

oshadmi (approved):

src/select/cmp.rs
MeirShpilraien: Removed

src/redisjson.rs
MeirShpilraien: Removed

src/lib.rs
MeirShpilraien: Removed

Cargo.toml
MeirShpilraien: Removed

src/select/select_value.rs
MeirShpilraien: Changed

src/select/select_value.rs
MeirShpilraien: Removed

src/json_node.rs
MeirShpilraien: Removed

src/json_node.rs
MeirShpilraien: Fixed

src/json_node.rs
MeirShpilraien: Fixed

src/json_node.rs
MeirShpilraien: Fixed

PR #344, opened 2021-05-04
[feature-search-json] Sync with master (2)

rafie:


src/lib.rs
```diff
@@ -4,8 +4,7 @@ extern crate redis_module;
 use std::{i64, usize};
 
 use redis_module::raw::RedisModuleTypeMethods;
-use redis_module::{native_types::RedisType, NotifyEvent};
-use redis_module::{raw as rawmod, NextArg};
+use redis_module::{native_types::RedisType, NextArg, NotifyEvent};
```
oshadmi, line 7: `raw as rawmod` was excluded from merge, should be something like:
`use redis_module::{native_types::RedisType, NextArg, NotifyEvent, raw as rawmod};`

PR #349, opened 2021-05-05
[feature-search-json] Review before merge to master

rafie:


gkorland (changes requested):

gkorland (changes requested):

src/c_api.rs
```diff
@@ -0,0 +1,612 @@
...
+        Value::Bool(ref b) => {
+            if *b {
+                return Some(1.0);
+            } else {
+                return Some(0.0);
+            }
+        }
+        _ => {}
+    }
+    return None;
```
gavrie, line 307: It would be more idiomatic in Rust to:
1. Get rid of all those `return`s and use a single expression instead.
2. There's no need for these explicit `ref`s in the `match` statement, the compiler infers those by itself.
3. When there's an `if` inside a `match` it can often me combined with the `match`.
4. Standard combinators like `ok()` can convert between `Result` and `Option`, no need for another `match` there.

So all of that code can be simplified to this:
```suggestion
    match value {
        Value::Number(n) if n.is_f64() => n.as_f64(),
        Value::String(s) => c_double::from_str(s.as_str()).ok(),
        Value::Bool(b) => Some(if *b { 1.0 } else { 0.0 }),
        _ => None,
    }
```

src/c_api.rs
```diff
@@ -0,0 +1,612 @@
...
+fn get_bool_value(value: &Value) -> Option<c_int> {
+    match value {
+        Value::Bool(ref b) => return if *b { Some(1) } else { Some(0) },
+        Value::Number(n) => {
+            if let Some(i) = n.as_i64() {
+                return if i != 0 { Some(1) } else { Some(0) };
+            } else if let Some(f) = n.as_f64() {
+                return if f != 0.0 { Some(1) } else { Some(0) };
+            }
+        }
```
gavrie, line 319: Can be simplified as above.

src/c_api.rs
```diff
@@ -0,0 +1,612 @@
...
+
+//
+// structs
+//
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct RedisModuleCtx {
+    _unused: [u8; 0],
+}
```
gavrie, line 29: This is duplicate code from `redismodule-rs`, this block can be removed.
```suggestion
```

src/c_api.rs
```diff
@@ -0,0 +1,612 @@
...
+#[allow(non_camel_case_types)]
+pub enum RedisReturnCode {
+    REDISMODULE_OK = 0,
+    REDISMODULE_ERR = 1,
+}
+
+//---------------------------------------------------------------------------------------------
+
+pub struct JSONApiKey<'a> {
+    _key: RedisKeyWritable,
```
gavrie, line 52: Why is the `_key` field necessary if it's unused?

src/c_api.rs
```diff
@@ -0,0 +1,612 @@
...
+        return RedisReturnCode::REDISMODULE_ERR as c_int;
+    }
+
+    fn get_value(&self, path: *const c_char) -> Option<&Value> {
+        let path = unsafe { CStr::from_ptr(path).to_str().unwrap() };
+        if let Ok(ref value) = self.redis_json.get_first(path) {
+            Some(value)
+        } else {
+            None
+        }
```
gavrie, line 79:
```suggestion
        self.redis_json.get_first(path).ok()
```

src/c_api.rs
```diff
@@ -0,0 +1,612 @@
...
+    Array = 5,
+    Null = 6,
+}
+
+#[repr(C)]
+#[allow(non_camel_case_types)]
+pub enum RedisReturnCode {
+    REDISMODULE_OK = 0,
+    REDISMODULE_ERR = 1,
+}
```
gavrie, line 47: As above -- we already have this in `redismodule-rs`.

src/c_api.rs
```diff
@@ -0,0 +1,612 @@
...
+#[no_mangle]
+pub extern "C" fn JSONAPI_openKey<'a>(
+    ctx: *mut rawmod::RedisModuleCtx,
+    key_str: *mut rawmod::RedisModuleString,
+) -> JSONApiKeyRef<'a> {
+    match JSONApiKey::new_from_redis_string(ctx, key_str) {
+        Ok(key) => Box::into_raw(Box::new(key)) as JSONApiKeyRef,
+        _ => null_mut(),
+    }
+}
```
gavrie, line 128: It might be better to use `bindgen` (as in `redismodule-rs`) to create function stubs from the header file. Less code to maintain, easier to change in the future and less chances of errors.

gavrie (changes requested):

Cargo.toml
```diff
@@ -14,10 +14,14 @@ serde_json = "1.0"
 serde = "1.0"
 libc = "0.2"
 jsonpath_lib = { git="https://github.com/RedisJSON/jsonpath.git", branch="public-parser" }
-redis-module = { version="0.15", features = ["experimental-api"]}
+# redis-module = { version="0.15", features = ["experimental-api"]}
+redis-module = { git = "https://github.com/RedisLabsModules/redismodule-rs", features = ["experimental-api"] }
+lazy_static = "1.4.0"
+# readies-wd40 = { path = "deps/readies/wd40" }
```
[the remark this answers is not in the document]
rafie: This is for debugging (enabling breakpoints when running under `gdb`). I think we can keep it.

src/c_api.rs
oshadmi: Agree, but I suggest to switch to `bindgen` in the next "phase" (first stabilize and merge to master)

src/c_api.rs
oshadmi: `RedisReturnCode` was newly introduced here 😊 

`REDISMODULE_OK` and `REDISMODULE_ERR` were already used in `redismodule-rs`
in `enum Status`
The values were generated by `bindgen` from `redismodule.h`
```
pub const REDISMODULE_OK: isize = 0;
pub const REDISMODULE_ERR: isize = 1;
```
Will use `Status` and get rid of `RedisReturnCode`

src/c_api.rs
oshadmi: It is needed for life-cycle management:
It will be dropped (and inner key will be closed) when `JSONAPI_closeKey` is called.
No need to use this field for other purposes.

gavrie (approved):

Cargo.toml
```diff
@@ -14,10 +14,14 @@ serde_json = "1.0"
 serde = "1.0"
 libc = "0.2"
 jsonpath_lib = { git="https://github.com/RedisJSON/jsonpath.git", branch="public-parser" }
-redis-module = { version="0.15", features = ["experimental-api"]}
+# redis-module = { version="0.15", features = ["experimental-api"]}
+redis-module = { git = "https://github.com/RedisLabsModules/redismodule-rs", features = ["experimental-api"] }
```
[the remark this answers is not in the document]
oshadmi: 👍🏼

src/c_api.rs
```diff
@@ -0,0 +1,612 @@
...
+    if !key.is_null() {
+        let key = unsafe { &mut *key };
+        if let Some(value) = key.get_value(path) {
+            if let Some(v) = get_int_value(value) {
+                unsafe { *val = v };
+                return RedisReturnCode::REDISMODULE_OK as c_int;
+            }
+        }
+    }
+    return RedisReturnCode::REDISMODULE_ERR as c_int;
```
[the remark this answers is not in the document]
oshadmi: 👍🏼

src/c_api.rs
```diff
@@ -0,0 +1,612 @@
...
+#[no_mangle]
+pub extern "C" fn JSONAPI_getDouble(json: JSONApiPathRef, val: *mut c_double) -> c_int {
+    if !json.is_null() {
+        let json = unsafe { &mut *json };
+        if let Some(v) = get_double_value(json.value) {
+            unsafe { *val = v };
+            return RedisReturnCode::REDISMODULE_OK as c_int;
+        }
+    }
+    return RedisReturnCode::REDISMODULE_ERR as c_int;
```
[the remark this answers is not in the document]
oshadmi: 👍🏼

src/c_api.rs
```diff
@@ -0,0 +1,612 @@
...
+    if !key.is_null() {
+        let key = unsafe { &mut *key };
+        if let Some(value) = key.get_value(path) {
+            if let Some(v) = get_double_value(value) {
+                unsafe { *val = v };
+                return RedisReturnCode::REDISMODULE_OK as c_int;
+            }
+        }
+    }
+    return RedisReturnCode::REDISMODULE_ERR as c_int;
```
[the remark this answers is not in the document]
oshadmi: 👍🏼

src/c_api.rs
```diff
@@ -0,0 +1,612 @@
...
+    if !key.is_null() {
+        let key = unsafe { &*key };
+        if let Some(value) = key.get_value(path) {
+            if let Some(v) = get_bool_value(value) {
+                unsafe { *val = v };
+                return RedisReturnCode::REDISMODULE_OK as c_int;
+            }
+        }
+    }
+    return RedisReturnCode::REDISMODULE_ERR as c_int;
```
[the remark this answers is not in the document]
oshadmi: 👍🏼

src/c_api.rs
```diff
@@ -0,0 +1,612 @@
...
+            if let Ok(s) = CString::new(from_str) {
+                unsafe {
+                    *str = s.as_bytes_with_nul().as_ptr() as *const c_char;
+                    *len = s.as_bytes().len();
+                }
+                self.cstr_val = Some(s);
+                return RedisReturnCode::REDISMODULE_OK as c_int;
+            }
+        }
+        return RedisReturnCode::REDISMODULE_ERR as c_int;
```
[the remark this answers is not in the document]
oshadmi: 👍🏼

src/c_api.rs
oshadmi: 👍🏼

src/c_api.rs
oshadmi: 👍🏼

src/c_api.rs
oshadmi: 👍🏼

src/redisjson.rs
```diff
@@ -509,6 +532,7 @@ pub mod type_methods {
     pub unsafe extern "C" fn rdb_save(rdb: *mut raw::RedisModuleIO, value: *mut c_void) {
         let json = &*(value as *mut RedisJSON);
         raw::save_string(rdb, &json.data.to_string());
+        raw::save_unsigned(rdb, 0);
```
gkorland, line 535: @rafie @oshadmi why do you store this fake 0?

src/redisjson.rs
```diff
@@ -490,6 +509,10 @@ pub mod type_methods {
             }
             3 => {
                 let data = raw::load_string(rdb);
+                if raw::load_unsigned(rdb) > 0 {
+                    raw::load_string(rdb);
+                    raw::load_string(rdb);
+                };
```
gkorland, line 515: @rafie @oshadmi these lines are only needed on version 2

gkorland (changes requested):

src/redisjson.rs
oshadmi: I will remove it (I think it was before we had db ver `3`)

src/redisjson.rs
oshadmi: @gkorland It was added in PR [#333](https://github.com/RedisJSON/RedisJSON/pull/333)
And merged to here from master.
I will remove it.

gkorland (approved):

oshadmi (approved):
👏🏼 👍🏼

PR #324, opened 2021-04-24
Fix json.arrappend performance

gkorland:
fix #295

PR #352, opened 2021-05-07
Fixes for CI benchmarks: iteration 2

filipecosta90:
- Including module artifact check on benchmarks. 
- Included extra dev docs explaining benchmark triggering.
- Increases the JSON.ARRAPPEND test duration given the major regression ( fixed by #324 ) is now merged.
- Fixes some wrong benchmark names 
- Make usage of readies/bin/getterraform and remove the explicit code for fetching terraform from repo


rafie:
@filipecosta90 I fixed the rust installation issue. However, I'm wary of the build cache (esp. how it's used here). I think we should stop using it. WDYT?

.circleci/config.yml
```diff
@@ -364,7 +365,7 @@ workflows:
           requires:
             - deploy_release
       - performance-automation:
-          <<: *on-integ-and-version-tags
+          <<: *not-on-integ-branch
```
rafie, line 368: shouldn't that be `on-integ-and-version-tags`?

filipecosta90: you're absolutly right @rafie . addressing it

filipecosta90: @rafie fixed

rafie (approved):

PR #358, opened 2021-05-12
CLang Sanitizer support in Makefile

rafie:


tests/pytest/tests.sh
```diff
@@ -121,6 +121,7 @@ if [[ ! -z $TEST ]]; then
 		RLTEST_ARGS+=" -s"
 		export BB=${BB:-1}
 	fi
+	export RUST_BACKTRACE=1
```
oshadmi, line 124: Consider using
`RUST_BACKTRACE=full`

(If I understand correctly, when `RUST_BACKTRACE=1`, you also get a hint telling you about `full` backtrace:
> note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

)

PR #361, opened 2021-05-19
JSON CAPI - return unquoted string value and add strict option

oshadmi:


oshadmi:
@rafie This change is too drastic, e.g.,  `test_json:testSearchUpdatedContent` fails:
`[1L, ['labelT', 'rex']] == [1L, ['labelT', '"rex"']]`

Need to unquote strings only in specific cases - not clear which.

oshadmi:
Replaced by PR #372

ashtul:
Can we get this back?

PR #365, opened 2021-05-20
Avoid premature key notification/replication

oshadmi:


oshadmi:
Redundant (done in other PR)

Issue #342, opened 2021-05-04
Check for Redis crash report on panic 


gkorland:

oshadmi:
Need to also consider how will stack trace be printed?

In release build, which is using optimizations, we may need to disable the optimization of removing frame pointers.
And we need to keep symbols.
For release, let's try build flags:
`RUSTFLAGS="-g -C force-frame-pointers=yes" cargo build --all --all-targets --release`

(the `-g` [flag](https://doc.rust-lang.org/rustc/command-line-arguments.html#option-g-debug) is an alias for `-C debuginfo=2`

oshadmi:
Also review environment variable `RUST_BACKTRACE=full`
And [panic hooks](https://doc.rust-lang.org/std/panic/fn.set_hook.html)
And [PanicInfo](https://doc.rust-lang.org/std/panic/struct.PanicInfo.html)

By default, panic should unwind and not abort:
https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/aborting-on-panic.html
https://doc.rust-lang.org/book/ch09-01-unrecoverable-errors-with-panic.html

oshadmi:
Created a fake panic and with debugInfo flag, trace includes `json_set` and `Vec index`.


![image](https://user-images.githubusercontent.com/76992134/117013940-fe1b4900-acf8-11eb-8fcf-a7bcfd3b65ee.png)

oshadmi:
With default build flags - seems not enough information
```

stack backtrace:
   0:     0x7f652282d540 - std::backtrace_rs::backtrace::libunwind::trace::h25e12e0d899beba0
                               at /rustc/cb75ad5db02783e8b0222fee363c5f63f7e2cf5b/library/std/src/../../backtrace/src/backtrace/libunwind.rs:90:5
   1:     0x7f652282d540 - std::backtrace_rs::backtrace::trace_unsynchronized::h70e61195d6ae3df6
                               at /rustc/cb75ad5db02783e8b0222fee363c5f63f7e2cf5b/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x7f652282d540 - std::sys_common::backtrace::_print_fmt::hba93ab80d779695a
                               at /rustc/cb75ad5db02783e8b0222fee363c5f63f7e2cf5b/library/std/src/sys_common/backtrace.rs:67:5
[...]

PR #372, opened 2021-05-26
JSON C API cleanup/refactor per review

oshadmi:


src/c_api.rs
```diff
@@ -152,12 +148,29 @@ pub extern "C" fn JSONAPI_getAt(
...
         }
     }
+    Status::Err as c_int
+}
+
+#[no_mangle]
+pub extern "C" fn JSONAPI_getType(json: JSONApiPathRef, jtype: *mut c_int) -> c_int {
+    if !json.is_null() {
+        let json = unsafe { &*json };
+        if !json.is_null() {
```
MeirShpilraien, line 167: Why another is_null check is require?

src/include/rejson_api.h
```diff
@@ -21,52 +21,54 @@ typedef const void *RedisJSONKey;
...
+  /* RedisJSON value functions
+   * Return REDISMODULE_OK if RedisJSON is of the correct JSONType,
+   * else REDISMODULE_ERR is returned
+   * */
+
+  // Return the length of Object/Array
+  int (*getLen)(RedisJSON json, size_t *count);
+
+  // Return the JSONType
+  int (*getType)(RedisJSON json, JSONType *type);
```
MeirShpilraien, line 45: Can getType ever failed? If not then why not returning the type as the return value of the function?

src/include/rejson_api.h
```diff
@@ -21,52 +21,54 @@ typedef const void *RedisJSONKey;
...
+
+  // Return int value from a Numeric field
+  int (*getInt)(RedisJSON json, long long *integer);
```
MeirShpilraien, line 48: If we will define that those function asserts that the type is correct we can make the API simpler and return the actual value (not via out parameters)

src/include/rejson_api.h
```diff
@@ -21,52 +21,54 @@ typedef const void *RedisJSONKey;
...
+  int (*getIntFromKey)(RedisJSONKey key, const char *path, long long *integer);
```
MeirShpilraien, line 49: Didn't we say we will remove those?

MeirShpilraien (commented):
:+1:
Let me know what you think about the comments

src/include/rejson_api.h
oshadmi: @MeirShpilraien Yes, will be done later - remove the `get<type>FromKey` APIs and remain with the `get<type>` APIs

src/include/rejson_api.h
oshadmi: @MeirShpilraien 
`getInt` may fail if path leads to a none numeric field - so panicking in this case seems too severe. WDYT?
`getType` - agreed in previous comment.

src/include/rejson_api.h
oshadmi: @MeirShpilraien Can fail only if `null` is passed to the API and as we discussed you recommend to panic
(make `RedisModule_Assert` avaialbe in Rust later. Maybe in issue #342 in panic hook)
So will change it to `JSONType (*getType)(RedisJSON json)`

src/c_api.rs
```diff
@@ -391,112 +373,72 @@ pub extern "C" fn JSONAPI_getBooleanFromKey(
...
-        let info = RedisJSON::get_type_and_size(self.value);
-        unsafe {
-            *jtype = info.0 as c_int;
-            if !count.is_null() {
-                *count = info.1;
-            }
+pub fn get_type_and_size(value: &Value, jtype: *mut c_int, count: *mut libc::size_t) {
+    let info = RedisJSON::get_type_and_size(value);
+    unsafe {
+        *jtype = info.0 as c_int;
```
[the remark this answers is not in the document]
oshadmi: @MeirShpilraien Right, `jtype` cannot be null.

src/c_api.rs
```diff
@@ -195,46 +200,40 @@ pub extern "C" fn JSONAPI_getStringFromKey(
         let key = unsafe { &mut *key };
         let path = unsafe { CStr::from_ptr(path).to_str().unwrap() };
         if let Ok(s) = key.redis_json.to_string(path, Format::JSON) {
```
rafie, line 202: This should be similar to `JSONAPI_getString` (that is, no `.to_string(path, Format::JSON)`.

src/c_api.rs
```diff
@@ -195,46 +200,40 @@ pub extern "C" fn JSONAPI_getStringFromKey(
...
-            if let Ok(s) = CString::new(s) {
-                unsafe {
-                    *str = s.as_bytes_with_nul().as_ptr() as *const c_char;
-                    *len = s.as_bytes().len();
-                }
-                key.cstr_val = Some(s);
-                return Status::Ok as c_int;
-            }
+            set_string(s.as_str(), str, len);
```
rafie, line 203: This, on the other hand, should be serializing (`.to_string(path, Format::JSON`).

src/c_api.rs
```diff
@@ -195,46 +200,40 @@ pub extern "C" fn JSONAPI_getStringFromKey(
...
+    ctx: *mut rawmod::RedisModuleCtx,
     str: *mut *mut rawmod::RedisModuleString,
 ) -> c_int {
     if !json.is_null() {
         let json = unsafe { &*json };
-        if let Some(ref p) = json.path {
-            if let Ok(res) = json.json_key.redis_json.to_string(p, Format::JSON) {
-                return json.json_key.create_rmstring(res.as_str(), str);
-            }
+        if let Ok(res) = serde_json::to_string(json) {
```
rafie, line 218: This, on the other hand, should be serializing (`.to_string(path, Format::JSON`).

src/c_api.rs
```diff
@@ -278,8 +265,6 @@ fn get_int_value(value: &Value) -> Option<c_long> {
 fn get_double_value(value: &Value) -> Option<c_double> {
     match value {
         Value::Number(n) if n.is_f64() => n.as_f64(),
-        Value::String(s) => c_double::from_str(s.as_str()).ok(),
```
rafie, line 281: Can we keep the string->double conversion here? It helps the API be less tight-ass.

oshadmi: @rafie OK.
Why only string->double? (Geo?)

src/c_api.rs
oshadmi: @rafie Right, we are serializing, but fixed to serialize directly from the current `Value` instead of traversing to it using `path`.

src/include/rejson_api.h
rafie: I'd like to keep the error control on the client side, especially with read operations.

src/include/rejson_api.h
```diff
@@ -21,52 +21,54 @@ typedef const void *RedisJSONKey;
...
+  RedisJSON (*get)(RedisJSONKey key, const char *path, JSONType *type);
+  RedisJSON (*getAt)(RedisJSON json, size_t index, JSONType *type);
+
+  /* RedisJSON value functions
+   * Return REDISMODULE_OK if RedisJSON is of the correct JSONType,
+   * else REDISMODULE_ERR is returned
+   * */
+
+  // Return the length of Object/Array
+  int (*getLen)(RedisJSON json, size_t *count);
```
rafie, line 42: Would `size_t (*getLen)(RedisJSON json)`, with 0 or -1 returned on a non aggregate/compound/string object, be out of the question?

rafie (approved):

src/include/rejson_api.h
oshadmi: @rafie OK, `getType` should not fail (only panics on null), and the rest of the APIs may return an error to the API client.

src/include/rejson_api.h
oshadmi: @rafie Currently `0` is returned for scalars (assuming caller can check the type, `-1` was avoided here)

PR #394, opened 2021-06-16
Move to RedisString

gkorland:


src/lib.rs
```diff
@@ -140,24 +140,24 @@ fn json_del(ctx: &Context, args: Vec<String>) -> RedisResult {
...
-            "XX" => set_option = SetOptions::AlreadyExists,
-            "FORMAT" => {
+        match s.try_as_str()? {
+            arg if arg.eq_ignore_ascii_case("NX") => set_option = SetOptions::NotExists,
+            arg if arg.eq_ignore_ascii_case("XX") => set_option = SetOptions::AlreadyExists,
+            arg if arg.eq_ignore_ascii_case("FORMAT") => {
                 format = Format::from_str(args.next_string()?.as_str())?;
             }
-            _ => break,
+            _ => break, // TODO shoudl return an error??
```
gavrie, line 160: It probably should.

src/lib.rs
```diff
@@ -523,22 +523,22 @@ fn json_arr_append(ctx: &Context, args: Vec<String>) -> RedisResult {
...
-    I: Iterator<Item = String>,
+    I: Iterator<Item = RedisString>,
 {
     if !value.is_array() {
         return Err(err_json(value, "array"));
     }
 
     let mut items: Vec<Value> = args
-        .map(|json| serde_json::from_str(&json))
+        .map(|json| serde_json::from_str(&json.into_string_lossy()))
```
gavrie, line 541: Is a lossy conversion what we want here? I know the Redis way is not to return errors, but can't this cause silent data corruption in case the input is not valid UTF-8? JSON should always be valid UTF-8.

src/lib.rs
oshadmi: Before commit 89b8a957e27e8f9a8e46b354f930937c6c331522 there seems to have been an error report `Err(RedisError::Str("ERR syntax error"))`

src/c_api.rs
```diff
@@ -82,18 +82,20 @@ impl<'a> JSONApiKey<'a> {
...
+        let key = ctx.open_key_writable(&key);
         JSONApiKey::new_from_key(key)
     }
 
     pub fn new_from_str(
         ctx: *mut rawmod::RedisModuleCtx,
         path: *const c_char,
     ) -> Result<JSONApiKey<'a>, RedisError> {
-        let ctx = Context::new(ctx);
         let path = unsafe { CStr::from_ptr(path).to_str().unwrap() };
```
oshadmi, line 95: This line not from this PR, but still would like to raise that since it is also related to UFT-8 and lossy conversion vs. error handling mentioned in this [comment](https://github.com/RedisJSON/RedisJSON/pull/394/files#r655953307)
`let path = unsafe { CStr::from_ptr(path).to_str().unwrap() };`
This would panic if invalid UTF-8 is passed here.

oshadmi: Since c_api.rs currently has no error reporting capability, we should detect conversion failure and the caller should return `null` on failure (assuming invalid UFT-8 can arrive from redis/from callers of `new_from_str`)

PR #397, opened 2021-06-18
Support RDB Short Read in RedisJSON (a.k.a diskless-load)

oshadmi:


src/redisjson.rs
```diff
@@ -506,6 +524,10 @@ pub mod type_methods {
         raw::save_string(rdb, &json.data.to_string());
     }
 
+    pub fn is_io_error(rdb: *mut raw::RedisModuleIO) -> bool {
```
gkorland, line 527: Should move to redismodule-rs

src/backward.rs
```diff
@@ -34,41 +39,70 @@ impl From<u64> for NodeType {
     }
 }
 
-pub fn json_rdb_load(rdb: *mut raw::RedisModuleIO) -> Value {
+pub fn json_rdb_load(rdb: *mut raw::RedisModuleIO) -> Result<Value, Error> {
     let node_type = raw::load_unsigned(rdb).into();
+    if is_io_error(rdb) {
+        return Err(Error::from("short read"));
```
gkorland, line 45: `Err(Error::from("short read"));` can be set as constant, repeats many times

src/backward.rs
```diff
@@ -34,41 +39,70 @@ impl From<u64> for NodeType {
...
             for _ in 0..len {
-                let t: NodeType = raw::load_unsigned(rdb).into();
+                let u = raw::load_unsigned(rdb);
+                if is_io_error(rdb) {
+                    return Err(Error::from("short read"));
+                }
+                let t: NodeType = u.into();
                 if t != NodeType::KeyVal {
-                    panic!("Can't load old RedisJSON RDB");
+                    return Err(Error::from("Can't load old RedisJSON RDB"));
```
gkorland, line 90: how will you be able to recover from it and not panic?

src/backward.rs
```diff
@@ -34,41 +39,70 @@ impl From<u64> for NodeType {
...
             for _ in 0..len {
-                v.push(json_rdb_load(rdb))
+                let nested = json_rdb_load(rdb)?;
+                v.push(nested);
             }
-            Value::Array(v)
+            Ok(Value::Array(v))
         }
-        NodeType::KeyVal => panic!("Can't load old RedisJSON RDB"),
+        NodeType::KeyVal => Err(Error::from("Can't load old RedisJSON RDB")),
```
gkorland, line 106: how will you be able to recover from it and not panic?

src/lib.rs
```diff
@@ -397,8 +399,48 @@ macro_rules! redis_json_module_create {(
...
+                Version {major: 255, minor: 255, patch: 255 } => true,
+                _ => false
+            }
+        }
+
         fn intialize(ctx: &Context, args: &Vec<RedisString>) -> Status {
             export_shared_api(ctx);
+            if check_minimal_version_for_short_read() {
+                // Enable RDB short read
+                unsafe {
```
gkorland, line 434: Should probably also move to redismodule-rs

src/redisjson.rs
```diff
@@ -470,21 +471,38 @@ pub mod type_methods {
...
                 let data = raw::load_string(rdb);
+                if is_io_error(rdb) {
+                    return 0 as *mut c_void;
+                }
                 // Backward support for modules that had AUX field for RediSarch
                 // TODO remove in future versions
-                if raw::load_unsigned(rdb) > 0 {
+                let u = raw::load_unsigned(rdb);
+                if is_io_error(rdb) {
+                    return 0 as *mut c_void;
```
gkorland, line 487: ptr::null_mut() I think is better than 0

src/redisjson.rs
```diff
@@ -516,12 +538,28 @@ pub mod type_methods {
         // TODO remove in future versions
         if (encver == 2 && when == raw::Aux::Before as i32) {
             let map_size = raw::load_unsigned(rdb);
+            if is_io_error(rdb) {
+                return Status::Err as i32;
```
gkorland, line 542: `return Status::Err as i32;`

Should be added as a method on Status to_i32().
We already have the other way around:
```impl From<c_int> for Status {
    fn from(v: c_int) -> Self {
        Status::from_i32(v).unwrap()
    }
}
```

src/backward.rs
oshadmi: when load fails, replica reports to master which should resend rdb

src/backward.rs
oshadmi: when load fails, replica reports to master which should resend rdb

Cargo.toml
```diff
@@ -14,8 +14,9 @@ bson = "0.14"
 serde_json = "1.0"
 serde = "1.0"
 libc = "0.2"
-jsonpath_lib = { git="https://github.com/RedisJSON/jsonpath.git", branch="generic_json_path" }
-redis-module = { version="0.22", features = ["experimental-api"]}
+jsonpath_lib = { git = "https://github.com/RedisJSON/jsonpath.git", branch = "generic_json_path" }
+#redis-module = { version="0.22", features = ["experimental-api"]}
+redis-module = { git = "https://github.com/RedisLabsModules/redismodule-rs.git", branch = "omer-short-read", features = ["experimental-api"] }
```
MeirShpilraien, line 19: Make sure to change that before merge ...

oshadmi: Sure, we need to bump redis-module version and use the new one here.

src/lib.rs
```diff
@@ -397,8 +400,45 @@ macro_rules! redis_json_module_create {(
             pre_command_function: $pre_command_function_expr,
         }
 
+        pub fn check_minimal_version_for_short_read(ctx: &Context) -> bool {
+            // Minimal versions: 6.2.5
```
MeirShpilraien, line 404: json do not have aux fields so you can support short read even before this version.

MeirShpilraien (commented):
Some small comments, we also need tests (as we have on search).

src/lib.rs
oshadmi: @MeirShpilraien So this would make `get_redis_version` in `redismodule-rs` unnecessary, right?

src/backward.rs
```diff
@@ -34,41 +38,48 @@ impl From<u64> for NodeType {
...
+            let n = raw::load_double(rdb)?;
+            Ok(Value::Number(Number::from_f64(n).unwrap()))
         }
-        NodeType::Integer => Value::Number(raw::load_signed(rdb).into()),
-        NodeType::Number => Value::Number(Number::from_f64(raw::load_double(rdb)).unwrap()),
         NodeType::String => {
-            let buffer = raw::load_string_buffer(rdb);
-            Value::String(buffer.to_string().unwrap())
+            let buffer = raw::load_string_buffer(rdb)?;
+            Ok(Value::String(buffer.to_string().unwrap()))
```
gavrie, line 59: Is this safe? What if the buffer is not a valid UTF-8 value -- will this crash?

system-setup.py
```diff
@@ -49,6 +49,7 @@ def common_last(self):
         self.pip_install("-r %s/tests/pytest/requirements.txt" % ROOT)
         self.pip_install("toml")
         self.pip_install("awscli")
+        self.pip_install("gevent")
```
gavrie, line 52: It would be cleaner to list the deps inside `requirements.txt` instead of hardcoding. You could add another separate `requirements.txt` file in addition to the one from the tests.

tests/pytest/common.py
```diff
@@ -0,0 +1,22 @@
...
+    def __init__(self, timeout):
+        self.timeout = timeout
+
+    def __enter__(self):
+        signal.signal(signal.SIGALRM, self.handler)
+        signal.setitimer(signal.ITIMER_REAL, self.timeout, 0)
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        signal.setitimer(signal.ITIMER_REAL, 0)
+        signal.signal(signal.SIGALRM, signal.SIG_DFL)
```
gavrie, line 19: The recommended style for writing context managers is to use the `contextlib.contextmanager` decorator instead of manually implementing these methods.

tests/pytest/test_short_read.py
```diff
@@ -0,0 +1,518 @@
...
+def random_json_value(nesting_level, max_nesting_level):
+    # Favor a top-level object/array
+    if nesting_level == 0 and nesting_level != max_nesting_level and not random.randint(0, 5):
+        kind = random.randint(6, 7)
+    else:
+        kind = random.randint(1, 7)
+    if kind == 1:
+        # null
+        return None
+    elif kind == 2:
```
gavrie, line 111: Instead of using these magic numbers, consider defining an `Enum` (or even a simple list of constants) of the different types and selecting one of its values randomly (using `random.choice`). That would be much more readable.

tests/pytest/test_short_read.py
```diff
@@ -0,0 +1,518 @@
...
+        if isinstance(value, bytes):
+            return value.decode('utf-8')
+        else:
+            return value
+
+    def readline(self):
+        return self.decoder(self.sockf.readline())
+
+    def send_bulk_header(self, data_len):
+        self.sockf.write(self.encoder('$%d\r\n' % data_len))
```
gavrie, line 206: As a rule, byte strings in Python (as well as in Rust...) should have the `b'...'` prefix. Plain strings are considered Unicode in Python 3, which is not what you want.

gavrie (changes requested):

src/backward.rs
oshadmi: I wasn't able to create JSON key with invalid UFT-8 (also tried with ReJSON 1.0.8) so was not able to create such an invalid rdb.
I changed to a more graceful failure than panicking since now we return a `Result`

oshadmi (commented):
Did some fixes - Thanks!

src/lib.rs
```diff
@@ -399,6 +402,15 @@ macro_rules! redis_json_module_create {(
 
         fn intialize(ctx: &Context, args: &Vec<RedisString>) -> Status {
             export_shared_api(ctx);
+            unsafe {
```
MeirShpilraien, line 405: Shouldn't we expose it as a function of the Context on redismodule-rs?

tests/pytest/test_short_read.py
```diff
@@ -0,0 +1,524 @@
...
+                    return False
+            else:
+                if not os.path.exists(path) or os.path.getsize(path) == 0:
+                    return False
+    return True
+
+
+def rand_name(k):
+    # rand alphabetic string with between 2 to k chars
+    return ''.join([chr(random.randint(ord('a'), ord('z'))) for _ in range(0, random.randint(2, max(2, k)))])
```
MeirShpilraien, line 57: We should use some random generate seed and log it so if the test failed it will be easier to reproduce it.

tests/pytest/test_short_read.py
```diff
@@ -0,0 +1,524 @@
...
+        # Close during replica is waiting for more RDB data (so replica will re-connect to master)
+        conn.close()
+
+        # Make sure replica did not crash
+        res = env.cmd('PING')
+        env.assertEqual(res, True)
+        conn = shardMock.GetConnection(timeout=3)
+        env.assertNotEqual(conn, None)
+
+        if is_shortread:
```
MeirShpilraien, line 507: It looks like you are checking the same thing on the if and else, am I missing something?

src/redisjson.rs
```diff
@@ -505,27 +519,35 @@ pub mod type_methods {
         raw::save_string(rdb, &json.data.to_string());
     }
 
-    #[allow(non_snake_case, unused)]
     pub unsafe extern "C" fn aux_load(rdb: *mut raw::RedisModuleIO, encver: i32, when: i32) -> i32 {
```
MeirShpilraien, line 522: Can't we drop the aux load? it was never released in any official version no?
@gkorland ^^^

MeirShpilraien: If we can not drop it then we can only support a short read from version 6.2.5 and we need to check for it.

MeirShpilraien (commented):
Look good, some comments

oshadmi (commented):
👍🏼 Thanks for the review!

src/redisjson.rs
oshadmi: Looking at issue #318 (this [comment](https://github.com/RedisJSON/RedisJSON/issues/318#issuecomment-824620167)), it seems like we need `aux_load`

oshadmi: After discussion with @gkorland and @MeirShpilraien, It was decided to drop `aux_load` (no need to support unofficial releases, users had plenty of time to migrate)

MeirShpilraien (dismissed):

Cargo.toml
```diff
@@ -15,8 +15,7 @@ serde_json = "1.0"
 serde = "1.0"
 libc = "0.2"
 jsonpath_lib = { git = "https://github.com/RedisJSON/jsonpath.git", branch = "generic_json_path" }
-#redis-module = { version="0.22", features = ["experimental-api"]}
-redis-module = { git = "https://github.com/RedisLabsModules/redismodule-rs.git", branch = "omer-short-read", features = ["experimental-api"] }
+redis-module = { version="0.23", features = ["experimental-api"]}
```
MeirShpilraien, line 18: We need to also bump it on hdt...

MeirShpilraien (approved):

PR #398, opened 2021-06-20
avoid path clone when not needed

gkorland:
This should reduce the useless clone of the path when there is no need for backward compatibility with RedisJSON 1.0.x

### DIFF
--- a/src/redisjson.rs
+++ b/src/redisjson.rs
@@ -46,23 +46,30 @@ impl Format {
 /// Backwards compatibility convertor for RedisJSON 1.x clients
 ///
 pub struct Path {
-    pub path: String,
-    pub fixed: String,
+    original_path: String,
+    fixed_path: Option<String>,
 }
 
 impl Path {
     pub fn new(path: String) -> Path {
-        let mut fixed = path.clone();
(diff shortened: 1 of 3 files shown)


src/redisjson.rs
```diff
@@ -46,23 +46,33 @@ impl Format {
 /// Backwards compatibility convertor for RedisJSON 1.x clients
 ///
 pub struct Path {
-    pub path: String,
-    pub fixed: String,
+    original_path: String,
+    fixed_path: Option<String>,
```
oshadmi, line 50: Just a thought,
Can we even consider (also later) avoiding the `fixed_path` member and calculate it when calling `get_path` which will return the moved String? (either original or fixed if needed)
(assuming that once we pass it to JSONPath selector we don't need it anymore)
This will shorten the lifetime of the cloned/mutated strings to the minimum, but it will probably have none to very small impact (unless `paths: &mut Vec<Path>` vector in method `to_json` is big)

oshadmi (approved):
LGTM👍🏼

oshadmi (approved):
👍🏼 LGTM

MeirShpilraien (approved):

PR #408, opened 2021-06-29
add multi result support of json.get

MeirShpilraien:


tests/pytest/test.py
```diff
@@ -829,6 +829,12 @@ def testIssue_80(env):
     r.execute_command('JSON.GET', 'test', '$.[?(@.code=="2")]')
 
 
+def testMultiPathResults(env):
+    env.expect("JSON.SET", "k", '$', '[1,2,3]').ok()
+    env.expect("JSON.GET", "k", '$[*]').equal(['1', '2', '3'])
+    env.expect("JSON.SET", "k", '$', '{"a":[1,2,3],"b":["c", "d", "e"],"c":"k"}').ok()
+    env.expect("JSON.GET", "k", '$.*[0,2]').equal(['1', '3', '"c"', '"e"'])
+
```
oshadmi, line 837: 1) What would be the result of the same `GET` command if the JSON would initially be
`{"a":[1,2,3],"b":["c", "d", "e"],"c":["k"]}`
(the third element would be an array but with only a single element)
Should it be `['1', '3', '"c"', '"e"', "k"]` ?
2) Do we want to increase the coverage of multi-paths (more wildcards, and unions, also unions by name, not only by indices, and array slices and recursive descents)
3) Do we want to also test the legacy mode? Some sanity test?
e.g., `env.expect("JSON.GET", "k", '.[*]').equal('1')`

oshadmi (commented):
👌🏼 Few questions

tests/pytest/test.py
MeirShpilraien: 1. yep, this is the json path specification, you can use condition to filter out this result if you want (base on the array size for example)
2. Yes, but not as part of this PR, this PR purpose to give initial support.
3. Good point, will add it.

src/redisjson.rs
```diff
@@ -240,22 +248,33 @@ impl RedisJSON {
...
             let mut selector = jsonpath_lib::selector(&self.data);
             // TODO: Creating a temp doc here duplicates memory usage. This can be very memory inefficient.
             // A better way would be to create a doc of references to the original doc but no current support
             // in serde_json. I'm going for this implementation anyway because serde_json isn't supposed to be
             // memory efficient and we're using it anyway. See https://github.com/serde-rs/json/issues/635.
-            temp_doc = Value::Object(paths.drain(..).fold(Map::new(), |mut acc, path| {
+            let temp_doc = Value::Object(paths.drain(..).fold(Map::new(), |mut acc, path| {
                 let value = match selector(&path.fixed) {
                     Ok(s) => match s.first() {
                         Some(v) => v,
```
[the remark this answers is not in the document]
MeirShpilraien: We decided with @emmanuel-keller  not to decide about it now and just leave it as is

oshadmi (approved):
👍🏼

PR #417, opened 2021-07-04
change llapi to support multiple jsonpath results

MeirShpilraien:


src/c_api.rs
```diff
@@ -407,26 +338,29 @@ pub fn json_api_get<M: Manager>(
         return null();
```
oshadmi, line 338: Do you think it would matter (slightly improve performance) if we take the optimistic approach, and assume less branch-none-taken will occur when we start by testing for the successful option that the path exists,
And only otherwise handle the failed option that the path does not exist and return null?

src/c_api.rs
```diff
@@ -391,12 +305,29 @@ fn json_api_get_type_internal<V: SelectValue>(v: &V) -> JSONType {
...
 }
 
-pub fn json_api_get<M: Manager>(
-    _: M,
-    key: *mut c_void,
-    path: *const c_char,
-    jtype: *mut c_int,
-) -> *const c_void {
+pub fn json_api_next<M: Manager>(_: M, iter: *mut c_void) -> *const c_void {
+    let iter = unsafe { &mut *(iter as *mut ResultsIterator<M::V>) };
```
oshadmi, line 309: should check for `null`?

src/c_api.rs
```diff
@@ -391,12 +305,29 @@ fn json_api_get_type_internal<V: SelectValue>(v: &V) -> JSONType {
...
+    } else {
+        let res = iter.results[iter.pos] as *const M::V as *const c_void;
+        iter.pos = iter.pos + 1;
+        res
+    }
+}
+
+pub fn json_api_len<M: Manager>(_: M, iter: *const c_void) -> c_ulong {
+    let iter = unsafe { &*(iter as *mut ResultsIterator<M::V>) };
+    iter.results.len() as c_ulong
```
oshadmi, line 321: Rust `Vec` index is of type `usize`, user is passing `size_t`, such as in `json_api_get_at`
And now we also add `c_ulong`
Do we need it? Don't we want to stick to `libc` types?
(currently I see in libc
`pub type size_t = usize;`
But maybe this may change on other platforms)

oshadmi (commented):
👍🏼
A few questions.

src/include/rejson_api.h
```diff
@@ -27,12 +28,13 @@ typedef struct RedisJSONAPI_V1 {
...
 
-  /* RedisJSON functions
-   * Return NULL if path/index does not exist
-   * `type` is an optional out parameter returning the JSONType (can be NULL)
-   **/
-  RedisJSON (*get)(RedisJSONKey key, const char *path, JSONType *type);
-  RedisJSON (*getAt)(RedisJSON json, size_t index, JSONType *type);
+  ResultsIterator* (*get)(RedisJSONKey key, const char *path);
+  
+  RedisJSON* (*next)(ResultsIterator* iter);
```
rafie, line 33: `RedisJSON*` -> `RedisJSON` (it's an opaque void*)

src/c_api.rs
```diff
@@ -102,34 +107,20 @@ pub extern "C" fn JSONAPI_closeKey(json: *mut c_void) {
...
 }
 
-fn json_api_get_at<M: Manager>(
-    _: M,
-    json: *const c_void,
-    index: libc::size_t,
-    jtype: *mut c_int,
-) -> *const c_void {
+fn json_api_get_at<M: Manager>(_: M, json: *const c_void, index: libc::size_t) -> *const c_void {
     let json = unsafe { &*(json as *const M::V) };
```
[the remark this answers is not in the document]
MeirShpilraien: The user should not pass NULL

src/c_api.rs
```diff
@@ -391,12 +305,29 @@ fn json_api_get_type_internal<V: SelectValue>(v: &V) -> JSONType {
...
+        null_mut()
+    } else {
+        let res = iter.results[iter.pos] as *const M::V as *const c_void;
+        iter.pos = iter.pos + 1;
+        res
+    }
+}
+
+pub fn json_api_len<M: Manager>(_: M, iter: *const c_void) -> c_ulong {
+    let iter = unsafe { &*(iter as *mut ResultsIterator<M::V>) };
```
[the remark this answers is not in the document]
MeirShpilraien: Same reply

src/c_api.rs
```diff
@@ -391,12 +305,29 @@ fn json_api_get_type_internal<V: SelectValue>(v: &V) -> JSONType {
...
+    iter.results.len() as c_ulong
+}
+
+pub fn json_api_free_iter<M: Manager>(_: M, iter: *mut c_void) {
+    unsafe {
+        Box::from_raw(iter as *mut ResultsIterator<M::V>);
```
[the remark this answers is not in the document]
MeirShpilraien: Same

src/c_api.rs
MeirShpilraien: I thought size_t == usize ...

src/c_api.rs
MeirShpilraien: This does not verify that the path exists, it verifies that the given path is a valid JSON path...

src/c_api.rs
```diff
@@ -1,10 +1,11 @@
+use libc::size_t;
 use std::ffi::CString;
 use std::os::raw::{c_double, c_int, c_long};
 use std::ptr::{null, null_mut};
 use std::slice;
 use std::{
     ffi::CStr,
-    os::raw::{c_char, c_void},
+    os::raw::{c_char, c_ulong, c_void},
```
oshadmi, line 8: `c_ulong` not needed?

oshadmi (dismissed):
👍🏼 Looking good!

src/c_api.rs
MeirShpilraien: Fixed

oshadmi (approved):
👍🏼 Thanks!

rafie (dismissed):

src/include/rejson_api.h
```diff
@@ -17,22 +17,21 @@ typedef enum JSONType {
     JSONType__EOF
 } JSONType;
 
-typedef const void *RedisJSONKey;
-typedef const void *RedisJSON;
+typedef const void* RedisJSON;
+typedef const void* ResultsIterator;
```
ashtul, line 21: This is generic. How about `JSONIterator`?

ashtul (changes requested):

ashtul (approved):
Looks good

PR #418, opened 2021-07-04
optimize set root by caching key value

MeirShpilraien:


src/manager.rs
```diff
@@ -214,6 +231,7 @@ impl KeyHolderWrite {
             }
             None => {
                 self.key.delete()?;
+                self.val = None;
```
oshadmi, line 234: Should we clear the cached value before calling `delete()`?
In case a keyspace notification would trigger an API call that might use the cached value although it was deleted?
The disadvantage of clearing the cache first, is that if delete fails, the cache was cleared in vain, but this is less severe in my opinion, than caching a stale value.

src/manager.rs
```diff
@@ -88,65 +88,65 @@ fn err_json(value: &Value, expected_value: &'static str) -> Error {
 pub struct KeyHolderWrite {
     key: RedisKeyWritable,
     key_name: String,
+    val: Option<*mut RedisJSON>,
```
oshadmi, line 91: Why pointer? So it doesn't get dropped?

oshadmi (commented):
👍🏼 Few questions.

src/manager.rs
MeirShpilraien: Good point :+1:

src/manager.rs
MeirShpilraien: Changed to reference with lifetime after a few hours of fighting the compiler :)

src/manager.rs
MeirShpilraien: Though if a key space notification will cause opening the key then it will be another write holder with its own cache. But I agree its safer this way...

src/c_api.rs
```diff
@@ -86,7 +86,13 @@ pub extern "C" fn JSONAPI_openKey<'a>(
         let bytes = slice::from_raw_parts(bytes as *const u8, len);
         String::from_utf8_lossy(bytes).into_owned()
     };
-    json_api_open_key_internal(RedisJsonKeyManager, ctx, &key) as *mut c_void
+    json_api_open_key_internal(
+        RedisJsonKeyManager {
+            phantom: std::marker::PhantomData,
```
oshadmi, line 91: Adding a use declaration would shorten the code (occurs many times)
from `std::marker::PhantomData` to `PhantomData`

MeirShpilraien: Fixed

oshadmi (approved):
👍🏼

PR #421, opened 2021-07-07
change llapi to be declare on the initialize macro so it can be used also on hdt

MeirShpilraien:


oshadmi (dismissed):
👍🏼 LGTM 👀 
(hoping that `pre_command_function_expr` that is now added to each C_API call does not incur a significant cost)

oshadmi (approved):
👍🏼 LGTM

PR #422, opened 2021-07-07
Set vec capcity to 1 by default

gkorland:


src/commands.rs
```diff
@@ -453,7 +453,8 @@ pub fn command_json_get<M: Manager>(
     let mut args = args.into_iter().skip(1);
     let key = args.next_arg()?;
 
-    let mut paths: Vec<Path> = vec![];
+    // Set Capcity to 1 assumiung the common case has one path
+    let mut paths: Vec<Path> = Vec::with_capacity(1);
```
oshadmi, line 457: If indeed the common usage is a single path arg - we could benefit from this, since
Looking at [raw_vec.rs](https://github.com/rust-lang/rust/blob/master/library/alloc/src/raw_vec.rs), before this change, `MIN_NON_ZERO_CAP` would be 4 (since `push` will call `reserve(1)` when the initial `len` and `capacity` are zero in [mod.rs](https://github.com/rust-lang/rust/blob/master/library/alloc/src/vec/mod.rs))

oshadmi (approved):
👍🏼

PR #432, opened 2021-07-26
Issue Template

chayim:
I've noticed that we receive a lot of issues that are missing information. This means we spend time going back-and-forth, requesting information before we can analyze the bugs.

I'm proposing we add this issue template across our projects (potentially tailored).  The syntax to modify this is [here](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms).  Feedback wanted.


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,64 @@
...
+    id: redis_module_version
+    attributes:
+      label: Redis Module Version
+      description: What version of this redis module was installed?
+      placeholder: latest
+
+  - type: input
+    id: was_redis_docker
+    attributes:
+      label: Docker Instllatio
```
oshadmi, line 42:
```suggestion
      label: Docker Installation
```

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,64 @@
...
+
+  - type: input
+    id: redis_version
+    attributes:
+      label: Redis Server Version
+      description: What version of redis-server are you running (redis-server --version)
+      placeholder: latest
+
+  - type: input
+    id: redis_module_version
```
oshadmi, line 33: Could be several modules - so user can copy-paste? Maybe add a 'name' attribute for it?

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,64 @@
...
+    id: installmethod
+    attributes:
+      label: Docker installation
+      description: Check this box, if you installed the release via a docker.
+      options:
+        - label: I installed the dockerhub release
+
+  - type: textarea
+    id: whathappened
+    label: What happened?
```
oshadmi, line 56: Do we want to have checkboxes for scary stuff, like crash, loss-of-data, hang/not-responding, so it could be easily found?

oshadmi (commented):
👍🏼 Cool!

.github/ISSUE_TEMPLATE.md
oshadmi: And what about milestone?

chayim: I think milestone is covered by version. We could definitely do the checkbox data loss thing, but given that someone is going to use words to describe _something_ I looked at it like duplication.

.github/ISSUE_TEMPLATE.md
chayim: So this redis-module-version bit was specific to this module. But yes, I could see a version(s) of the thing being reported against.

rafie (approved):

PR #467, opened 2021-10-04
Return count of cleared results

gkorland:


src/commands.rs
```diff
@@ -1137,12 +1137,12 @@ pub fn command_json_clear<M: Manager>(
 
     let paths = find_paths(path, root, |_v| true)?;
```
oshadmi, line 1138: Do we want to align with other commands that handle only matching keys?
```suggestion
    let paths = find_paths(path, root, |v| {
        v.get_type() == SelectValueType::Array || v.get_type() == SelectValueType::Object
    })?;
```

tests/pytest/test.py
```diff
@@ -485,6 +485,12 @@ def testClear(env):
     # Make sure only appropriate content (obj and arr) was cleared - and that errors were printed for inappropriate content (string and numeric)
```
oshadmi, line 485:
```suggestion
    # Make sure only appropriate content (obj and arr) was cleared
```

oshadmi (approved):
LGTM👍🏼

PR #493, opened 2021-10-28
enable fat optimizations on linking

MeirShpilraien:


MeirShpilraien:
Not merging for now as it does not seem to improve the performance.

PR #487, opened 2021-10-27
Merge multipath PRs (#473 and #477)

oshadmi:
#473 
#477

PR #496, opened 2021-10-29
Multipath - GET command always returning an array

oshadmi:
Return an array of results, even when only a single path is given
(not with legacy path, only if at least one path begins with a `$`)

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -205,18 +205,12 @@ impl<'a, V: SelectValue> KeyValue<'a, V> {
         newline: Option<&str>,
         space: Option<&str>,
     ) -> Result<RedisValue, Error> {
-        if paths.len() > 1 {
-            let mut res: Vec<Vec<&V>> = vec![];
-            for path in paths {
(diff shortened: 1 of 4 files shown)


MeirShpilraien (approved):
@emmanuel-keller approved it?

oshadmi:
> @emmanuel-keller approved it?

Not yet.
@emmanuel-keller Do you approve this change?
Also mentioned [here](https://github.com/RedisJSON/RedisJSON/pull/487/files#r738150854)

emmanuelkeller:
@MeirShpilraien I would not do that, mainly for the following reason.

JSON.GET with only one path is probably the most used command.  Using multiple path is probably much less used.
Which means that the most common usage will pay the price of a useless additional squared bracket.

I would then suggest to split the two behaviour, and introduce a new command for multiple paths.

JSON.GET would take only one path.
JSON.GETS (?) would take multiple path.
...
JSON MGET getting multiple keys (to complete the GET family overview)

oshadmi:
Replaced with #516

PR #497, opened 2021-10-31
Multipath - Add MGET backward compatibility

oshadmi:


tests/pytest/test.py
```diff
@@ -308,30 +308,30 @@ def testMgetCommand(env):
     r.assertEqual(len(raw), 5)
     for d in range(0, 5):
         key = 'doc:{}'.format(d)
-        r.assertEqual(json.loads(raw[d]), [docs['basic']], d)
+        r.assertEqual(json.loads(raw[d]), docs['basic'], d)
```
MeirShpilraien, line 311: I guess this reverts the tests changes?

tests/pytest/test_multi.py
```diff
@@ -159,8 +159,8 @@ def testMGetCommand(env):
     """Test REJSON.MGET command"""
     r = env
     # Test mget with multi paths
-    r.assertOk(r.execute_command('JSON.SET', 'doc1', '$', '{"a":1, "b": 2, "nested": {"a": 3}, "c": null, "nested2": {"a": null}} '))
-    r.assertOk(r.execute_command('JSON.SET', 'doc2', '$', '{"a":4, "b": 5, "nested": {"a": 6}, "c": null, "nested2": {"a": [null]}}'))
+    r.assertOk(r.execute_command('JSON.SET', 'doc1', '$', '{"a":1, "b": 2, "nested1": {"a": 3}, "c": null, "nested2": {"a": null}}'))
+    r.assertOk(r.execute_command('JSON.SET', 'doc2', '$', '{"a":4, "b": 5, "nested3": {"a": 6}, "c": null, "nested4": {"a": [null]}}'))
```
MeirShpilraien, line 163: Can you explain this change?

MeirShpilraien (commented):
Small questions

tests/pytest/test.py
oshadmi: Yes, the previous change ignored the previous behavior. Since the path here starts with a `.`, the old behavior is now kept.

tests/pytest/test_multi.py
oshadmi: Wanted a path that would match only one of the keys (for the missing path/key scenario)

MeirShpilraien: So why change existing test and not add a new one?

oshadmi: This had no impact on existing tests (just names of internal paths)
And it is now used in a new test

MeirShpilraien (approved):

PR #503, opened 2021-11-01
Multipath: handle default arguments

### LINKED ISSUE #502: JSON.ARRLEN behaviour doesn't match documentation, with defaults
Based on the documentation for [JSON.ARRLEN](https://oss.redis.com/redisjson/commands/#jsonarrlen) if the path isn't provided '.' is assumed. Similarly if a key is not provided, a null should be returned: 

``path defaults to root if not provided. If the key or path do not exist, null is returned.``

But, using our example below, when testing with a key that does not exist an error is raised, instead of None.  

```
127.0.0.1:36379> json.arrlen 'smirf'
(error) ERR wrong number of arguments for 'json.arrlen' command
127.0.0.1:36379> json.arrlen 'smirf' .
(nil)
127.0.0.1:36379> json.arrlen 'smirf' 
(error) ERR wrong number of arguments for 'json.arrlen' command
127.0.0.1:36379> 
```

In our second example - when setting a value, and attempting to run an _arrlen_ without specifying a path,  the default path of '.' should be used. But as seen below, it was not.

```
127.0.0.1:36379> json.set foo . '[5,6,7,8,9]'
OK
127.0.0.1:36379> json.arrlen foo
(error) ERR wrong number of arguments for 'json.arrlen' command
127.0.0.1:36379> json.arrlen foo .
(integer) 5
```

### LINKED ISSUE #504: JSON.STRLEN behaviour doesn't match documentation, with defaults
Based on the documentation for [JSON.STRLEN](https://oss.redis.com/redisjson/commands/#jsonstrlen) if the path isn't provided '.' is assumed. This is similar to what occurred in #502. If a key is not provided, a null should be returned: 

``path defaults to root if not provided. If the key or path do not exist, null is returned.``

But, using our example below, when testing with a key that does not exist an error is raised, instead of None.  

```
127.0.0.1:36379> json.strlen 'smirf'
(error) ERR wrong number of arguments for 'json.strlen' command
127.0.0.1:36379> json.arrlen 'strlen' .
(nil)
127.0.0.1:36379> json.arrlen 'strlen' 
(error) ERR wrong number of arguments for 'json.strlen' command
127.0.0.1:36379> 
```

In our second example - when setting a value, and attempting to run an strlen without specifying a path,  the default path of '.' should be used. But as seen below, it was not.

```
127.0.0.1:36379> json.set foo . '[5,6,7,8,9]'
OK
127.0.0.1:36379> json.strlen foo
(error) ERR wrong number of arguments for 'json.strlen' command
127.0.0.1:36379> json.strlen foo .
(integer) 5
```

oshadmi:
Fix #502
Fix #504 

In addition, handle/add tests for other default args: objkeys objlen resp get strappend arrpop

### DIFF
--- a/tests/pytest/test_multi.py
+++ b/tests/pytest/test_multi.py
@@ -455,6 +455,11 @@ def testArrLenCommand(env):
     # Test missing key
     r.assertEqual(r.execute_command('JSON.ARRLEN', 'non_existing_doc', '..a'), None)
 
+    # Test missing path
+    r.assertOk(r.execute_command('JSON.SET', 'doc1', '$', '[0, 1, 2, 3, 4]'))
+    res = r.execute_command('JSON.ARRLEN', 'doc1')
+    r.assertEqual(res, 5)
+
 def testArrPopCommand(env):
     """
     Test REJSON.ARRPOP command
--- a/tests/pytest/test_multi.py
+++ b/tests/pytest/test_multi.py
@@ -98,6 +98,8 @@ def testForgetCommand(env):
 
 def testSetAndGetCommands(env):
     """Test REJSON.SET command"""
+    """Test REJSON.GET command"""
+
     r = env
     # Test set and get on large nested key
     r.assertIsNone(r.execute_command('JSON.SET', 'doc1', '$', nested_large_key, 'XX'))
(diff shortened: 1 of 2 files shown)


tests/pytest/test_multi.py
```diff
@@ -489,14 +509,25 @@ def testArrPopCommand(env):
     res = r.execute_command('JSON.GET', 'doc1', '$')
     r.assertEqual(json.loads(res), [{"a": [], "nested1": {"a": ["hello", "world"]}, "nested2": {"a": 31}}])
     # Test single
-    res = r.execute_command('JSON.ARRPOP', 'doc1', '.nested1.a', -2, '"baz"')
+    res = r.execute_command('JSON.ARRPOP', 'doc1', '.nested1.a', -2)
```
MeirShpilraien, line 512: Can you explain this change?

MeirShpilraien (commented):
Small question

tests/pytest/test_multi.py
oshadmi: This is a superfluous argument
(we could consider adding `args.done()?;` to ensure no unnecessary args are supplied)

MeirShpilraien (approved):

oshadmi:
@emmanuelkeller I am closing #502 and #504 using this PR:
Addressing the default path argument, which was indeed missing,
But not addressing the `If the key or path do not exist, null is returned.`
Since in the new multipath behavior, `null` could be a legitimate return value, so we currently:
- fail on missing key with "new" path (beginning with a `$`)
- return `null` as before using "old" path (beginning with `.`)

Please reopen if you object.

PR #512, opened 2021-11-02
fixing incorrect arrtrim docs

chayim:


docs/commands.md
```diff
@@ -337,7 +337,7 @@ JSON.ARRTRIM <key> <path> <start> <stop>
 
 Trim an array so that it contains only the specified inclusive range of elements.
 
-This command is extremely forgiving and using it with out of range indexes will not produce an error. If `start` is larger than the array's size or `start` > `stop`, the result will be an empty array. If `start` is < 0 then it will be treated as 0. If `stop` is larger than the end of the array, it will be treated like the last element in it.
+This command is extremely forgiving and using it with out of range indexes will not produce an error. If `start` is larger than the array's size or `start` > `stop`, the result will be 0. If `start` is < 0 then it will be treated as 0. If `stop` is larger than the end of the array, it will be treated like the last element in it.
```
oshadmi, line 340: The sentence `the result will be an empty array` is not specifying the **Return value**.
It is specifying the outcome of the command - what will be the value of the key in the modified path.
Maybe the phrasing could be improved, such as
`the resulting array will be an empty array`

chayim: No, the phrasing  ``the result will be an empty array`` linguistically indicates that you receive both an array, and that it's empty. They're semantically the same thing, just with fewer words.

chayim:
Superceeded by #538

PR #515, opened 2021-11-02
JSON.DEL count deleted null value

oshadmi:
`JSON.DEL` with a path leading to a json `null` should be counted by the return value

### DIFF
--- a/src/redisjson.rs
+++ b/src/redisjson.rs
@@ -247,10 +247,8 @@ impl RedisJSON {
 
     pub fn delete_path(&mut self, path: &str) -> Result<usize, Error> {
         let mut deleted = 0;
-        self.data = jsonpath_lib::replace_with(self.data.take(), path, |v| {
-            if !v.is_null() {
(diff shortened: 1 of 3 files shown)


MeirShpilraien (dismissed):

src/redisjson.rs
```diff
@@ -247,10 +247,8 @@ impl RedisJSON {
 
     pub fn delete_path(&mut self, path: &str) -> Result<usize, Error> {
         let mut deleted = 0;
-        self.data = jsonpath_lib::replace_with(self.data.take(), path, |v| {
-            if !v.is_null() {
-                deleted += 1; // might delete more than a single value
-            }
+        self.data = jsonpath_lib::replace_with(self.data.take(), path, |_v| {
+            deleted += 1; // might delete more than a single value
```
MeirShpilraien, line 251: So this is not needed right?

MeirShpilraien (approved):

PR #531, opened 2021-11-13
Multipath - DEL - order by array index even if path lengths are different

### LINKED ISSUE #529: JSON.DEL doesn't delete all values with recursive descent (..)
```
127.0.0.1:6379> JSON.SET doc1 $ '[{"a00": [{"a00": "a00_00"}, {"a01": "a00_01"}, {"a02": "a00_02"}, {"a03": "a00_03"}]}, {"a01": [{"a00": "a01_00"}, {"a01": "a01_01"}, {"a02": "a01_02"}, {"a03": "a01_03"}]}, {"a02": [{"a00": "a02_00"}, {"a01": "a02_01"}, {"a02": "a02_02"}, {"a03": "a02_03"}]}, {"a03": [{"a00": "a03_00"}, {"a01": "a03_01"}, {"a02": "a03_02"}, {"a03": "a03_03"}]}]'
OK
127.0.0.1:6379> JSON.ARRLEN doc1 $.*[*]
1) (integer) 4
2) (integer) 4
3) (integer) 4
4) (integer) 4
127.0.0.1:6379> JSON.DEL doc1 $..[2]
(integer) 3
```
`JSON.DEL` should return `4`, deleting the entire object containing `a02`, and each element at index 2 in the other 3 arrays.
The array in the object containing `a03`, should have lost one element, but it didn't:
```
127.0.0.1:6379> JSON.ARRLEN doc1 $.*[*]
1) (integer) 3
2) (integer) 3
3) (integer) 4
```
(last value of the response should have been `3` instead of `4`)

oshadmi:
Need to delete a longer path before a shorter path, if longer path has higher array index compared to the shorter path
Fixes #529

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -16,7 +16,7 @@ use crate::redisjson::SetOptions;
 use serde_json::{Number, Value};
 
 use itertools::FoldWhile::{Continue, Done};
-use itertools::Itertools;
+use itertools::{EitherOrBoth, Itertools};
 use serde::{Serialize, Serializer};
 use std::collections::HashMap;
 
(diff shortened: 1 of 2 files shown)


tests/pytest/test_multi.py
```diff
@@ -66,7 +66,7 @@ def testDelCommand(env):
     r.assertEqual(res, 3)
 
     r.assertOk(r.execute_command('JSON.SET', 'doc2', '$', '[1, 2, 3]'))
-    res = r.execute_command('JSON.DEL', 'doc2', '$[2,1,0]')
+    res = r.execute_command('JSON.DEL', 'doc2', '$[1,2,0]')
```
MeirShpilraien, line 69: Why this change?

src/commands.rs
```diff
@@ -758,40 +758,43 @@ where
...
-                        } else {
-                            Continue(Ordering::Equal)
+    paths.sort_by(|v1, v2| {
+        v1.iter()
+            .zip_longest(v2.iter())
+            .fold_while(Ordering::Equal, |_acc, v| {
+                match v {
+                    EitherOrBoth::Left(_) => Done(Ordering::Greater), // Shorter paths before longer paths
+                    EitherOrBoth::Right(_) => Done(Ordering::Less), // Shorter paths before longer paths
+                    EitherOrBoth::Both(p1, p2) => {
```
MeirShpilraien, line 771: What would have happened before if lenght of the paths are different?

MeirShpilraien (commented):
Small questions

tests/pytest/test_multi.py
oshadmi: It was initially in a none-descending order, but was worked around a bug, and now can be restored to a less trivial order which would test the reordering required by DEL.

src/commands.rs
oshadmi: Before, we did not compare content of vectors with different lengths.
We only compared content for similar lengths, but this was wrong since a longer path could overlap with a shorter path and must be handled before, as seen in #529

tests/pytest/test_multi.py
MeirShpilraien: So let's keep them both?

MeirShpilraien (dismissed):
A small suggestion of keep a test and add another. Not a must.

MeirShpilraien (approved):

PR #538, opened 2021-11-15
Doc 2.0

emmanuelkeller:


emmanuelkeller:
@oshadmi Can you review ? And then update the Big O notation for each commands? (if needed)

oshadmi:
> @oshadmi Can you review ? And then update the Big O notation for each commands? (if needed)

@emmanuelkeller
Sure, I'll review.
Regarding complexity, can you please review and copy it from [commands.json](https://github.com/RedisJSON/RedisJSON/blob/master/commands.json)

emmanuelkeller:
@oshadmi Thanks ! Done, I have updated the time complexity for each command

docs/commands.md
```diff
@@ -235,7 +322,20 @@ Append the `json` value(s) into the array at `path` after the last element in it
 
 #### Return value
 
-[Integer][2], specifically the array's new size.
+[Array][4] of [Integer][2], specifically, for each path, the array's new size, or [null][6] element if the matching JSON value is not an array
```
oshadmi, line 324:
```suggestion
[Array][4] of [Integer][2], specifically, for each path, the array's new size, or [null][6] element if the matching JSON value is not an array.
```

docs/path.md
```diff
@@ -2,6 +2,22 @@
...
+RedisJSON decide which implementation to used depending on the first character of the path query. If the query starts with the character `$` it is considered as a JSONPath query. Otherwise is is interpreted as a legacy path syntax.
+
+## JSONPath support (RedisJSON v2)
+
+RedisJSON 2.0 introduces the support of [JSONPath](http://goessner.net/articles/JsonPath/). It follows the syntax described by Goessner in his article.
+
+JSONPath query can resolve to several location in the JSON documents. In this case, the JSON commands are applying the operation to every possible location. This is a major improvement over the legacy support, which was operating only on the first path.
+
+Notice that the structure of the command response is most of the time different when using JSONPath. This two behavior are described in the [Commands](/commands) page.
+
```
oshadmi, line 16: Do we want to mention the limitation of using colon (`:`) character in a path using the dot notation (MOD-1635) and the workaround of using bracket notation?

docs/path.md
```diff
@@ -2,6 +2,22 @@
...
+## Legacy Path syntax (RedisJSON v1)
```
oshadmi, line 17: What about `JSON.TOGGLE` and `JSON.CLEAR` ?

docs/commands.md
```diff
@@ -341,7 +518,25 @@ This command is extremely forgiving and using it with out of range indexes will
```
oshadmi, line 516:
```suggestion
This command is extremely forgiving and using it with out of range indexes will not produce an error. If `start` is larger than the array's size or `start` > `stop`, the return value will be 0 and the resulting array will be empty. If `start` is < 0 then it is interpreted from the end. If `stop` is larger than the end of the array, it will be treated like the last element in it.
```

docs/path.md
```diff
@@ -2,6 +2,22 @@
 
 Since there does not exist a standard for path syntax, RedisJSON implements its own. RedisJSON's syntax is a subset of common best practices and resembles [JSONPath](http://goessner.net/articles/JsonPath/) not by accident.
 
+There is currently two concurrent implementations. One is a legacy from the first version of RedisJSON. Describe below as the `legacy path syntax`.
```
oshadmi, line 5: What does an implementation "from the first version of RedisJSON" mean? That we still have the exact same implementation?
2.0 behavior is not identical to 1.0, although tries to have as much as backward compatibility as possible (we changed ARRTRIM negative start index behavior, some error messages, and maybe a few more differences)

oshadmi (commented):
Some comments

docs/commands.md
```diff
@@ -172,7 +223,18 @@ Multiplies the number value stored at `path` by `number`.
...
+[Bulk String][3], specifically the stringified new values for each path.
+
+#### Example
+
+```sql
+127.0.0.1:6379> JSON.SET doc . '{"a":"b","b":[{"a":2}, {"a":5}, {"a":"c"}]}'
+OK
+127.0.0.1:6379> JSON.NUMINCRBY doc $.a 2
+"[null]"
+127.0.0.1:6379> JSON.NUMINCRBY doc $..a 2
```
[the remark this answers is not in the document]
emmanuelkeller: Let's not promote it ..

docs/path.md
emmanuelkeller: Brilliant !

docs/path.md
emmanuelkeller: Rephrasing ...

docs/path.md
emmanuelkeller: Let's have them documented with the upcoming HDT version

docs/indexing_JSON.md
```diff
@@ -0,0 +1,208 @@
...
+2) "org:1"
+3) 1) "$"
+   2) "{\"name\":\"Headquarters\",\"address\":[\"Suite 250\",\"Mountain View\"],\"cp\":\"CA 94040\"}"
+```
+
+### JSON strings and numbers as to be indexed as TEXT and NUMERIC
+
+- JSON Strings can only be indexed as TEXT, TAG, and GEO (using the correct syntax).
+- JSON numbers can only be indexes as NUMERIC.
+- Boolean and NULL values are ignored.
```
oshadmi, line 194: Not ignored I think.
Boolean should be indexed as TAG.

docs/indexing_JSON.md
```diff
@@ -0,0 +1,208 @@
...
+
+We can now specify `ON JSON` to inform RediSearch that we want to index JSON documents.
+
+Then, on the `SCHEMA` part, you can provide JSONPath expressions.
+The result of each `JSON Path` expression is indexed and associated with a logical name (`attribute`).
+Use the attribute name in the query part.
+
+Here is the basic syntax for indexing a JSON document:
+
+    FT.CREATE {index_name} ON JSON SCHEMA {json_path} AS {attribute} {type}
```
oshadmi, line 25: Should we emphasize that it is currently **required** to define an alias, in order to use it as an attribute in `FT.SEARCH` and `FT.AGGREGATE`
(there is a limitation in the query parser that might fail with syntax error with complex JSONPath expressions)

docs/indexing_JSON.md
```diff
@@ -28,6 +28,8 @@ And here's a concrete example:
 
     FT.CREATE userIdx ON JSON SCHEMA $.user.name AS name TEXT $.user.tag AS country TAG
 
+Note: The attribute is optional, but it is mandatory as you can't use a JSON Path expression inside the query (FT.SEARCH).
```
oshadmi, line 31:
```suggestion
Note: The attribute is optional in `FT.CREATE`, but it is required to use attributes in attribute modifiers in the query of `FT.SEARCH` and  `FT.AGGREGATE`, and avoid using JSON Path expressions, which are not fully supported by the query parser.
```

PR #567, opened 2021-12-08
Added memory consumption test

rafie:


tests/pytest/test_mem.py
```diff
@@ -0,0 +1,78 @@
+
+from common import *
+import time
+import datetime
+
+
+CHECK_MEMRECLAIM = os.getenv('MEMRECLAIM', '0') == '1'
+
+JSON_FILES = [
```
gkorland, line 9: @rafie I think it's better to copy these files so it won't change without us knowing about it.

tests/pytest/test_mem.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def __del__(self):
+        for jfile in JSON_FILES:
+            os.unlink(jfile['path'])
+
+    def testKeys(self):
+        def add_and_check(title):
+            t0 = time.monotonic()
+            for i in range(0, 100):
+                env.execute_command('json.set', f'json{i}', '.', jfile['doc'])
```
gkorland, line 33:
```suggestion
                env.execute_command('json.set', f'json{i}', '$', jfile['doc'])
```

tests/pytest/test_mem.py
```diff
@@ -0,0 +1,78 @@
...
+                add_and_check(f"add after del (keys {fi})")
+            env.execute_command('flushall')
+            env.stop()
+
+    def testFields(self):
+        def add_and_check(title):
+            t0 = time.monotonic()
+            env.execute_command('json.set', 'json', '.', '{}')
+            for i in range(0, 100):
+                env.execute_command('json.set', 'json', f'.json{i}', jfile['doc'])
```
gkorland, line 59:
```suggestion
                env.execute_command('json.set', 'json', f'$.json{i}', jfile['doc'])
```

tests/pytest/tests.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 [[ $VERBOSE == 1 ]] && set -x
-[[ $IGNERR == 1 ]] || set -e
+# [[ $IGNERR == 1 ]] || set -e
```
gkorland, line 4: @rafie revert?

gkorland (changes requested):

tests/pytest/test_mem.py
rafie: This was intentional, as I used it for 1.0 too, but here we can go with $.

tests/pytest/test_mem.py
rafie: They can become quite big and I don't want to introduce git-lfs. When it breaks, we'll fix it.

tests/pytest/tests.sh
rafie: Intentional. `shibumi/defs` installs error handler. However, the `set -x` need to be commented out as it creates too much noise when going `VERBOSE=1`.

tests/pytest/common.py
```diff
@@ -14,3 +17,23 @@ def handler(signum, frame):
...
+    vms = meminfo.vms / 1024
+    rss = meminfo.rss / 1024
+    return {'vsz': vms, 'rss': rss }
+
+def checkEnvMem(env, expected_vsz=None, vsz0=0, threshold=0.1, title=None):
+    if MEMINFO:
+        if title is not None:
+            print(f"--- {title}")
+        pid = env.envRunner.masterProcess.pid
+        print(paella.sh(f'cat /proc/{pid}/status | grep ^Vm', join=False))
```
oshadmi, line 33: Do we always want to print?  Maybe use `env.debugPrint`?

tests/pytest/test_mem.py
```diff
@@ -0,0 +1,79 @@
...
+        for jfile in JSON_FILES:
+            os.unlink(jfile['path'])
+
+    def testKeys(self):
+        def add_and_check(title):
+            t0 = time.monotonic()
+            for i in range(0, 100):
+                env.execute_command('json.set', f'json{i}', '$', jfile['doc'])
+            title += f" t={datetime.timedelta(seconds=time.monotonic() - t0)}"
+            checkEnvMem(env, expected_vsz=jfile['vsz'], vsz0=vsz0, title=title)
```
oshadmi, line 36: Also here and other `print` - do we always want to print?
What is `vsz0` in right-hand-side of `vsz0=vsz0` ?

tests/pytest/test_mem.py
```diff
@@ -0,0 +1,79 @@
...
+        for jfile in JSON_FILES:
+            fi += 1
+            env = Env()
+            vsz0 = checkEnvMem(env, title=f"before (keys {fi})")
+            add_and_check(f"add (keys {fi})")
+            if CHECK_MEMRECLAIM:
+                delete()
+                add_and_check(f"add after del (keys {fi})")
+            env.execute_command('flushall')
+            env.stop()
```
oshadmi, line 53: Can we also check mem here? Should be the initial mem before adding any key?

tests/pytest/common.py
rafie: This runs only with `MEMINFO=1`.

tests/pytest/test_mem.py
rafie: This is a named argument.

tests/pytest/test_mem.py
rafie: Redis does not give memory back that easily. It's hard to get a consistent behavior (not like with fields, where we manage memory ourselves).

tests/pytest/common.py
```diff
@@ -14,3 +17,23 @@ def handler(signum, frame):
...
+    mem = envMem(env)
```
MeirShpilraien, line 34: Why not use `info memory`?

tests/pytest/test_mem.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def __del__(self):
+        for jfile in JSON_FILES:
+            os.unlink(jfile['path'])
+
+    def testKeys(self):
+        def add_and_check(title):
+            t0 = time.monotonic()
+            for i in range(0, 100):
+                env.execute_command('json.set', f'json{i}', '$', jfile['doc'])
```
MeirShpilraien, line 34: I would make jfile as an argument to the function, its not clear where it is coming from..

tests/pytest/test_mem.py
```diff
@@ -0,0 +1,79 @@
...
+            title += f" t={datetime.timedelta(seconds=time.monotonic() - t0)}"
+            checkEnvMem(env, expected_vsz=jfile['vsz'], vsz0=vsz0, title=title)
+        def delete():
+            t0 = time.monotonic()
+            for i in range(0, 100):
+                env.execute_command('json.del', f'json{i}')
+            print(f"--- del: t={datetime.timedelta(seconds=time.monotonic() - t0)}")
```
MeirShpilraien, line 41: env.debug_print maybe?

tests/pytest/common.py
MeirShpilraien: Agree with @oshadmi

tests/pytest/test_mem.py
```diff
@@ -0,0 +1,79 @@
...
+        def delete():
+            t0 = time.monotonic()
+            for i in range(0, 100):
+                env.execute_command('json.del', 'json', f'json{i}')
+            print(f"--- del: t={datetime.timedelta(seconds=time.monotonic() - t0)}")
+
+        fi = 0
+        for jfile in JSON_FILES:
+            fi += 1
+            env = Env()
```
MeirShpilraien, line 72: You should not create more than one env inside a test, either use the same env or make multiple tests.

tests/pytest/test_mem.py
```diff
@@ -0,0 +1,79 @@
...
+        def delete():
+            t0 = time.monotonic()
+            for i in range(0, 100):
+                env.execute_command('json.del', f'json{i}')
+            print(f"--- del: t={datetime.timedelta(seconds=time.monotonic() - t0)}")
+
+        fi = 0
+        for jfile in JSON_FILES:
+            fi += 1
+            env = Env()
```
MeirShpilraien, line 46: Do not create more than one env inside a test

tests/pytest/test_mem.py
```diff
@@ -0,0 +1,79 @@
...
+            env.execute_command('json.set', 'json', '.', '{}')
+            for i in range(0, 100):
+                env.execute_command('json.set', 'json', f'$.json{i}', jfile['doc'])
+            title += f" t={datetime.timedelta(seconds=time.monotonic() - t0)}"
+            checkEnvMem(env, expected_vsz=jfile['vsz'], vsz0=vsz0, title=title)
+        def delete():
+            t0 = time.monotonic()
+            for i in range(0, 100):
+                env.execute_command('json.del', 'json', f'json{i}')
+            print(f"--- del: t={datetime.timedelta(seconds=time.monotonic() - t0)}")
```
MeirShpilraien, line 67: debugPrint

tests/pytest/test_mem.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+CHECK_MEMRECLAIM = os.getenv('MEMRECLAIM', '0') == '1'
+SERDE_JSON = os.getenv('SERDE_JSON', '0') == '1'
+
+JSON_FILES = [
+    {'file': 'https://raw.githubusercontent.com/mloskot/json_benchmark/master/data/canada.json',
+     'vsz': 2880000 if SERDE_JSON else 598000},
+    {'file': 'https://raw.githubusercontent.com/RichardHightower/json-parsers-benchmark/master/data/citm_catalog.json',
+     'vsz': 847000 if SERDE_JSON else 200000},
```
MeirShpilraien, line 14: Why not run the tests both with serde_json and ijson?

MeirShpilraien (commented):
Only review the tests part without the build changes.

tests/pytest/test_mem.py
rafie: but they do. when we run with serde_json, `SERDE_JSON` env var is 1.

tests/pytest/test_mem.py
rafie: it's coming from the closure, but I can add.

tests/pytest/common.py
rafie: Do you prefer redis-server internal indications over system ones when investigating memory issues?
In most cases you'll be measuring `vsz` anyway so why not use it in the first place?

tests/pytest/common.py
rafie: Well, I tried `env.debugPrint()` (with `RLTest --debug-print`) and got such a load of irrelevant garbage that I immediately reverted to `print()` (with `MEMINFO=1`). This is what happens when Truth (c) collides with user experience.

tests/pytest/test_mem.py
MeirShpilraien: I meant why you want to control it from outside and not just run on both here?

tests/pytest/common.py
MeirShpilraien: `info memory` gives you both RSS and the memory Redis think it consumes so I do not see why not use it instead of using `env.envRunner.masterProcess.pid` which is internal to RLTest and might change someday and break this test.

tests/pytest/common.py
MeirShpilraien: Still its better to us `debugPrint`, debug prints should be verbose, you only enable it when you want better logging for debug. Also if you use `debugPrint` you will get those prints when you run with `-vv`.

tests/pytest/test_mem.py
rafie: There's a full cycle of tests running with `JSON_BACKEND SERDE_JSON`, and when this is the case, we'll be testing it's memory usage. Otherwise, we'll be doing both backends on each cycle, which is redundant and somewhat confusing.

tests/pytest/common.py
rafie: Ok.

tests/pytest/common.py
rafie: ok.

PR #574, opened 2021-12-13
CLang sanitizer support

rafie:


Makefile
```diff
@@ -55,6 +55,8 @@ make platform      # build for specific Linux distribution
   ARTIFACTS=1        # copy artifacts from docker image
   PUBLISH=1          # publish (i.e. docker push) after build
 
+make sanbox        # create container with CLang Sanitizer
```
oshadmi, line 58: Need to mention the /w arg for volume?

oshadmi (approved):

Makefile
rafie: It's a sort of hidden feature at the moment. :-)

rafie:
@gkorland Memory sanitizer disabled: all tests pass now.

Makefile
```diff
@@ -156,7 +172,7 @@ endif
 test: pytest
 
 pytest:
-	$(SHOW)MODULE=$(abspath $(TARGET)) ./tests/pytest/tests.sh
+	$(SHOW)MODULE=$(abspath $(TARGET)) $(realpath ./tests/pytest/tests.sh)
```
chayim, line 175: realpath (sadly) still isn't completely portable. Can you use realink -f?

tests/pytest/tests.sh
```diff
@@ -49,9 +50,39 @@ install_git_lfs() {
 
 #----------------------------------------------------------------------------------------------
 
-check_redis_server() {
-	if ! command -v redis-server > /dev/null; then
-		echo "Cannot find redis-server. Aborting."
+setup_redis_server() {
+	if [[ -n $SAN ]]; then
```
chayim, line 54: ${SAN} (etc) makes for safer shell scripts

tests/pytest/tests.sh
```diff
@@ -90,32 +121,115 @@ valgrind_summary() {
...
 	cd $ROOT/tests/pytest
-	[[ ! -z $TESTMOD ]] && RLTEST_ARGS+=--module $TESTMOD
-	$OP python3 -m RLTest --clear-logs --module $MODULE --module-args "JSON_BACKEND SERDE_JSON" $RLTEST_ARGS
-	$OP python3 -m RLTest --clear-logs --module $MODULE $RLTEST_ARGS
+
+	[[ $SERDE_JSON == 1 ]] && MODARGS+=" JSON_BACKEND SERDE_JSON"
+	
+	if [[ $EXISTING_ENV != 1 ]]; then
+		rltest_config=$(mktemp "${TMPDIR:-/tmp}/rltest.XXXXXXX")
+		rm -f $rltest_config
```
chayim, line 136: You shouldn't have to remote this file if you're calling mktemp on it.

tests/pytest/tests.sh
```diff
@@ -90,32 +121,115 @@ valgrind_summary() {
...
+
+	else # existing env
+		xredis_conf=$(mktemp "${TMPDIR:-/tmp}/xredis_conf.XXXXXXX")
+		rm -f $xredis_conf
+		cat <<-EOF > $xredis_conf
+			loadmodule $MODULE $MODARGS
+			EOF
+
+		rltest_config=$(mktemp "${TMPDIR:-/tmp}/xredis_rltest.XXXXXXX")
+		rm -f $rltest_config
```
chayim, line 156: Same - if calling mktemp you don't need to remove the file, even though bashwarting (cat <<-EOF) to the new file.

Makefile
rafie: This in gnu make's `realpath`, not `/bin/realpath`. This fortunately works properly.

tests/pytest/tests.sh
rafie: So you'd like to wrap any variable reference with {}, even when it's clear we're on the safe side? Wouldn't that hurt readability a bit? BTW I'm more concerned with quoting (in this case bash handles this ok so it can be omited, but in too many cases you do see `A=$B`).

tests/pytest/tests.sh
rafie: Right, of course. I really don't know where this came from.

Makefile
rafie: BTW the reason we require `realpath` here in the first place is because using a relative path causing `test.sh` not to print program name properly when it detects an error. So this needs to be fixed to (probably something like `PROGNAME="$0"` at the top).

tests/pytest/tests.sh
```diff
@@ -19,22 +19,23 @@ help() {
...
-		SLAVES=0|1       Tests with --test-slaves
+		GEN=1            General tests
+		AOF=1            Tests with --test-aof
+		SLAVES=1         Tests with --test-slaves
+		CLUSTER=1        Test with OSS cluster, one shard
+		QUICK=1          Run general tests only
 		
-		TEST=test        Run specific test (e.g. test.py:test_name)
+		VALGRIND|VG=1    Run with Valgrind
+		SAN=type         Use LLVM sanitizer (type=address|memory|leak|thread)
```
oshadmi, line 33: Do we want to remove `leak` and `thread` from the help string in the meantime?

oshadmi (approved):
👍🏼

PR #586, opened 2021-12-23
unit tests for error.rs

gkorland:


rafie (approved):

src/error.rs
```diff
@@ -72,3 +72,67 @@ impl From<Error> for redis_module::RedisError {
...
+        let err: Error = JsonPathError::Path("JsonPathError".to_string()).into();
+        assert_eq!(err.msg, "JSON Path error: path error: \\nJsonPathError\\n");
+    }
+
+    #[test]
+    fn test_from_str() {
+        let err: Error = "error from str".into();
+        assert_eq!(err.msg, "error from str");
+    }
+}
```
oshadmi, line 138: Consider adding
```suggestion

    #[test]
    fn test_from_error_to_redis_error() {
        let err: Error = String::from("error from String").into();
        let redis_err: RedisError = err.into();

        if let RedisError::String(s) = redis_err {
            assert_eq!(s, "error from String");
        }
    }

    #[test]
    fn test_from_redis_error_to_error() {
        let redis_err = RedisError::String(String::from("redis error from String"));
        let err: Error = redis_err.into();
        assert_eq!(err.msg, "ERR redis error from String");
    }
}
```

oshadmi (approved):

src/error.rs
gkorland: @oshadmi I don't get the suggestion, it seems like you suggest to delete on `}`

oshadmi: @gkorland Suggested adding 2 tests at the end
(the closing bracket `}` is removed, but also added with 2 additional tests)

PR #587, opened 2021-12-23
complete unit tests

gkorland:


.circleci/config.yml
```diff
@@ -153,7 +153,7 @@ commands:
           name: Run all tests
           command: |
             bash -l -c "python3 -m RLTest --version"
-            bash -l -c "make pytest"
+            bash -l -c "make test"
```
rafie, line 156: You can state `shell: /bin/bash -l -eo pipefail` (parallel to `command:`) and avoid repeating the `bash -l -c` thing.

gkorland: I see this how it's being used through out the scripts...

rafie: @gkorland The reason for `bash -l` is that it enables execution of scripts that were put in `/etc/profile.d` by various setup tools. CircleCI makes it possible to change the way bash is invoked for the whole `run` scriptlet, hence enabling `bash -l` for all commands at once.

PR #590, opened 2021-12-24
@node-redis+json@1.0.1_@node-redis+client@1.0.1\node_modules\@node-redis\json\dist\commands\index.js:88:18

hypo-thesis:
return jsons.map(transformRedisJsonNullReply);
                 ^
TypeError: jsons.map is not a function

when trying to do arrpop.
I resolved it by > 
from > 
```

function transformRedisJsonNullArrayNullReply(jsons) {
    if (jsons === null)
        return null;
    return jsons.map(transformRedisJsonNullReply);
}
```

to 

```
function transformRedisJsonNullArrayNullReply(jsons) {
    if (jsons === null)
        return null;
        jsons = [jsons];
    return jsons.map(transformRedisJsonNullReply);
}
```

@node-redis+json@1.0.1_@node-redis+client@1.0.1\node_modules\@node-redis\json\dist\commands\index.js:88:18


gkorland:
@hypo-thesis I'm not sure how this PR related to the bug in node-redis.
Correct me if I'm wrong but it seems like you meant to report a bug in node-redis and not open a pull request.
If that is the case please open the issue in this repo - https://github.com/redis/node-redis/issues

PR #591, opened 2021-12-25
Avoid crash on overflow in numincrby or nummultby

oshadmi:
Incrementing beyond range of number is crashing redis
```
127.0.0.1:6379> JSON.SET key $ 1.6350000000001314e+308
OK
127.0.0.1:6379> JSON.NUMINCRBY key $ 1.6350000000001314e+308
```

The problem was that `update` method was taking the value being updated, e.g., then applying a closure on it, and then setting the value to the value returned from the closure, and 2 problems could have been caused:
1) The closure could have failed, e.g., on a numeric overflow - and was unwrapped
2) Value was not being restored upon error, leaving a null in place of the previous value

The fix is to handle errors properly, and to set the value by ref, so it is not lost on error.
No public API was broken.

More details:
About `replace`
```
///
/// Replaces a value at a given `path`, starting from `root`
///
/// The new value is the value returned from `func`, which is called on the current value.
///
/// If the returned value from `func` is `None`, the current value is removed.
/// If the returned value from `func` is `Err`, the current value remains (although it could be modified by `func`)
///
fn replace<F: FnMut(&mut IValue) -> Result<Option<IValue>, Error>>
```

About `update`
```
///
/// Updates a value at a given `path`, starting from `root`
///
/// The value is modified by `func`, which is called on the current value.
/// If the returned value from `func` is `None`, the current value is removed.
/// If the returned value from `func` is `Err`, the current value remains (although it could be modified by `func`)
///
fn update<F: FnMut(&mut IValue) -> Result<Option<()>, Error>>
```

### DIFF
--- a/src/ivalue_manager.rs
+++ b/src/ivalue_manager.rs
@@ -98,8 +98,16 @@ impl<'a> IValueKeyHolderWrite<'a> {
         if paths.is_empty() {
             // updating the root require special treatment
             let root = self.get_value().unwrap().unwrap();
-            let res = (op_fun)(root.take())?;
-            self.set_root(res)?;
+            let prev_val = root.take();
+            let rollback_val = prev_val.clone();
+            let res = (op_fun)(prev_val);
+            match res {
+                Ok(res) => self.set_root(res)?,
+                Err(err) => {
+                    self.set_root(Some(rollback_val))?;
+                    return Err(RedisError::String(err.msg));
+                }
+            }
         } else {
             update(&paths, self.get_value().unwrap().unwrap(), op_fun)?;
         }
@@ -126,14 +134,18 @@ impl<'a> IValueKeyHolderWrite<'a> {
                     v.as_number().unwrap().has_decimal_point(),
                     in_value.as_i64(),
                 ) {
-                    (false, Some(num2)) => ((op1_fun)(v.to_i64().unwrap(), num2)).into(),
+                    (false, Some(num2)) => Ok(((op1_fun)(v.to_i64().unwrap(), num2)).into()),
                     _ => {
                         let num1 = v.to_f64().unwrap();
                         let num2 = in_value.as_f64().unwrap();
-                        INumber::try_from((op2_fun)(num1, num2)).unwrap()
+                        if let Ok(num) = INumber::try_from((op2_fun)(num1, num2)) {
+                            Ok(num)
+                        } else {
+                            Err(RedisError::Str("result is not a number"))
+                        }
                     }
                 };
-                res = Some(IValue::from(num_res));
+                res = Some(IValue::from(num_res?));
                 Ok(res.clone())
             })?;
             match res {
--- a/src/ivalue_manager.rs
+++ b/src/ivalue_manager.rs
@@ -5,7 +5,7 @@ use crate::redisjson::normalize_arr_start_index;
 use crate::Format;
 use crate::REDIS_JSON_TYPE;
 use ijson::object::Entry;
-use ijson::{INumber, IValue, ValueType};
+use ijson::{DestructuredMut, INumber, IString, IValue, ValueType};
 use redis_module::key::{verify_type, RedisKey, RedisKeyWritable};
 use redis_module::raw::{RedisModuleKey, Status};
 use redis_module::rediserror::RedisError;
@@ -29,7 +29,7 @@ pub struct IValueKeyHolderWrite<'a> {
     val: Option<&'a mut RedisJSON<IValue>>,
 }
 
-fn update<F: FnMut(IValue) -> Result<Option<IValue>, Error>>(
+fn update<F: FnMut(&mut IValue) -> Result<Option<IValue>, Error>>(
     path: &Vec<String>,
     root: &mut IValue,
     mut func: F,
@@ -46,11 +46,16 @@ fn update<F: FnMut(IValue) -> Result<Option<IValue>, Error>>(
                 let obj = target_once.as_object_mut().unwrap();
                 if is_last {
                     if let Entry::Occupied(mut e) = obj.entry(token) {
-                        let v = e.insert(IValue::NULL);
-                        if let Some(res) = (func)(v)? {
-                            e.insert(res);
-                        } else {
-                            e.remove();
+                        let mut v = e.get_mut();
+                        match (func)(&mut v) {
+                            Ok(res) => {
+                                if let Some(res) = res {
+                                    *v = res;
+                                } else {
+                                    e.remove();
+                                }
+                            }
+                            Err(err) => return Err(err),
                         }
                     }
                     return Ok(());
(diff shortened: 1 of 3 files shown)


tests/pytest/test.py
```diff
@@ -807,6 +807,12 @@ def testNumIncrCommand(env):
     r.assertEqual(1, res['foo'])
     r.assertEqual(84, res['bar'])
 
+    # test overflow
+    r.assertOk(r.execute_command('JSON.SET', 'big_num', '.', '1.6350000000001313e+308'))
+    r.expect('JSON.NUMINCRBY', 'big_num', '.', '1.6350000000001313e+308').raiseError()
+    r.expect('JSON.NUMMULTBY', 'big_num', '.', '2').raiseError()
+    # (value remains)
+    r.assertEqual(r.execute_command('JSON.GET', 'big_num', '.'), '1.6350000000001313e308')
```
MeirShpilraien, line 815: Can we also test it on a none root element?

MeirShpilraien (commented):
Suggestion

src/ivalue_manager.rs
```diff
@@ -46,11 +46,16 @@ fn update<F: FnMut(IValue) -> Result<Option<IValue>, Error>>(
...
-                        let v = e.insert(IValue::NULL);
-                        if let Some(res) = (func)(v)? {
-                            e.insert(res);
-                        } else {
-                            e.remove();
+                        let mut v = e.get_mut();
+                        match (func)(&mut v) {
+                            Ok(res) => {
+                                if let Some(res) = res {
+                                    *v = res;
```
MeirShpilraien, line 53: Maybe I am missing something but how this API allows me to edit the existing entry and keep it?

src/ivalue_manager.rs
```diff
@@ -63,11 +68,81 @@ fn update<F: FnMut(IValue) -> Result<Option<IValue>, Error>>(
...
-                            let v = std::mem::replace(&mut arr[x], IValue::NULL);
-                            if let Some(res) = (func)(v)? {
-                                arr[x] = res;
-                            } else {
-                                arr.remove(x);
+                            let mut v = &mut arr.as_mut_slice()[x];
+                            match (func)(&mut v) {
+                                Ok(res) => {
+                                    if let Some(res) = res {
+                                        *v = res;
```
MeirShpilraien, line 75: Same question here, how do I edit existing entry and keep it?

src/ivalue_manager.rs
```diff
@@ -63,11 +68,81 @@ fn update<F: FnMut(IValue) -> Result<Option<IValue>, Error>>(
...
+                                    }
+                                }
+                                Err(err) => return Err(err),
+                            }
+                        }
+                        return Ok(());
+                    }
+                    arr.get_mut(x)
+                } else {
+                    None
```
MeirShpilraien, line 87: Can this happened? if not asset.

src/ivalue_manager.rs
```diff
@@ -63,11 +68,81 @@ fn update<F: FnMut(IValue) -> Result<Option<IValue>, Error>>(
...
+                }
+            }
+            _ => None,
```
MeirShpilraien, line 90: Can this happened? if not asset.

src/ivalue_manager.rs
```diff
@@ -63,11 +68,81 @@ fn update<F: FnMut(IValue) -> Result<Option<IValue>, Error>>(
...
+            target = t;
+        } else {
+            break;
+        }
+    }
+
+    Ok(())
+}
+
+fn update_in_closure<F: FnMut(&mut IValue) -> Result<Option<()>, Error>>(
```
MeirShpilraien, line 103: Can you add a documentation to this function that explain where is it used and how is it different from the one above?

src/ivalue_manager.rs
```diff
@@ -247,25 +337,24 @@ impl<'a> WriteHolder<IValue, IValue> for IValueKeyHolderWrite<'a> {
                     o.insert(key.to_string(), v.take());
                 }
             }
-            let val = root.take();
-            self.set_root(Some(val))?;
         } else {
-            update(&path, self.get_value().unwrap().unwrap(), |mut val| {
-                if let Some(o) = val.as_object_mut() {
+            update_in_closure(&path, self.get_value().unwrap().unwrap(), |val| {
```
MeirShpilraien, line 341: I guess this allows you to update an existing value, maybe call one update and the other replace?

src/ivalue_manager.rs
```diff
@@ -287,9 +376,12 @@ impl<'a> WriteHolder<IValue, IValue> for IValueKeyHolderWrite<'a> {
     fn bool_toggle(&mut self, path: Vec<String>) -> Result<bool, RedisError> {
         let mut res = None;
         self.do_op(path, |v| {
-            let val = v.to_bool().unwrap() ^ true;
-            res = Some(val);
-            Ok(Some(IValue::from(val)))
+            if let DestructuredMut::Bool(mut bool_mut) = v.destructure_mut() {
```
MeirShpilraien, line 379: Can you explain what `DestructuredMut` is for?

src/manager.rs
```diff
@@ -186,8 +186,16 @@ impl<'a> KeyHolderWrite<'a> {
         if paths.is_empty() {
             // updating the root require special treatment
             let root = self.get_value().unwrap().unwrap();
-            let res = (op_fun)(root.take())?;
-            self.set_root(res)?;
+            let prev_val = root.take();
+            let rollback_val = prev_val.clone();
```
MeirShpilraien, line 190: So you do use clone?

tests/pytest/test.py
```diff
@@ -645,22 +647,30 @@ def testArrInsertCommand(env):
...
+
+        r.assertEqual(r.execute_command('JSON.ARRINSERT', 'test', jpath, 1, '5'), 5)
+        r.assertEqual(r.execute_command('JSON.ARRINSERT', 'test', jpath, -2, '6'), 6)    
+        r.assertEqual(r.execute_command('JSON.GET', 'test', jpath), "[3,5,2,6,1,4]")
+        
+        r.assertEqual(r.execute_command('JSON.ARRINSERT', 'test', jpath, -3, '7', '{"A":"Z"}', '9'), 9)
+        r.assertEqual(r.execute_command('JSON.GET', 'test', jpath), '[3,5,2,7,{"A":"Z"},9,6,1,4]')
+
+        r.expect('JSON.ARRINSERT', 'test', jpath, -10, '10').raiseError()
+        if not IS_SERDE_JSON:
```
MeirShpilraien, line 666: So the fix is not working for serde json?

tests/pytest/test.py
```diff
@@ -645,22 +647,30 @@ def testArrInsertCommand(env):
...
+            # Value should remain
+            r.assertEqual(r.execute_command('JSON.GET', 'test', jpath), '[3,5,2,7,{"A":"Z"},9,6,1,4]')
+        r.expect('JSON.ARRINSERT', 'test', jpath, 10, '10').raiseError()
+        if not IS_SERDE_JSON:
```
MeirShpilraien, line 670: Same comment

MeirShpilraien (commented):
Looks good, added some comments. I also suggest to write the technical details of the solution on the top comment and explain why this solution is safe with regards to performance (no cloning, though I do see a place where you still clone, added a comment about that).

src/ivalue_manager.rs
oshadmi: Added explanation comment in `fn update`

src/ivalue_manager.rs
oshadmi: What about `update` and `calculate`?
(the caller is doing the actual replacement)

src/ivalue_manager.rs
oshadmi: It is an `Enum` used to obtain a mutable reference to `IValue`,
Specifically for Boolean, `IValue` does not directly contain Boolean,
So this is a proxy type which allows getting and setting the original `IValue` as Boolean.

src/manager.rs
oshadmi: This is outdated.
Checking where is commit ea44cf5 ?

tests/pytest/test.py
oshadmi: Not yet done for `serde_json`
After review is done for ijson, will proceed with `serde_json`.
Do we still want to invest in it BTW?

src/manager.rs
MeirShpilraien: How come its outdated and I see it in this PR?

src/ivalue_manager.rs
MeirShpilraien: Please document.

src/ivalue_manager.rs
MeirShpilraien: I believe that if something can not happened assert it, this way it will be easier to catch bugs.

src/ivalue_manager.rs
MeirShpilraien: Don't see it

src/ivalue_manager.rs
MeirShpilraien: How come the caller doing the replacement? The caller returns a new value and the update functions replace it no?

tests/pytest/test.py
MeirShpilraien: I believe we should still support serde json.

src/ivalue_manager.rs
MeirShpilraien: Lets explain this in a comment in the code and also on top comment on the PR.

src/ivalue_manager.rs
MeirShpilraien: Same comment as above.

src/manager.rs
oshadmi: This was a preliminary commit which was also addressing `serde_json`.
Now we currently putting `serde_json` aside.
I suggest to keep it anyway (no need to rollback IMHO)

src/ivalue_manager.rs
oshadmi: Will add in code (not only in GitHub comment)

src/ivalue_manager.rs
oshadmi: Will rename to `update` and `replace`

tests/pytest/test.py
oshadmi: If we decide to keep maintaining `serde_json` it will be in a different PR once `ijson` is done.

MeirShpilraien (approved):

PR #594, opened 2021-12-26
Handle "MEMORY USAGE" and JSON.DEBUG

gkorland:


src/ivalue_manager.rs
```diff
@@ -527,15 +526,23 @@ impl<'a> Manager for RedisIValueJsonKeyManager<'a> {
...
-        };
+        let res = size_of::<IValue>()
+            + match v.type_() {
+                ValueType::Null | ValueType::Bool | ValueType::Number => 0,
+                ValueType::String => v.as_string().unwrap().len(),
+                ValueType::Array => v
+                    .as_array()
+                    .unwrap()
+                    .into_iter()
+                    .map(|v| self.get_memory(v).unwrap())
```
oshadmi, line 537: Don't we want to count the raw vector itself (len, ptr and vector's capacity beyond its length)
(vector with zero elements does consume some memory)

src/ivalue_manager.rs
```diff
@@ -527,15 +526,23 @@ impl<'a> Manager for RedisIValueJsonKeyManager<'a> {
...
+                    .sum(),
+                ValueType::Object => v
+                    .as_object()
+                    .unwrap()
+                    .into_iter()
+                    .map(|(s, v)| s.len() + self.get_memory(v).unwrap())
```
oshadmi, line 543: Don't we want to count the size of the raw map itself?
(empty map does consume some memory)

tests/pytest/test_multi.py
```diff
@@ -785,31 +785,35 @@ def testToggleCommand(env):
...
+    r.assertEqual(res, [8])
 
     # Test legacy
     res = r.execute_command('JSON.DEBUG', 'MEMORY', 'doc1', '..a')
-    r.assertEqual(res, 72)
+    r.assertEqual(res, 8)
     # Test missing path (defaults to root)
     res = r.execute_command('JSON.DEBUG', 'MEMORY', 'doc1')
-    r.assertEqual(res, 72)
+    r.assertEqual(res, 179)
```
oshadmi, line 812: Can also compare to the result of `memory usage doc1` ?

MeirShpilraien (approved):

PR #597, opened 2021-12-27
Coverage support

rafie:


Makefile
```diff
@@ -106,10 +111,56 @@ CARGO_FLAGS += --release
 TARGET_DIR=$(BINDIR)/target/release
 endif
 
+ifeq ($(COV),1)
```
chayim, line 114: We should add comments as to why. There's no way anyone remembers this in the future.

Makefile
```diff
@@ -209,6 +260,25 @@ pack:
 
 #----------------------------------------------------------------------------------------------
 
+COV_EXCLUDE_DIRS += deps tests
```
chayim, line 263: Can these move into readies/mk/coverage* and maybe add a flag for USES_RUST or similar?

chayim (dismissed):

oshadmi (approved):
👍🏼

gkorland:
fix #406

PR #612, opened 2022-01-10
Adapt docs to JSON 2.0 syntax

gkorland:


K-Jo (approved):

K-Jo:
@gkorland actually, we might want to add an example for string scalar in json.set

K-Jo (approved):

PR #628, opened 2022-01-19
Added node-redis as a client with RedisJSON support.

simonprickett:
node-redis supports RedisJSON from v4.


gkorland:
@simonprickett notice it should also be added in the README

simonprickett:
@gkorland good catch - addressed that here https://github.com/simonprickett/RedisJSON/commit/c49d449f900e0f42d31f73875fa5826c9cab0da8

PR #629, opened 2022-01-20
Indexed array explainer

slorello89:
Noticed there was not any explenation as to how to index JSON arrays, adding that here - similar to redisearch/redisearch#2523


oshadmi (approved):
👍🏼

PR #638, opened 2022-02-01
oracle8 is the new centos8

chayim:


rafie (approved):

sbin/get-redisearch
```diff
@@ -64,8 +64,8 @@ else
 		if [[ $nick != bullseye ]]; then
 			nick=ubuntu18.04
 		fi
-	elif [[ $dist == centos || $dist == redhat || $dist == fedora ]]; then
-		if [[ $nick == centos8 ]]; then
+	elif [[ $dist == centos || $dist == redhat || $dist == fedora || $dist == ol ]]; then
```
DvirDukhan, line 67: `ol` without `nick == ol8` falls to `rhel7`
is this right? Does this flow exist?

tests/qa/qatests
```diff
@@ -28,7 +28,10 @@ RS_PLATFORMS = {
     'cenos7': {
         'platform': 'Linux-rhel7',
         'env': 'rhel7.7-amd64-aws' },
-    'cenos8': {
+    'centos8': {
```
DvirDukhan, line 31: why keep centos8 if it is replaced??

sbin/get-redisearch
rafie: Practically, we have two options for CentOS-like systems, rhel7 and rhel8 (note that those are RLEC-style names which we're forced to use in packages). The former will work on most systems (so unless we're a type-8 system - which falls under the category of osnick == centos8/ol8, because we don't distinguish between rhel8 and centos8 on the osnick level).
If you find it a conceptual hell, you're not wrong. It is.

tests/qa/qatests
rafie: It's not being replaced. I think once they fix it, it will be back. Note that 'centos8' is not just CentOS 8 literally, but the class of systems which are compatible with RHEL8. The point is that some systems, like Oracle Linux and Amazon Linux are not exactly 100% compatible with it (for instance, names of some packages may be different), so we keep them a little bit apart until we know them better. They do qualify for `redhat_compat` in `system-setup.py` though.

rafie (approved):

PR #656, opened 2022-02-13
Pinning rltest to the last release prior to redis-py 4 as a requirement

chayim:
This allows a longer-tail migration to the latest version as each repository is ready, without breaking individual projects.


gkorland:
Not relevant anymore

PR #664, opened 2022-02-22
New docker build and artifacts upload

rafie:


sbin/system-setup.py
```diff
@@ -55,6 +58,7 @@ def common_last(self):
 
 parser = argparse.ArgumentParser(description='Set up system for build.')
 parser.add_argument('-n', '--nop', action="store_true", help='no operation')
+parser.add_argument('--coverage', action="store_true", help='setup for coverage analisys (older CLang version)')
```
oshadmi, line 61:
```suggestion
parser.add_argument('--coverage', action="store_true", help='setup for coverage analysis (older CLang version)')
```

rafie: I should have removed the option altogether as we no longer check for it.

oshadmi (approved):

PR #678, opened 2022-02-27
Add test for issue 674

oshadmi:


tests/pytest/test_wildcards.py
```diff
@@ -0,0 +1,24 @@
...
+
+from RLTest import Defaults
+
+Defaults.decode_responses = True
+
+# ----------------------------------------------------------------------------------------------
+
+def testRecursiveDescentWithFilter(env):
+    with open('issue674.json') as file:
+        json_doc = file.read()
```
MeirShpilraien, line 20: Why not generating it by the test? Is it taking to long?

MeirShpilraien (commented):
Question...

tests/pytest/test_wildcards.py
oshadmi: We got it from the related issue creator, and it seems to reproduce the issue.
If we want some alternation or randomization, we could also generate on-the-fly,
But we would need to have some expected nested values somewhere.

zhangever: could you reproduce the issue? if needed, i could offer a bigger one which will take the jsonPath query for minutes.

oshadmi: @zhangever It seems to demonstrate the problem, so at least for now, there's no need for a bigger one, thank you.

PR #689, opened 2022-03-16
deps: Remove unnecessary features table

5HT2:
warning: the cargo feature `edition2021` has been stabilized in the 1.56 release and is no longer necessary to be listed in the manifest
See https://doc.rust-lang.org/cargo/reference/manifest.html#the-edition-field for more information about using this feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,3 @@
-cargo-features = ["edition2021"]
-
 [package]
 name = "redisjson"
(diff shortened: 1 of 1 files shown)


5HT2:
Oopsies, I did not realize that deleting a repo now also closes the pulls it opened. That didn't used to be the case..

I'll happily open a new PR if this is desired, but it is a two-line deletion.

PR #718, opened 2022-05-05
fixing link

slorello89:


docs/docs/indexing_JSON.md
```diff
@@ -168,7 +168,7 @@ FT.SEARCH userIdx '@name:(John)' RETURN 1 name HIGHLIGHT FIELDS 1 name TAGS '<b>
 
 ## Aggregation with JSON Path expression
 
-[Aggregation](/redisearch/aggregations) is a powerful feature. You can use it to generate statistics or build facet queries.
+[Aggregation](/docs/stack/search/reference/aggregations) is a powerful feature. You can use it to generate statistics or build facet queries.
```
oshadmi, line 171: Just adding missing `aggregations` hierarchy (the mapping should be done by the doc generator)
```suggestion
[Aggregation](/redisearch/reference/aggregations) is a powerful feature. You can use it to generate statistics or build facet queries.
```

slorello89: hmmm, that doesn't seem to work right - it will just bounce back to the root of the site, I see the proposed link working from the redisearch docs in the identical article (maybe we should de-duplicate this as a general rule and redirect to the redisearch doc). Regardless it doesn't seem to resolve correctly automatically.

oshadmi: @slorello89 The doc generator maps links from top-level `/redisearch/`, or other modules, to the actual hierarchy in the doc website, e.g., to `/docs/stack/search/`, and it is triggered on a commit [push](https://github.com/RediSearch/RediSearch/blob/017fafa4d3485bc24275fbb14749b9be68d77608/.github/workflows/trigger-build.yml#L5) (and I think [Hugo](https://gohugo.io/), the doc generator, then maps it to the actual URL).
Seems to me that the problem is the missing `reference` path hierarchy (not as I wrote in previous comment that the `aggregations` is missing) and as you wrote, a similar links is working properly, e.g., in [FT.AGGREGATE](https://redis.io/commands/ft.aggregate/)

oshadmi:
@itamarhaber This links is from RedisJSON referring to content in RediSearch.
Could there be a problem with generating cross-module links?

PR #736, opened 2022-06-01
Fix using the wrong manager from LLAPI

MeirShpilraien:
the LLAPI was using the wrong manager to access the json data. Share code with RedisJson commands to use the right manager.


oshadmi (approved):
👍🏼

PR #761, opened 2022-06-26
Add RedisJSON_V2 API with getPathInfoFlags

oshadmi:
Adding JSON API Version 2,
Currently with new APIs `pathIsStatic` and `pathHasDefinedOrder`

ReJSON is exporting both API Versions 1 and 2, so it could still be used by an older RediSearch module.

### DIFF
--- a/src/c_api.rs
+++ b/src/c_api.rs
@@ -472,21 +472,6 @@ macro_rules! redis_json_module_export_shared_api {
         };
 
         static JSONAPI_V2: RedisJSONAPI_V2 = RedisJSONAPI_V2 {
-            openKey: JSONAPI_openKey,
-            openKeyFromStr: JSONAPI_openKeyFromStr,
-            get: JSONAPI_get,
-            next: JSONAPI_next,
-            len: JSONAPI_len,
-            freeIter: JSONAPI_freeIter,
-            getAt: JSONAPI_getAt,
-            getLen: JSONAPI_getLen,
-            getType: JSONAPI_getType,
-            getInt: JSONAPI_getInt,
-            getDouble: JSONAPI_getDouble,
-            getBoolean: JSONAPI_getBoolean,
-            getString: JSONAPI_getString,
-            getJSON: JSONAPI_getJSON,
-            isJSON: JSONAPI_isJSON,
             pathParse: JSONAPI_pathParse,
             pathFree: JSONAPI_pathFree,
             pathIsStatic: JSONAPI_pathIsStatic,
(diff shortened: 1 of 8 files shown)


src/nodevisitor.rs
```diff
@@ -48,6 +59,19 @@ impl<'a> StaticPathParser<'a> {
         visitor.visit(&node);
         Ok(visitor)
     }
+
+    pub fn get_path_info(path: &'a str) -> PathInfoFlags {
+        Self::check(path).map_or_else(
+            |_| PathInfoFlags::INVALID,
+            |parser| {
+                if parser.valid == VisitStatus::Valid {
+                    PathInfoFlags::STATIC | PathInfoFlags::DEFINED_ORDER
```
oshadmi, line 68: @MeirShpilraien We currently do not differentiate between `STATIC` and `DEFINED_ORDERED`.
We currently do not accept, in `StaticPathParser`, Unions and closed positive Slices (Ranges),
Which could also be considered `STATIC` but might not be considered `DEFINED_ORDERED` if they appear more than once,
e.g., path with `[a,b][c,d]` could be traversed as `[a,c], [a,d], [b,c], [b,d]` or as `[a,c], [b,c], [a,d], [b,d]`,
So order is not well defined, although it is static.

src/include/rejson_api.h
```diff
@@ -65,6 +71,54 @@ typedef struct RedisJSONAPI_V1 {
 
 } RedisJSONAPI_V1;
 
+typedef struct RedisJSONAPI_V2 {
+  /* RedisJSONKey functions */
```
ashtul, line 75: `RedisJSONKey` => `RedisJSON`

src/c_api.rs
```diff
@@ -435,6 +451,25 @@ macro_rules! redis_json_module_export_shared_api {
             isJSON: JSONAPI_isJSON,
         };
 
+        static JSONAPI_V2: RedisJSONAPI_V2 = RedisJSONAPI_V2 {
```
MeirShpilraien, line 454: Why do we need to different structs here?  We only extended the struct so we can use the same struct for both.

src/c_api.rs
```diff
@@ -467,5 +502,39 @@ macro_rules! redis_json_module_export_shared_api {
             ) -> c_int,
             pub isJSON: extern "C" fn(key: *mut rawmod::RedisModuleKey) -> c_int,
         }
+
+        #[repr(C)]
+        #[derive(Copy, Clone)]
+        #[allow(non_snake_case)]
+        pub struct RedisJSONAPI_V2 {
```
MeirShpilraien, line 509: Same comment at above...

src/commands.rs
```diff
@@ -763,6 +763,9 @@ where
 /// And longer paths precede shorter paths
 /// And if a path is a sub-path of the other, then only paths with shallower hierarchy (closer to the top-level) remain
 fn prepare_paths_for_deletion(paths: &mut Vec<Vec<String>>) {
+    if paths.len() < 2 {
```
MeirShpilraien, line 766: Can you explain this addition? Is this a bug?

src/include/rejson_api.h
```diff
@@ -65,6 +71,54 @@ typedef struct RedisJSONAPI_V1 {
 
 } RedisJSONAPI_V1;
 
+typedef struct RedisJSONAPI_V2 {
```
MeirShpilraien, line 74: Put `RedisJSONAPI_V1` as the first element here instead of copy it

MeirShpilraien (commented):
Some comments union the API1 and API2 code.
I believe it is a good idea to separate the API to something like:
1. `JSONAPI_compilePath` -> will return a compiled object representing the path (or error)
2. `JSONAPI_pathIsStatic` -> will return whether or not a given path is static

In the future we will be able to extend the path API with more capabilities like `isFixed` ...

src/include/rejson_api.h
oshadmi: But it will not be flat, and I wanted to be able to call v1 APIs indifferently whether v1 or v2 API was actually loaded.

src/commands.rs
oshadmi: Minor efficiency improvement (not a bug - just avoid the overhead of re-ordering when less than 2 paths exist)

src/include/rejson_api.h
rafie: When we go C++ we can inherit, in the meantime we can play around with macros.

MeirShpilraien: You can return v2 api even if v2 is requested, v2 is just an extention on v1..

src/commands.rs
MeirShpilraien: So lets add a comment in the code explaining this optimization.

src/nodevisitor.rs
```diff
@@ -48,6 +65,14 @@ impl<'a> StaticPathParser<'a> {
         visitor.visit(&node);
         Ok(visitor)
     }
+
+    pub fn get_path_info(path: &'a str) -> Result<PathInfoFlags, String> {
+        let parser = Self::check(path)?;
+        match parser.valid {
+            VisitStatus::Valid => Ok(PathInfoFlags::STATIC | PathInfoFlags::DEFINED_ORDER),
+            _ => Ok(PathInfoFlags::INVALID),
```
MeirShpilraien, line 73: Why not error here? With some relevant error message? What the INVALID means?

MeirShpilraien (commented):
:+1: 
Last small comments

src/nodevisitor.rs
oshadmi: We return an error on path compilation failure (from `check` method)
`INVALID` was needed before `pathParse` could fail.
I will rename it now to `NONE` to mark some initial value.
Anyway, no need to expose the flags now, since there are dedicated APIs (`pathIsStatic`, `pathHasDefinedOrder`)

PR #767, opened 2022-07-08
Addressing some ambiguity with regards to multi-value indexing

slorello89:
The current explenation in the docs regarding multi-value indexing, and indexing within objects, is somewhat ambiguous, this PR is meant to spruce the docs up a bit so it removes this ambiguity and provides some clearer examples for this feature.

CC @K-Jo

### DIFF
--- a/docs/docs/indexing_JSON.md
+++ b/docs/docs/indexing_JSON.md
@@ -79,37 +79,50 @@ FT.SEARCH userIdx '@name:(John)'
    2) "{\"user\":{\"name\":\"John Smith\",\"tag\":\"foo,bar\",\"hp\":1000,\"dmg\":150}}"
 ```
 
-## Indexing JSON arrays with tags
+## Indexing JSON arrays
 
-It is possible to index scalar string and boolean values in JSON arrays by using the wildcard operator in the JSON Path. For example if you were indexing blog posts you might have a field called `tags` which is an array of tags that apply to the blog post.
+Any JSON path that resolves to an array of strings or booleans is indexable as a TAG field.
+This means that we can index any JSON array of strings or booleans, as well as any objects
+we can derive a JSON path from to resolve to an array of strings or booleans.
 
-```JSON
+Take the following example, which has an array of string (`phone_numbers`), and an array of objects `addresses`:
+
+```json
 {
-   "title":"Using RedisJson is Easy and Fun",
-   "tags":["redis","json","redisjson"]
+   "name":"Bob Fry",
+   "phone_numbers":
+   [
+      "555-555-5555",
+      "123-456-7890"
+   ],
+   "addresses":
+   [
+      {
+         "street_address":"123 Main Street",
+         "city":"Orlando",
+         "state":"Florida",
+         "postal_code":"32789"
+      },
+      {
+         "street_address":"234 Drury Lane",
+         "city":"Atlanta",
+         "state":"Georgia",
+         "postal_code":"30301"
+      }
+   ]
 }
 ```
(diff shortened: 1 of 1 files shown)


slorello89:
Corresponding RediSearch PR: RediSearch/RediSearch#2880

docs/docs/indexing_JSON.md
```diff
@@ -79,37 +79,50 @@ FT.SEARCH userIdx '@name:(John)'
    2) "{\"user\":{\"name\":\"John Smith\",\"tag\":\"foo,bar\",\"hp\":1000,\"dmg\":150}}"
 ```
 
-## Indexing JSON arrays with tags
+## Indexing JSON arrays
 
-It is possible to index scalar string and boolean values in JSON arrays by using the wildcard operator in the JSON Path. For example if you were indexing blog posts you might have a field called `tags` which is an array of tags that apply to the blog post.
+Any JSON path that resolves to an array of strings or booleans is indexable as a TAG field.
+This means that we can index any JSON array of strings or booleans, as well as any objects
+we can derive a JSON path from to resolve to an array of strings or booleans.
```
oshadmi, line 86:
```suggestion
This means that we can index any JSON array of strings, as well as any objects
we can derive a JSON path from, to resolve to an array of strings.
```

oshadmi (commented):
Looks good!
Just a few comments

docs/docs/indexing_JSON.md
```diff
@@ -186,54 +199,71 @@ FT.AGGREGATE userIdx '*' LOAD 6 $.user.hp AS hp $.user.dmg AS dmg APPLY '@hp-@dm
...
+   },
+   "phone_numbers":[
+      "555-555-5555",
+      "123-456-7890"
+   ]
 }
 ```
 
-Before you can index the array under the `address` key, you have to create two fields:
+The path `$.address` returns the whole object for address, so this cannot be indexed. However, you can index the components of `address` with a JSON path resolving to the scalars within it e.g. `$.address.unit`. Additionally if you wanted to create a rule where you'd consider the first element of the `phone_numbers` array the primary phone number, you could use the path $.phone_numbers[0] to resolve to that item:
```
nermiller, line 235:
```suggestion
The path `$.address` returns the whole object for address, so this cannot be indexed. However, you can index the components of `address` with a JSON path resolving to the scalars within it, for example, `$.address.unit`. Additionally, if you want to create a rule where you consider the first element of the `phone_numbers` array the primary phone number, you can use the path $.phone_numbers[0] to resolve to that item:
```

nermiller (approved):
LGTM with some minor edits

docs/docs/indexing_JSON.md
```diff
@@ -79,37 +79,50 @@ FT.SEARCH userIdx '@name:(John)'
...
+         "city":"Atlanta",
+         "state":"Georgia",
+         "postal_code":"30301"
+      }
+   ]
 }
 ```
 
-You can apply an index to the `tags` field by specifying the JSON Path `$.tags.*` in your schema creation:
+You can index any JSON path that will resolve to an array. For example, `$.phone_numbers.*` resolves to the array `['555-555-5555', 123-456-7890]` so it can be indexed. Also, the string fields WITHIN the objects of the `addresses` field can also be indexed. For example, the JSON path `$.addresses[*].city` resolves to an array of strings `[Orlando, Atlanta]`. Consequentially, you can create a corresponding index to index both of these fields:
```
guyroyse, line 116: Missing quotes on the second phone number.

PR #774, opened 2022-07-14
avoid unwrap/panic when can return with error

gkorland:


src/redisjson.rs
```diff
@@ -188,7 +188,7 @@ pub mod type_methods {
                 let data = raw::load_string(rdb)?;
                 data.try_as_str()?.to_string()
             }
-            _ => panic!("Can't load old RedisJSON RDB"),
+            _ => return Err("Can't load old RedisJSON RDB".into()),
```
oshadmi, line 191: Caller (`rdb_load`) should log some error, since now without panic, it will just return null and the error message would be gone.
Currently:
`Err(_) => null_mut(),`

PR #784, opened 2022-07-24
Add test for JSON object nesting depth

oshadmi:
Limit is from [serde_json](https://github.com/serde-rs/json/blob/dab5ed3ee97cef5e2b796513f8d9e4c7416e44bf/src/error.rs#L245) lib


MeirShpilraien (approved):

docs/docs/_index.md
```diff
@@ -173,6 +173,11 @@ Lastly, you can also use the [`MODULE LOAD`](/commands/module-load) command. Not
 
 Once the module has been loaded successfully, the Redis log should have lines similar to:
 
+
+### Limitation
```
nermiller, line 177: Should this limitation go after the log lines?

docs/docs/_index.md
```diff
@@ -173,6 +173,11 @@ Lastly, you can also use the [`MODULE LOAD`](/commands/module-load) command. Not
...
+
+A JSON value passed to a command can have a depth of up to 128. When passing to a command a JSON value containing an object or an array with a nesting level of more than 128, then the command will return an error.
```
nermiller, line 179:
```suggestion
A JSON value passed to a command can have a depth of up to 128. If you pass to a command a JSON value that contains an object or an array with a nesting level of more than 128, the command returns an error.
```

nermiller (changes requested):
Just some minor feedback

tests/pytest/test.py
```diff
@@ -1114,6 +1115,46 @@ def testCopyCommand(env):
     r.assertEqual(json.loads(res), [new_values])
 
 
+def nest_object(depth, max_name_len, leaf_key, leaf_val):
```
MeirShpilraien, line 1118: Why not build it as python dict and then decode it as json?

tests/pytest/test.py
```diff
@@ -1114,6 +1115,46 @@ def testCopyCommand(env):
...
+    res = r.execute_command('JSON.GET', 'test', '$..__leaf')
+    r.assertEqual(res, '[42]')
+
+    # No overall max nesting level (can exceeded the single value max nesting level)
+    doc = nest_object(depth, 5, "__deep_leaf", 420)
+    r.execute_command('JSON.SET', 'test', '$..__leaf', doc)
+    res = r.execute_command('JSON.GET', 'test', '$..__deep_leaf')
+    r.assertEqual(res, '[420]')
+
+    doc = nest_object(depth, 5, "__helms_deep_leaf", 42000)
```
MeirShpilraien, line 1146: So we can create as deep nesting object as we want but not in a single `json.set` command? Is that right? Should we have some total limit as well (Maybe configurable)?

MeirShpilraien (commented):
Small comments

tests/pytest/test.py
oshadmi: Initially tried with python dict but It was throwing some exception

tests/pytest/test.py
oshadmi: Yes, the `test` doc will exceed the limit.
The limit is coming from `serde_json` when we deserialize the JSON value argument in the command.
Why limit if we don't have to? I suppose few users, if any, would get here.
Seems enough for me to document and add a test.

tests/pytest/test.py
MeirShpilraien: So lets understand the exception and fix them...

tests/pytest/test.py
MeirShpilraien: Lets discuss this on another issue...

tests/pytest/test.py
oshadmi: Fixed

MeirShpilraien (approved):

PR #832, opened 2022-10-06
C test module working. Rust test module in development

efeldblum-redis:


oshadmi:
@efeldblum-redis Regarding building a shared .so library in Rust, need to set the `crate-type` to `cdylib`, such as in the Cargo.toml file.
See some [doc](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-crate-type-field) and as reference, see RedisJSON [Cargo.toml](https://github.com/RedisJSON/RedisJSON/blob/eddfcb9f47409d5f1e514ee9f164b4d7d980d094/Cargo.toml#L16)

ephraimfeldblum:
to run tests using Rust (or C) library:
```sh
cd tests/module
make run [C=1]
```

tests/module/src/include/rejson_api.h
```diff
@@ -0,0 +1,100 @@
+#pragma once
```
oshadmi, line 1: Should use the existing header at src/include/rejson_api.h

tests/module/src/testLLAPI.c
```diff
@@ -0,0 +1,249 @@
+
```
oshadmi, line 1: Remove and ignore testLLAPI.o

tests/module/Cargo.lock
```diff
@@ -0,0 +1,1327 @@
+# This file is automatically @generated by Cargo.
```
oshadmi, line 1: With a Cargo.lock file we will not take the latest RedisJSON (unless doing a `cargo update` before build)
Seems we can just remove and ignore the lock.

oshadmi:
Seems we can remove and ignore binaries such as libtestmod.so and dump.rdb

tests/module/src/include/redismodule.h
```diff
@@ -0,0 +1,1177 @@
+#ifndef REDISMODULE_H
```
oshadmi, line 1: Better use existing header at src/redismodule.h

PR #848, opened 2022-11-02
Add to JSONPath filter the regexp match operator

oshadmi:
The syntax already supports a `=~` operator, but with empty implementation.
Adding implementation for matching a value with a given regular expression string.

For example, filter array values containing `foo`
```
127.0.0.1:6379> JSON.SET key $ '{"arr": ["kaboom", "kafoosh", "four", "bar", "foolish", "ffool"]}'
OK

127.0.0.1:6379> JSON.GET key '$.arr[?(@ =~ ".*foo")]'
"[\"kafoosh\",\"foolish\",\"ffool\"]"

```

### DIFF
--- a/src/jsonpath/json_path.rs
+++ b/src/jsonpath/json_path.rs
@@ -3,6 +3,7 @@ use pest::Parser;
 use std::cmp::Ordering;
 
 use crate::jsonpath::select_value::{SelectValue, SelectValueType};
+use regex::Regex;
 use std::fmt::Debug;
 
 #[derive(Parser)]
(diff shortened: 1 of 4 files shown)


src/jsonpath/json_path.rs
```diff
@@ -433,8 +434,24 @@ impl<'i, 'j, S: SelectValue> TermEvaluationResult<'i, 'j, S> {
...
 
-    fn re(&self, _s: &Self) -> bool {
-        false
+    fn re_is_match(regex: &str, s: &str) -> bool {
+        Regex::new(regex).map_or_else(|_| false, |re| Regex::is_match(&re, s))
+    }
+
+    fn re_match(&self, s: &Self) -> bool {
+        match (self, s) {
+            (TermEvaluationResult::Value(v), TermEvaluationResult::Str(regex)) => {
```
MeirShpilraien, line 443: Can't the regex value be another part of the json? I mean can't it by itself be a Value?
Also can't the right side be a Str?
Also, what about float? int? did you check how they behave?

MeirShpilraien (commented):
One comment about handing value other then `Value` and `Str`.
Also let add an end to end test.

MeirShpilraien:
Also lets update docs.

src/jsonpath/json_path.rs
oshadmi: Matching against a regexp which is stored in a JSON field seems a more advanced use case to me
And I left it aside for now. We can add it when needed, unless you find it mandatory.

Numbers currently do not match since we would need to convert them to strings, and, in addition, it can be done by ORing with numerical predicates (test `op_string_regexp_match` is filtering out numbers)

MeirShpilraien: I think it should be feature complete before merge so we should check those stuff.

MeirShpilraien (approved):
Looks good,
Please add to the top comment the details about what behaviure we should expect in each case (string, json value, number,...)
Also lets state the differeneces between us and the other common engines and maybe some words about why we believe those differences are ok.

MeirShpilraien (approved):

Issue #875, opened 2022-11-27
Redis sentinel with Redis JSON and Redisearch crashed - crash report


nadavtsa:
My use case: 
I'm running a redis sentinel on k8s deployment (using redis-ha template). I compiled the Redis-JSOn and Redisearch modules manually inside the docker image (redis version - 6.2.7)

Crash report:
1:M 27 Nov 2022 11:59:34.830 # Redis 6.2.7 crashed by signal: 11, si_code: 128
1:M 27 Nov 2022 11:59:34.830 # Accessing address: (nil)
1:M 27 Nov 2022 11:59:34.830 # Crashed running the instruction at: 0x7f5f86415602

------ STACK TRACE ------
EIP:
/lib/x86_64-linux-gnu/libc.so.6(abort+0x1ee)[0x7f5f86415602]

Backtrace:
/lib/x86_64-linux-gnu/libpthread.so.0(+0x13140)[0x7f5f865db140]
/lib/x86_64-linux-gnu/libc.so.6(abort+0x1ee)[0x7f5f86415602]
/usr/lib/redis/modules/rejson.so(+0x11d507)[0x7f5f8636b507]
/usr/lib/redis/modules/rejson.so(+0x11b80e)[0x7f5f8636980e]
/usr/lib/redis/modules/rejson.so(+0x11b5a7)[0x7f5f863695a7]
/usr/lib/redis/modules/rejson.so(+0x11b359)[0x7f5f86369359]
/usr/lib/redis/modules/rejson.so(+0x11a264)[0x7f5f86368264]
/usr/lib/redis/modules/rejson.so(+0x11b0c9)[0x7f5f863690c9]
/usr/lib/redis/modules/rejson.so(+0x17b43)[0x7f5f86265b43]
/usr/lib/redis/modules/rejson.so(+0x17a0d)[0x7f5f86265a0d]
/usr/lib/redis/modules/rejson.so(+0xcb9a2)[0x7f5f863199a2]
/usr/lib/redis/modules/rejson.so(JSONAPI_getDouble+0x54)[0x7f5f862ad934]
/usr/lib/redis/modules/redisearch.so(JSON_StoreNumericInDocField+0xd4)[0x7f5f856e8354]
/usr/lib/redis/modules/redisearch.so(JSON_StoreNumericInDocFieldFromArr+0x47)[0x7f5f856e84c7]
[...]

PR #883, opened 2022-12-11
MOD-4551: Workaround indexing u64 integer greater than i64::MAX

oshadmi:
Currently there is an API `get_long(&self) -> i64`
Workaround to avoid panic on unwrap when an integer value greater than `i64::MAX` is indexed by RediSearch.
Workaround is needed until an API `get_ulong(&self) -> u64` is added to `SelectValue` trait.

Related #875

### DIFF
--- a/tests/pytest/test.py
+++ b/tests/pytest/test.py
@@ -1257,6 +1257,53 @@ def testFilter(env):
     r.expect('JSON.GET', 'doc', '$.arr[?(@ == $.pat_plain)]').equal('["(?i)^[f][o][o]$"]')
     
 
+def testOutOfRangeValues(env):
+    # Test numeric min/max values
+    r = env
+
+    doc_int_ok = {
+        # i64
+        "max_i64": 9223372036854775807,
+        "beyond_max_i64": 9223372036854775808, # as u64
+        "min_i64": -9223372036854775808,
+        "below_min_i64": -9223372036854775809, # as f64
+        # u64
+        "max_u64": 18446744073709551615,
+        "beyond_max_u64": 18446744073709551616, # as f64
+    }
+
+    doc_float_ok = {
+        "max_f64": 1.7976931348623157e+308,
+        "beyond_max_f64": 1.7976931348623158e+308, # why OK?
+        "min_f64": -1.7976931348623157e+308,
+        "below_min_f64": -1.7976931348623158e+308, # why OK?
+    }
+
+    doc_bad_values = {
+        "beyond1_max_f64": 1.7976931348623159e+308,
+        "below1_min_f64": -1.7976931348623159e+308
+    }
+    
+    r.expect('JSON.SET', 'doc_int_ok', '$', json.dumps(doc_int_ok)).ok()
+    r.expect('JSON.SET', 'doc_float_ok', '$', json.dumps(doc_float_ok)).ok()
+    epsilon = 0
+    for k, v in iter(doc_int_ok.items()):
+        r.assertTrue(True, message='GET {}={}'.format(k, v))
+        res = r.execute_command('JSON.GET', 'doc_int_ok', '$.{}'.format(k))
+        r.assertAlmostEqual(json.loads(res)[0], v, epsilon, message=res)
+    
+    epsilon = sys.float_info.epsilon
+    for k, v in iter(doc_float_ok.items()):
+        r.assertTrue(True, message='GET {}={}'.format(k, v))
+        res = r.execute_command('JSON.GET', 'doc_float_ok', '$.{}'.format(k))
+        r.assertAlmostEqual(json.loads(res)[0], v, epsilon, message=res)
+    
+    # Do not use json.dumps with out-of-range values here (would be converted to a string representation such as 'Infinity')
+    r.expect('JSON.SET', 'doc_bad_values', '$', '{}').ok()
+    for k, v in iter(doc_bad_values.items()):
+        r.assertTrue(True, message='SET {}={}'.format(k, v))
+        r.expect('JSON.SET', 'doc_bad_values', '$.{}'.format(k), '{}'.format(v)).error()
+
 # class CacheTestCase(BaseReJSONTest):
 #     @property
 #     def module_args(env):
--- a/src/jsonpath/json_node.rs
+++ b/src/jsonpath/json_node.rs
@@ -250,7 +250,12 @@ impl SelectValue for IValue {
                 if n.has_decimal_point() {
                     panic!("not a long");
                 } else {
-                    n.to_i64().unwrap()
+                    if let Some(n) = n.to_i64() {
+                        n
+                    } else {
+                        // Workaround until adding an API `get_ulong(&self) -> u64`
+                        i64::MAX
+                    }
                 }
             }
             _ => {


MeirShpilraien (dismissed):
Maybe we should document it?

oshadmi:
> Maybe we should document it?

Sure.
@nermiller Should we document this workaround and limitation in the release notes, in the documentation itself or both?

nermiller:
@oshadmi both would be good until the limitation is removed. Let's work on adding a note.

tests/pytest/test.py
```diff
@@ -1280,6 +1280,87 @@ def testFilterPrecedence(env):
...
+    r.expect('JSON.SET', 'doc_int_ok', '$', json.dumps(doc_int_ok)).ok()
+    r.expect('JSON.SET', 'doc_float_ok', '$', json.dumps(doc_float_ok)).ok()
+    r.expect('JSON.SET', 'doc_float_ok_2', '$', json.dumps(doc_float_ok_2)).ok()
+
+    def check_object_values(obj, name, epsilon):
+        # Test values from JSON.GET are equal to JSON.SET
+        # Check no crash (using get_long internally)
+        arr = []
+        for k, v in iter(obj.items()):
+            r.assertTrue(True, message='GET {}={}'.format(k, v))
```
MeirShpilraien, line 1331: ?

tests/pytest/test.py
```diff
@@ -1280,6 +1280,87 @@ def testFilterPrecedence(env):
...
+            res = r.execute_command('JSON.GET', name, '$.{}'.format(k))
+            r.assertAlmostEqual(json.loads(res)[0], v, epsilon, message=res)
+            try:
+                # Check no crash on overflow
+                res = r.execute_command('JSON.NUMINCRBY', name, '$.{}'.format(k), '1')
```
MeirShpilraien, line 1336: Can we verify the result?

tests/pytest/test.py
```diff
@@ -1280,6 +1280,87 @@ def testFilterPrecedence(env):
...
+                #r.assertAlmostEqual(json.loads(res)[0], v+1, epsilon, message=res)
+                res = r.execute_command('JSON.NUMMULTBY', name, '$.{}'.format(k), '2')            
+                #r.assertAlmostEqual(json.loads(res)[0], v*2, epsilon, message=res)
+            except Exception as e:
+                pass
+            arr.append(v)
+        # Check no crash using values in a filter
+        r.expect('JSON.SET', 'arr', '$', json.dumps(arr)).ok()
+        res = r.execute_command('JSON.GET', 'arr', '$[?(@!=0)]')
```
MeirShpilraien, line 1345: Not following, shouldn't this return error now?

tests/pytest/test.py
```diff
@@ -1280,6 +1280,87 @@ def testFilterPrecedence(env):
...
+        for res_i, arr_i in zip(json.loads(res), arr):
+            r.assertAlmostEqual(res_i, arr_i, epsilon, message=res)
+        # Check no crash with other commands 
+        r.execute_command('JSON.ARRINDEX', 'arr', '$', '10')
```
MeirShpilraien, line 1349: Verify reply?

tests/pytest/test.py
```diff
@@ -1280,6 +1280,87 @@ def testFilterPrecedence(env):
...
+        r.execute_command('JSON.RESP', 'arr')
```
MeirShpilraien, line 1350: Verify reply

tests/pytest/test.py
```diff
@@ -1280,6 +1280,87 @@ def testFilterPrecedence(env):
...
+            r.execute_command('JSON.CLEAR', name, '$.{}'.format(k))
+
+    check_object_values(doc_int_ok, 'doc_int_ok', 0)
+    check_object_values(doc_float_ok, 'doc_float_ok', sys.float_info.epsilon)
+    check_object_values(doc_float_ok_2, 'doc_float_ok_2', sys.float_info.epsilon)
+
+    # Not using json.dumps with out-of-range values here (would be converted to a string representation such as 'Infinity')
+    r.expect('JSON.SET', 'doc_bad_values', '$', '{}').ok()
+    for k, v in iter(doc_bad_values.items()):
+        r.assertTrue(True, message='SET {}={}'.format(k, v))
```
MeirShpilraien, line 1361: ?

MeirShpilraien (commented):
I do not see where the new error `currently unsupported` are checked on the tests.

MeirShpilraien:
Also lets update top comment.

tests/pytest/test.py
oshadmi: This is just printing in order to see the current value in the loop

tests/pytest/test.py
oshadmi: Some fail due to overflow. Could be done such as by moving values that overflow to separate keys.

tests/pytest/test.py
oshadmi: Did not propagate the error all the way up.
A failed term (a long which is u64) is filtered out since evaluated as `NotCmparable`

tests/pytest/test.py
oshadmi: This is just printing in order to see the current value in the loop

tests/pytest/test.py
MeirShpilraien: We have debugPrint for that
https://github.com/RedisLabsModules/RLTest/blob/09ced571f7d706f39c6a505d84ea2ebb05653b9a/RLTest/env.py#L58

tests/pytest/test.py
MeirShpilraien: What do you mean fail? Do we get an error? If so, isn't it a breaking change?
And in any case we can verfit the error.

tests/pytest/test.py
MeirShpilraien: So we need to decide if this might be considered a breaking change in the future? @gkorland ?

tests/pytest/test.py
MeirShpilraien: Debugprint

src/commands.rs
```diff
@@ -128,7 +128,9 @@ impl<'a, V: SelectValue + 'a> KeyValue<'a, V> {
                 }
             }
 
-            SelectValueType::Long => RedisValue::Integer(v.get_long()),
+            SelectValueType::Long => {
+                RedisValue::Integer(v.get_long().map_or_else(|_| i64::MAX, |v| v))
```
MeirShpilraien, line 132: Isn't this can be consider a breaking change in the future? Can we raise error?

src/commands.rs
```diff
@@ -417,7 +419,10 @@ impl<'a, V: SelectValue + 'a> KeyValue<'a, V> {
         match (a.get_type(), b.get_type()) {
             (SelectValueType::Null, SelectValueType::Null) => true,
             (SelectValueType::Bool, SelectValueType::Bool) => a.get_bool() == b.get_bool(),
-            (SelectValueType::Long, SelectValueType::Long) => a.get_long() == b.get_long(),
+            (SelectValueType::Long, SelectValueType::Long) => match (a.get_long(), b.get_long()) {
+                (Ok(a), Ok(b)) => a == b,
+                _ => false,
```
MeirShpilraien, line 424: Same question here, wouldn't it will be better to raise error? @gkorland ?

src/commands.rs
```diff
@@ -1933,7 +1938,7 @@ pub fn json_clear<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>)
 
     let paths = find_paths(path, root, |v| match v.get_type() {
         SelectValueType::Array | SelectValueType::Object => v.len().unwrap() > 0,
-        SelectValueType::Long => v.get_long() != 0,
+        SelectValueType::Long => v.get_long().map_or_else(|_| false, |v| v != 0),
```
MeirShpilraien, line 1941: Don't we want to clear it here anyway?

src/ivalue_manager.rs
```diff
@@ -219,7 +220,7 @@ impl<'a> IValueKeyHolderWrite<'a> {
         mut op2_fun: F2,
     ) -> Result<Number, RedisError>
     where
-        F1: FnMut(i64, i64) -> i64,
+        F1: FnMut(i64, i64) -> (i64, bool),
```
MeirShpilraien, line 223: Whats the bool is used for?

tests/pytest/test.py
```diff
@@ -1280,6 +1280,123 @@ def testFilterPrecedence(env):
...
+        for k, v in iter(obj.items()):
+            r.assertTrue(True, message='{} GET {}={}'.format(name, k, v))
+            res = r.execute_command('JSON.GET', name, '$.{}'.format(k))
+            r.assertAlmostEqual(json.loads(res)[0], v, epsilon, message=res)
+            try:
+                # Check no crash on overflow
+                res = r.execute_command('JSON.NUMINCRBY', name, '$.{}'.format(k), '1')
+                if check_num_incr:
+                    r.assertAlmostEqual(json.loads(res)[0], v+1, epsilon, message=res)
+                r.execute_command('JSON.NUMINCRBY', name, '$.{}'.format(k), '9223372036854775808')
```
MeirShpilraien, line 1348: I am not follow, if one of the checks raise error we are not checking the rest? What am I missing?

tests/pytest/test.py
```diff
@@ -1280,6 +1280,123 @@ def testFilterPrecedence(env):
...
+        res = r.execute_command('JSON.RESP', 'arr')
+        for res_i, arr_i in zip(res[1:], arr):
+            if type(res_i) is int:
+                # Integers greater then max i64 are returned as max i64 by RESP
+                r.assertAlmostEqual(res_i, arr_i if arr_i <= sys.maxsize else sys.maxsize, epsilon, message=res)
+            elif type(res_i) is str:
+                # Double are returned as strings
+                r.assertAlmostEqual(float(res_i), arr_i, epsilon, message=res)
+        for k, v in iter(obj.items()):            
+            r.execute_command('JSON.CLEAR', name, '$.{}'.format(k))
```
MeirShpilraien, line 1383: Check result?

tests/pytest/test.py
```diff
@@ -1280,6 +1280,123 @@ def testFilterPrecedence(env):
...
+    check_object_values(doc_float_ok, 'doc_float_ok', sys.float_info.epsilon)
+    check_object_values(doc_float_ok_overflow_mult, 'doc_float_ok_overflow_mult', 0, True, False)
+
+    check_object_values(doc_float_ok_2_overflow_mult, 'doc_float_ok_2_overflow_mult', sys.float_info.epsilon, True, False)
+    
+    # Not using json.dumps with out-of-range values here (would be converted to a string representation such as 'Infinity')
+    r.expect('JSON.SET', 'doc_bad_values', '$', '{}').ok()
+    for k, v in iter(doc_bad_values.items()):
+        r.assertTrue(True, message='SET {}={}'.format(k, v))
+        r.expect('JSON.SET', 'doc_bad_values', '$.{}'.format(k), '{}'.format(v)).error()
```
MeirShpilraien, line 1398:
```suggestion
        r.expect('JSON.SET', 'doc_bad_values', '$.{}'.format(k), '{}'.format(v)).error().contians(<some substring from the expected error>)
```

oshadmi: This errors are JSON parse errors, such as with these commands
```
"JSON.SET" "doc_bad_values" "$.beyond_max_f64" "inf"
"JSON.SET" "doc_bad_values" "$.below_min_f64" "-inf"
```
these error messages are issued:
`expected value at line 1 column 1`
`invalid number at line 1 column 2`

tests/pytest/test.py
oshadmi: It is not enabled by default
`--debug-print         print debug messages (default: False)`

tests/pytest/test.py
oshadmi: The verification will fail, e.g., the result of NUMINCRBY will not be `v+1`

tests/pytest/test.py
oshadmi: Not enabled by default

src/commands.rs
oshadmi: Yes, fixing it.

src/ivalue_manager.rs
oshadmi: The bool returns whether there was an overflow.
Using `overflowing_add` and `overflowing_mul` returning the wrapped value if overflowed (should be the same result as with Release builds, only not skipping the overflow check)

tests/pytest/test.py
oshadmi: Right, seemed enough to check first overflow/unsupported value, but we can test all.

src/c_api.rs
```diff
@@ -177,6 +191,19 @@ pub fn json_api_get_double<M: Manager>(_: M, json: *const c_void, val: *mut c_do
             unsafe { *val = json.get_long() as f64 };
             Status::Ok as c_int
         }
+
+        SelectValueType::ULong => {
```
MeirShpilraien, line 195: Why would someone call get_double if the value is ulong? I see we did it for long as well but I wonder if its not an API missuse?

src/commands.rs
```diff
@@ -418,6 +425,7 @@ impl<'a, V: SelectValue + 'a> KeyValue<'a, V> {
             (SelectValueType::Null, SelectValueType::Null) => true,
             (SelectValueType::Bool, SelectValueType::Bool) => a.get_bool() == b.get_bool(),
             (SelectValueType::Long, SelectValueType::Long) => a.get_long() == b.get_long(),
+            (SelectValueType::ULong, SelectValueType::ULong) => a.get_ulong() == b.get_ulong(),
```
MeirShpilraien, line 428: So ulong will never be equal to long? I guess this claim also true when comparing long to double, but maybe here it will be more noticeable?

src/include/rejson_api.h
```diff
@@ -20,7 +20,11 @@ typedef enum JSONType {
   JSONType_Object = 4,
   JSONType_Array = 5,
   JSONType_Null = 6,
-  JSONType__EOF
+  JSONType__EOF = 7, // keep this for backward compatibility
```
MeirShpilraien, line 23: What is this value means and where is it used?

src/ivalue_manager.rs
```diff
@@ -382,15 +442,33 @@ impl<'a> WriteHolder<IValue, IValue> for IValueKeyHolderWrite<'a> {
     }
 
     fn incr_by(&mut self, path: Vec<String>, num: &str) -> Result<Number, RedisError> {
-        self.do_num_op(path, num, |i1, i2| i1 + i2, |f1, f2| f1 + f2)
+        self.do_num_op(
+            path,
+            num,
+            |i1, i2| i1.overflowing_add(i2),
```
MeirShpilraien, line 448: Can this really happened now? If it will overflow we will move to u64 no?

src/ivalue_manager.rs
```diff
@@ -382,15 +442,33 @@ impl<'a> WriteHolder<IValue, IValue> for IValueKeyHolderWrite<'a> {
...
+            |u1, u2| u1.overflowing_add(u2),
+        )
     }
 
     fn mult_by(&mut self, path: Vec<String>, num: &str) -> Result<Number, RedisError> {
-        self.do_num_op(path, num, |i1, i2| i1 * i2, |f1, f2| f1 * f2)
+        self.do_num_op(
+            path,
+            num,
+            |i1, i2| i1.overflowing_mul(i2),
```
MeirShpilraien, line 458: Same question as above

src/ivalue_manager.rs
```diff
@@ -382,15 +442,33 @@ impl<'a> WriteHolder<IValue, IValue> for IValueKeyHolderWrite<'a> {
...
+            |u1, u2| u1.overflowing_mul(u2),
+        )
     }
 
     fn pow_by(&mut self, path: Vec<String>, num: &str) -> Result<Number, RedisError> {
-        self.do_num_op(path, num, |i1, i2| i1.pow(i2 as u32), f64::powf)
+        self.do_num_op(
+            path,
+            num,
+            |i1, i2| i1.overflowing_pow(i2 as u32),
```
MeirShpilraien, line 468: Same question as above

src/c_api.rs
oshadmi: Basically this is an optimization, for example, an array of doubles, in which some may be integers, not requiring to consult the type of each, or avoid [getDouble](https://github.com/RediSearch/RediSearch/blob/4387b045b764c7bfc802646dd0b8af52845b56b3/src/json.c#L143) to fail and then re-try as int.

tests/pytest/test.py
```diff
@@ -1280,6 +1280,156 @@ def testFilterPrecedence(env):
...
+        for k, v in iter(obj.items()):
+            r.assertTrue(True, message='{} GET {}={}'.format(name, k, v))
+            res = r.execute_command('JSON.GET', name, '$.{}'.format(k))
+            r.assertAlmostEqual(json.loads(res)[0], v, epsilon, message=res)
+            
+            values = ['9223372036854775808', '18446744073709551615', '18446744073709551616', '-9223372036854775808', '-18446744073709551615', '-18446744073709551616']
+            
+            # Check NUMINCRBY
+            try:
+                r.assertTrue(True, message='{} NUMINCRBY {}: {} + 1'.format(name, k, v))
```
MeirShpilraien, line 1353: If this is only for debug prints please use r.printDebug (or debugPrint, I do not remember exactly)..

src/include/rejson_api.h
oshadmi: Used by RediSearch `JSONType`. Marks a non-existing type (maybe it was wrong to introduce it, but now it exists)

tests/pytest/test.py
```diff
@@ -1280,6 +1280,156 @@ def testFilterPrecedence(env):
...
+    r.expect('JSON.SET', 'doc_int_ok_overflow_mult', '$', json.dumps(doc_int_ok_overflow_mult)).ok()
+    r.expect('JSON.SET', 'doc_int_ok_overflow_incr', '$', json.dumps(doc_int_ok_overflow_incr)).ok()
+    r.expect('JSON.SET', 'doc_int_ok_overflow_both', '$', json.dumps(doc_int_ok_overflow_both)).ok()
+
+    r.expect('JSON.SET', 'doc_float_ok', '$', json.dumps(doc_float_ok)).ok()
+    r.expect('JSON.SET', 'doc_float_ok_overflow_incr', '$', json.dumps(doc_float_ok_overflow_incr)).ok()
+
+    r.expect('JSON.SET', 'doc_float_ok_2_overflow_incr', '$', json.dumps(doc_float_ok_2_overflow_incr)).ok()
+    
+    def check_object_values(obj, name, epsilon, check_num_incr=True, check_num_mult=True):
```
MeirShpilraien, line 1339: I honestly would prefer to have a test which is easier to read even at a cost of code duplication. It is very hard to follow this test and understand what it is testing. I also prefer a more simple and short tests even at the cost of many test functions.

MeirShpilraien (commented):
Few more comments.

src/c_api.rs
MeirShpilraien: It could be done on the user side of the API, I believe API's should be minimal and simple.

src/commands.rs
MeirShpilraien: Ohh nice, lets add a comment here that explains it.

src/include/rejson_api.h
MeirShpilraien: Do we ever return such a value? How come we will have none existing type?

src/ivalue_manager.rs
oshadmi: Although `NUMPOWBY` is deprecated, since it is still here, same behavior as add and mul

src/include/rejson_api.h
oshadmi: We do not return it, but if we would use value `7` for a new type, an older RediSearch module might error.
Maybe it was used as the type for uninitialized value.

src/c_api.rs
oshadmi: For the common use case, checking the type, and asking for the corresponding value is minimal and simple.
For double, it can also do more in case it is requested to (and `ULong` is now aligned with `Long`)
Do you think we should revert this optimization? (could affect RediSearch VecSim)

Issue #879, opened 2022-12-09
"\" not working properly when using regex


buraktanriverdi:
Hi, I tried to generate query with regex while using JSON.GET command. The example in the document works fine, but when I use "\\" I noticed that the regex doesn't work properly.

```
> json.get store '$.inventory.headphones[?(@.date =~ "2+")]'
[{"id":12345,"name":"Noise-cancelling Bluetooth headphones","description":"Wireless Bluetooth headphones with noise-cancelling technology","wireless":true,"connection":"Bluetooth","price":99.98,"stock":25,"free-shipping":false,"colors":["black","silver"],"date":"2014-01-01"}]
> json.get store '$.inventory.headphones[?(@.date =~ "\\d{4}")]'
[]
> json.get store '$.inventory.headphones[?(@.date =~ "\d{4}")]'
Error occurred on position 35, "$.inventory.headphones[?(@.date =~  ---->>>> "\d{4}")]", expected one of the following: decimal, from_current, from_root, boolean_true, boolean_false.

```

PR #884, opened 2022-12-12
fix regex pattern with escape chars

inceabdullah:
in the issue https://github.com/RedisJSON/RedisJSON/issues/879 says that:

regex patterns with  backslash (`\`) cannot be used properly

example snippet is:
```bash
> json.get store '$.inventory.headphones[?(@.date =~ "2+")]'
[{"id":12345,"name":"Noise-cancelling Bluetooth headphones","description":"Wireless Bluetooth headphones with noise-cancelling technology","wireless":true,"connection":"Bluetooth","price":99.98,"stock":25,"free-shipping":false,"colors":["black","silver"],"date":"2014-01-01"}]
> json.get store '$.inventory.headphones[?(@.date =~ "\\d{4}")]'
[]
> json.get store '$.inventory.headphones[?(@.date =~ "\d{4}")]'
Error occurred on position 35, "$.inventory.headphones[?(@.date =~  ---->>>> "\d{4}")]", expected one of the following: decimal, from_current, from_root, boolean_true, boolean_false.
```

As a result of my debug, I found out that, when we use escape chars in pattern it pass through the func `re_match` as `String`, not `Str`.

My debug output:

```
self: Value("2022-12-11"), s: Str("2{1}")

self: Value("2022-12-11"), s: String("\\d{1}")
```

However in `match` operator used `Str` not `String`. I have just added a case for `String`:

```rust
            (TermEvaluationResult::Value(v), TermEvaluationResult::String(regex)) => {
                let regex = &regex
                    .replace("\\\\", "\\").replace("\\\"", "\"")
                    .replace("\\\\", "\\").replace("\\'", "'");
                match v.get_type() {
                    SelectValueType::String => Self::re_is_match(regex, v.as_str()),
                    _ => false
                }
            }
```

and it works! 

Regard.

### DIFF
--- a/src/jsonpath/json_path.rs
+++ b/src/jsonpath/json_path.rs
@@ -452,6 +452,15 @@ impl<'i, 'j, S: SelectValue> TermEvaluationResult<'i, 'j, S> {
                     _ => false,
                 }
             }
+            (TermEvaluationResult::Value(v), TermEvaluationResult::String(regex)) => {
+                let regex = &regex
+                    .replace("\\\\", "\\").replace("\\\"", "\"")
+                    .replace("\\\\", "\\").replace("\\'", "'");
+                match v.get_type() {
+                    SelectValueType::String => Self::re_is_match(regex, v.as_str()),
+                    _ => false
+                }
+            }
             (TermEvaluationResult::Value(v1), TermEvaluationResult::Value(v2)) => {
                 match (v1.get_type(), v2.get_type()) {
                     (SelectValueType::String, SelectValueType::String) => {


oshadmi:
@gkorland @MeirShpilraien I think the grammar should differentiate between a common string term, such as the one in a `$.term` path, and a regex, which can be used only in specific contexts, such as following a filter operator  `=~`.
Otherwise we bring stuff from [regex syntax](https://docs.rs/regex/latest/regex/#syntax) which we may not want into the common string term and vice versa, such as `"\d"` or `"\"+"` which are valid as regex.

However this would not be aligned with the current capability of reading a regex from a JSON String at a specific path on the right-hand side of the `=~`, for example, `\d` in a JSON String would have to be stored as `\\d`. Cleaning up escape chars may change or invalidate a regex, e.g., a regex such as `"[\\]+"` so such pre-processing of a String in order to make it a regex seems problematic (for example a `\\d` in JSON could be ambiguous as literal `\` and then a `d`, or a `\d` for digit that was added with a `\` at the beginning to make it legal JSON)

We could have a straightforward full blown regex on the right-hand side which will not require further escaping and can be passed to the regex builder as-is, and a crippled escaped regex in a JSON String value, that would require pre-processing.
Or we can always require the escaped regex, which would require pre-processing.
I think we should drop the support for getting the regex from a JSON String value at a specific path in order to make this simpler/consistent.
WDYT?

PR #885, opened 2022-12-12
[DOC UPDATE] Update return info in JSON.GET topics 

### LINKED ISSUE #882: RedisJSON Should elaborate on why there are extraneous array brackets around returned values
The page `https://redis.io/docs/stack/json/` shows examples of reading JSON values from the database.

However, it is surprising, and not explained, why there seem to be an extraneous array bracket pair around the returned values.

For instance, the first example is:

```
127.0.0.1:6379> JSON.SET animal $ '"dog"'
"OK"
127.0.0.1:6379> JSON.GET animal $
"[\"dog\"]"
127.0.0.1:6379> JSON.TYPE animal $
1) "string"
```

But, since the value associated with animal is a _string_, it is surprising that the returned value looks like (a string representation of) a JSON array containing the value.

(I am completely new to Redis, but presumably that makes me the target audience for this page).

nermiller:
Fixes #882, explains returns of JSON array and JSON object

### DIFF
--- a/docs/commands/json.get.md
+++ b/docs/commands/json.get.md
@@ -56,7 +56,7 @@ Produce pretty-formatted JSON with `redis-cli` by following this example:
 JSON.GET returns a bulk string representing a JSON array of string replies. 
 Each string is the JSON serialization of each JSON value that matches a path. 
 Using multiple paths, JSON.GET returns a bulk string representing a JSON object with string values. 
-Each string value is the JSON serialization of each JSON value that matches a path.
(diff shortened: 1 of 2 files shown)


nermiller:
@oshadmi @rafie I'm still working on the index page but you can get started with reviews of the completed edits.

oshadmi (commented):
👍🏼 Just some alignment suggestion

oshadmi (approved):
LGTM 👍🏼

PR #890, opened 2022-12-18
MOD-4602: handle jsonpath filter with 3 or more operands

### LINKED ISSUE #889: Filtering with more than two keys
While filtering json array using json path it works fine for less than 3 conditions
For below json 
```
[{"a":1,"b":3,"c":4},{"a":2,"b":2,"c":3},{"a":3,"c":5,"b":1}]
```
While filtering it with one or two conditions it works perfectly fine, but it ignores any more conditions
For example
```json.get test '$[?(@.a>=2 && @.b>=1)]'```
returns 
```[{"a":2,"b":2,"c":3},{"a":3,"c":5,"b":1}]```
but when applied with third condition it ignores the last condition
For example
```json.get test '$[?(@.a>=2 && @.b>=1 && @.c>5)]'```
returns 
```[{"a":2,"b":2,"c":3},{"a":3,"c":5,"b":1}]```

### PR DESCRIPTION
**Pull request by oshadmi**
Fix #889

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,8 @@ crate-type = ["cdylib", "rlib"]
 name = "rejson"
 
 [dev-dependencies]
-env_logger = "0.8"
+env_logger = "0.10.0"
+redis-module = { version="1.0", features = ["experimental-api", "test"]}
 
 [dependencies]
 pest = "2.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,9 @@ categories = ["database-implementations"]
 crate-type = ["cdylib", "rlib"]
 name = "rejson"
 
+[dev-dependencies]
+env_logger = "0.8"
+
 [dependencies]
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: Cargo.toml [hunk 1]
```diff
@@ -16,6 +16,10 @@ categories = ["database-implementations"]
 crate-type = ["cdylib", "rlib"]
 name = "rejson"
 
+[dev-dependencies]
+env_logger = "0.10.0"
+redis-module = { version="1.0", features = ["experimental-api", "test"]}
```
MeirShpilraien, line 21: Lets just make sure this will not somehow effect the final .so file itself, we need to see that the redisjson.so are still using the Redis allocator.

MeirShpilraien (commented):
Looks good,
One comment.

Cargo.toml
oshadmi: I'll comment it out

MeirShpilraien (approved):

PR #897, opened 2022-12-25
simplify grammar

gkorland:


MeirShpilraien:
We need to make sure those changes does not reduce performance.

gkorland:
@MeirShpilraien do you expect any? I think it might improve performance.

src/jsonpath/grammer.pest
```diff
@@ -44,20 +45,22 @@ term = _{select_term | decimal | string | boolean}
...
-all = {"*"}
+all = @{"*"}
 
-full_scan = {".."}
+full_scan = @{".."}
 
-bracket = _{numbers_range | numbers_list | string_list | all | "?" ~ filter}
+square_bracket = _{"[" ~ (numbers_range | numbers_list | string_list | all | "?" ~ filter) ~ "]"}
 
-element = _{"." ~ literal | full_scan ~ (literal | all | "[" ~ bracket ~ "]") | "." ~ all | ("."? ~ "[" ~ bracket ~ "]")}
```
MeirShpilraien, line 60: I believe this was done for performance improvement where we assume most likeliy we will encounter `.literal` and the parser will match it faster.

MeirShpilraien: But not sure so we should check

src/jsonpath/grammer.pest
```diff
@@ -44,20 +45,22 @@ term = _{select_term | decimal | string | boolean}
...
+element = _{(full_scan | ".") ~ (literal | all | square_bracket) | square_bracket}
+
+function = _{literal ~ "(" ~ (literal)? ~ ("," ~ literal)* ~ ")"}
```
MeirShpilraien, line 63: Where do we use this? Or is it a preparation for functions support?

MeirShpilraien (commented):
Small comments about performance.

src/jsonpath/grammer.pest
gkorland: my mistake, I forgot it in the code for future support

PR #904, opened 2022-12-29
Added `rustis` client library

mcatanzariti:
[rustis](https://github.com/dahomey-technologies/rustis)


gkorland:
@mcatanzariti thanks for adding the new client, you might also want to add it on the list in the README

gkorland (approved):

PR #912, opened 2023-01-11
MOD-4787: Calculate the IValue actual memory usage

gkorland:
ref https://github.com/Diggsey/ijson/issues/23

### DIFF
--- a/src/ivalue_manager.rs
+++ b/src/ivalue_manager.rs
@@ -630,23 +630,56 @@ impl<'a> Manager for RedisIValueJsonKeyManager<'a> {
         }
     }
 
+    ///
+    /// following https://github.com/Diggsey/ijson/issues/23#issuecomment-1377270111
+    ///
     fn get_memory(&self, v: &Self::V) -> Result<usize, RedisError> {
         let res = size_of::<IValue>()
             + match v.type_() {
-                ValueType::Null | ValueType::Bool | ValueType::Number => 0,
+                ValueType::Null | ValueType::Bool => 0,
+                ValueType::Number => {
+                    let num = v.as_number().unwrap();
+                    if num.has_decimal_point() {
+                        // 64bit float
+                        16
+                    } else if num > &INumber::from(-128) && num <= &INumber::from(383) {
+                        // 8bit
+                        0
+                    } else if num > &INumber::from(-8_388_608) && num <= &INumber::from(8_388_607) {
+                        // 24bit
+                        4
+                    } else {
+                        // 64bit
+                        16
+                    }
+                }
                 ValueType::String => v.as_string().unwrap().len(),
-                ValueType::Array => v
-                    .as_array()
-                    .unwrap()
-                    .into_iter()
-                    .map(|v| self.get_memory(v).unwrap())
-                    .sum(),
-                ValueType::Object => v
-                    .as_object()
-                    .unwrap()
-                    .into_iter()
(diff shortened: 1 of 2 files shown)


filipecosta90:
@gkorland when this PR is ready for review please label it with `action:run-benchmark` and check the graphs for dataset memory usage. We should be able to verify that there's a difference on perf charts.

src/ivalue_manager.rs
```diff
@@ -637,23 +637,56 @@ impl<'a> Manager for RedisIValueJsonKeyManager<'a> {
...
+                        0
+                    } else if num > &INumber::from(-8_388_608) && num <= &INumber::from(8_388_607) {
+                        // 24bit
+                        4
+                    } else {
+                        // 64bit
+                        16
+                    }
+                }
                 ValueType::String => v.as_string().unwrap().len(),
```
Diggsey, line 663: Be aware that this may signficantly over-estimate the size, since strings are interned.

src/ivalue_manager.rs
```diff
@@ -637,23 +637,56 @@ impl<'a> Manager for RedisIValueJsonKeyManager<'a> {
...
+                ValueType::Object => {
+                    let val = v.as_object().unwrap();
+                    let capacity = val.capacity();
+                    if capacity == 0 {
+                        0
+                    } else {
+                        size_of::<usize>() * (capacity * 3 + 2)
+                            + val
+                                .into_iter()
+                                .map(|(s, v)| s.len() + self.get_memory(v).unwrap())
```
Diggsey, line 686: Be aware that this may signficantly over-estimate the size, since strings are interned.

src/ivalue_manager.rs
```diff
@@ -637,23 +637,56 @@ impl<'a> Manager for RedisIValueJsonKeyManager<'a> {
...
         let res = size_of::<IValue>()
             + match v.type_() {
-                ValueType::Null | ValueType::Bool | ValueType::Number => 0,
+                ValueType::Null | ValueType::Bool => 0,
+                ValueType::Number => {
+                    let num = v.as_number().unwrap();
+                    if num.has_decimal_point() {
+                        // 64bit float
+                        16
+                    } else if num > &INumber::from(-128) && num <= &INumber::from(383) {
```
oshadmi, line 652: The comment in mentioned ijson issue 23 says
> Else if in the range -128..=383 (none)

Isn't it inclusive? Should be `else if num >= &INumber::from(-128) && ...` ?

src/ivalue_manager.rs
```diff
@@ -637,23 +637,56 @@ impl<'a> Manager for RedisIValueJsonKeyManager<'a> {
...
+                        // 8bit
```
oshadmi, line 653: The comment is a bit unclear - this case is for storing `i8` or `u8` values,
But this range uses more than 8 bits? (-128 til 383 has 512 values)

src/ivalue_manager.rs
gkorland: yes, I'm aware but it still give a single JSON it's potential cost

src/ivalue_manager.rs
gkorland: @Diggsey ?

src/ivalue_manager.rs
gkorland: @Diggsey

oshadmi (approved):

Issue #415, opened 2021-07-03
Add support for JSON.SET + custom ttl (JSON.SETEX)


stavalfi:
I would like to do it in one command instead of two.

gkorland:
@stavalfi RedisJSON doesn't support such command but you can use Redis [Multi](https://redis.io/commands/multi)

stavalfi:
mset has much better performance compared to multi

gkorland:
The new command should be similar to https://redis.io/commands/setex/

```
JSON.SET key seconds path value [ NX | XX]
```

gkorland:
@stavalfi why did you close the issue? it was not implemented yet, I just suggested the new API

smcardle:
From my perspective I would like to see both JSON.SETEX and JSON.GETEX implmented

We mostly use Redis as a cache for JSON Objects with various cache policies, one being where we set the entry EXPIRY when setting the object into the cache and then reset the cache entry EXPIRY time each time the entry is accessed. If the entry hasn't been accessed in X seconds the entry is evicted from the cache.

This policy allows an entry to stay in the cache while a user is actively using the system.

smcardle:
Any idea if and when Redis intends to address this ?

It would be great for my use cases to have a single ATOMIC call for both get and set JSON sara

PR #913, opened 2023-01-15
Add EX and PX options to JSON.SET

gkorland:
Ref #415

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -14,7 +14,9 @@ use crate::redisjson::{normalize_arr_indices, Format, Path};
 use redis_module::{Context, RedisValue};
 use redis_module::{NextArg, RedisError, RedisResult, RedisString, REDIS_OK};
 use std::cmp::Ordering;
+use std::num::NonZeroU64;
 use std::str::FromStr;
+use std::time::Duration;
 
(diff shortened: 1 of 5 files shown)


src/commands.rs
```diff
@@ -662,7 +681,14 @@ pub fn json_set<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>) -
                 ))
             }
         }
+    };
+    // if json.set was successful and we have an expire we set it
+    if result.is_ok() {
+        if let Some(ex) = expire {
```
oshadmi, line 687: We can also support `-1` to allow to `PERSIST` a key which may have a TTL?

gkorland: I'm following Redis semantic, if a user wants  without TTL he should just not send EX or PX

bjwschaap:
Any indications on as when this will be merged?

PaulAtkins88:
bump on this PR

timeimp:
I was expected a 1:1 match with the SET command for `JSON.SET` - imagine my confusion when the `EX` value was never accepted!

Is there an outstanding item preventing merge? Would really like to see this normal `EX` value be set on `JSON.SET` commands

PR #914, opened 2023-01-16
Ignore case on spell check

chayim:


gkorland (dismissed):

gkorland:
@chayim please notice spelling is still failing on `jedis`

chayim:
I don't think this actually works - even though the setting says it does.

gkorland (approved):

Issue #150, opened 2020-04-20
Add support for JSON Patch


gkorland:
https://docs.rs/json-patch/0.2.6/json_patch/

PR #916, opened 2023-01-18
Add JSON.MERGE

gkorland:


src/serde_value_manager.rs
```diff
@@ -222,14 +222,30 @@ impl<'a> WriteHolder<Value, Value> for KeyHolderWrite<'a> {
...
-            update(&path, self.get_value().unwrap().unwrap(), |_v| {
+            replace(&path, self.get_value()?.unwrap(), |_v| {
                 updated = true;
                 Ok(Some(v.take()))
             })?;
         }
         Ok(updated)
     }
 
+    fn merge_value(&mut self, path: Vec<String>, v: Value) -> Result<bool, RedisError> {
```
oshadmi, line 233: Not using `json-patch` crate (Ref #150) since it only works with serde_json `Value` and not wiith `IValue`?

tests/pytest/test.py
```diff
@@ -1279,6 +1279,15 @@ def testFilterPrecedence(env):
     r.assertEqual(json.loads(res), [doc[0]])
     r.expect('JSON.GET', 'doc', '$[?(@.f==true || @.one==1 && @.t==false)]').equal('[]')
 
+def testMerge(env):
```
oshadmi, line 1282: Need test for key deletion? (merge with `null`)

src/ivalue_manager.rs
```diff
@@ -569,6 +585,25 @@ impl ReadHolder<IValue> for IValueKeyHolderRead {
     }
 }
 
+fn merge(doc: &mut IValue, patch: &IValue) {
+    if !patch.is_object() {
+        *doc = patch.clone();
+        return;
+    }
+
+    if !doc.is_object() {
```
oshadmi, line 594: If doc is not an object, it will be completely overwritten (replaced) by the patch value? (merge takes all from the right-hand-side patch object, and nothing from the left-hand-side array or scalar)

oshadmi (commented):
Few comments

src/serde_value_manager.rs
gkorland: right, this is why I had to implement merge myself

src/commands.rs
```diff
@@ -665,6 +665,85 @@ pub fn json_set<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>) -
...
+    let mut redis_key = manager.open_key_write(ctx, key)?;
+    let current = redis_key.get_value()?;
+
+    let val = manager.from_str(value, format)?;
+
+    match current {
+        Some(doc) => {
+            if path.get_path() == JSON_ROOT_PATH {
+                redis_key.merge_value(Vec::new(), val)?;
+                redis_key.apply_changes(ctx, "json.merge")?;
```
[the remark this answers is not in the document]
gkorland: https://github.com/RediSearch/RediSearch/pull/3342

src/ivalue_manager.rs
gkorland: yes, I didn't do the arrays support yet

joshuakoh1:
@gkorland Can we also add this "merge" functionality to numincrby where it treats the original value as 0 if the key does not exist?

src/ivalue_manager.rs
gkorland: I think this is what we agreed on merge arrays, right? @adrianoamaral ?

gkorland:
> functionality

I think it worth a different feature 
@adrianoamaral

src/ivalue_manager.rs
adrianoamaral: yes, following the RFC7396 - array, replaced by the new value. Other behaviours we can handle on the JSON.ARRAPPEND

adrianoamaral:
Already on the plans for enhance JSON.NUMINCRBY @joshuakoh1, thank you for asking.

tests/pytest/test.py
oshadmi, line 1: Need to also test key notifications? As with [`JSON.MSET`](https://github.com/RedisJSON/RedisJSON/blob/8202665ae20e00a93be9aa6ecf6021cdd0c5bf5a/tests/pytest/test_keyspace_notifications.py#LL33C56-L33C56)

gkorland: done

oshadmi (approved):

PR #940, opened 2023-03-02
use expected instead of panic

gkorland:


src/jsonpath/json_path.rs
```diff
@@ -45,22 +45,23 @@ impl<'i> Query<'i> {
...
-                    let first_on_list = last.into_inner().next();
-                    first_on_list.map(|first| (first.as_str().to_string(), JsonPathToken::String))
-                }
-                _ => panic!("pop last was used in a none static path"),
-            },
-            None => None,
+        let last = self
+            .root
+            .next_back()
+            .expect("pop last was used in a none static path");
```
oshadmi, line 51: When `None` it was returning `None` and not panicking?
(before there was 1 `panic` and now there are `expect` and `panic`)

gkorland: @oshadmi good catch not sure why I did it

src/formatter.rs
```diff
@@ -175,6 +175,7 @@ mod tests {
     use super::*;
 
     #[test]
+    #[allow(clippy::cognitive_complexity)]
```
oshadmi, line 178: We are linting cognitive_complexity in tests?

oshadmi (approved):
LGTM!

src/formatter.rs
gkorland: clippy does

Issue #821, opened 2022-09-07
Review the code in `fn json_set` when updating a doc - can `apply_changes` be missed?


oshadmi:
Need to review the code that is calling `apply_changes` in `json_set`: Can it be missed when `res` is set to `true`, but later on in the `for` loop, `res` may be overwritten with a `false`?
See https://github.com/RedisJSON/RedisJSON/blob/35cda93827b40da29b3b927f2de8d80fd9269d1a/src/commands.rs#L633

Could this result with an actual doc being updated, but no `notify_keyspace_event` is notified?

Should probably be
```rust
res = res
    || match ui { ...
```

PR #944, opened 2023-03-21
add json.mset

gkorland:


src/commands.rs
```diff
@@ -578,6 +578,10 @@ pub fn json_get<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>) -
 pub fn json_set<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>) -> RedisResult {
     let mut args = args.into_iter().skip(1);
 
+    if args.len() < 3 {
+        return Err(RedisError::WrongArity);
+    }
```
MeirShpilraien, line 583: Why is this needed? Wouldn't the `?` bellow will cover it?

gkorland: you're right, I thought it will be more standard this way

src/commands.rs
```diff
@@ -662,6 +666,81 @@ pub fn json_set<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>) -
...
+        let value = manager.from_str(value_str, Format::JSON, true)?;
+
+        actions.push((redis_key, update_info, value));
+    }
+
+    actions
+        .drain(..)
+        .for_each(|(mut redis_key, update_info, value)| {
+            if let Some(mut update_info) = update_info {
+                if !update_info.is_empty() {
```
MeirShpilraien, line 710: Maybe we can extract this to a function somehow? Cause its the same as json.set no?

src/commands.rs
```diff
@@ -662,6 +666,81 @@ pub fn json_set<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>) -
...
+    while let Ok(key) = args.next_arg() {
+        let mut redis_key = manager.open_key_write(ctx, key)?;
+
+        // Verify the key is a JSON type
+        let key_value = redis_key.get_value()?;
+
+        // Verify the path is valid and get all the update info
+        let path = Path::new(args.next_str()?);
+        let update_info = if path.get_path() == JSON_ROOT_PATH {
+            None
```
MeirShpilraien, line 690: Why on the root we do not have to do any updates?

src/commands.rs
oshadmi: Can do preliminary check on the arg count that it is a multiplication of 3

src/commands.rs
```diff
@@ -662,6 +666,81 @@ pub fn json_set<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>) -
...
+        } else if let Some(value) = key_value {
+            Some(KeyValue::new(value).find_paths(path.get_path(), &SetOptions::None)?)
```
oshadmi, line 692: * Since we find the paths on the first pass - we may fail updating the same key multiple times,

Which is possible in multi exec
e.g.,
This is OK
```
json.set a $ '{"x":[10,20]}'
json.set a '$.x[1]' '30'

json.get a $
"[{\"x\":[10,30]}]"
```

But this will fail, since before the `json.mset` command, `$.x[1]` does not exist:
```
json.mset a $ '{"x":[10,20]}' a '$.x[1]' '30'
```

* Since we modify the key on the second path, we may fail although the first pass does not detect it,
e.g., 
```
json.set a $ '{"x": {"y":[10,20], "z":[30,40]}}'

json.mset a $.x '{}' a '$.x.z[1]' '50'
```

The other alternative would be to do modifications in one pass, and if any fails, rollback, but that would be less efficient.
For multiple updates on the same key we can add a limitation that updates are incremental (and that path validations are not incremental?)

src/commands.rs
```diff
@@ -662,6 +666,81 @@ pub fn json_set<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>) -
...
+                                    redis_key.set_value(sui.path, value.clone()).unwrap()
+                                }
+                                UpdateInfo::AUI(aui) => redis_key
+                                    .dict_add(aui.path, &aui.key, value.clone())
+                                    .unwrap(),
+                            }
+                        }
+                    }
+                    if res {
+                        redis_key.apply_changes(ctx, "json.mset").unwrap();
```
oshadmi, line 734: Need to handle notification in RediSearch, e.g., RediSearch/RediSearch/pull/3342

src/commands.rs
```diff
@@ -662,6 +666,81 @@ pub fn json_set<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>) -
...
+                    }
+                }
+            } else {
+                redis_key.set_value(Vec::new(), value).unwrap();
```
oshadmi, line 738: No need to apply changes also here?

src/commands.rs
```diff
@@ -662,6 +666,81 @@ pub fn json_set<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>) -
...
+                    if update_info.len() == 1 {
+                        res = match update_info.pop().unwrap() {
+                            UpdateInfo::SUI(sui) => redis_key.set_value(sui.path, value).unwrap(),
+                            UpdateInfo::AUI(aui) => {
+                                redis_key.dict_add(aui.path, &aui.key, value).unwrap()
+                            }
+                        }
+                    } else {
+                        for ui in update_info {
+                            res = match ui {
```
oshadmi, line 723: Related to https://github.com/RedisJSON/RedisJSON/issues/821#issue-1364660409
Maybe should be something like `res = res || match ui {` ?

src/commands.rs
gkorland: it does, see bellow:
```
                redis_key.set_value(Vec::new(), value).unwrap();
```

src/commands.rs
gkorland: Don't we have the issue in JSON.SET?

src/commands.rs
gkorland: maybe we can, it's not exactly the same, since in JSON.SET we also have the NX/EX flags

gkorland: done

src/commands.rs
gkorland: @oshadmi see my last commit

src/commands.rs
```diff
@@ -662,6 +645,88 @@ pub fn json_set<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>) -
...
+                redis_key.set_value(Vec::new(), value).unwrap()
+            };
+            if updated {
+                redis_key.apply_changes(ctx, "json.mset").unwrap();
+            }
+        });
+    REDIS_OK
+}
+
+fn apply_updates<M: Manager>(
```
MeirShpilraien, line 701: Just to make sure, you just extracted the code above into a function right?

src/commands.rs
```diff
@@ -662,6 +645,88 @@ pub fn json_set<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>) -
...
+    }
+
+    actions
+        .drain(..)
+        .for_each(|(mut redis_key, update_info, value)| {
+            let updated = if let Some(update_info) = update_info {
+                !update_info.is_empty() && apply_updates::<M>(&mut redis_key, value, update_info)
+            } else {
+                // In case it is a root path
+                redis_key.set_value(Vec::new(), value).unwrap()
```
MeirShpilraien, line 692:
```suggestion
                redis_key.set_value(Vec::new(), value).expect("<Some informative message>")
```

src/commands.rs
gkorland: one change, the `for ui in update_info` now doesn't stop on the first error

oshadmi: @gkorland
> the for ui in update_info now doesn't stop on the first error

Consider testing that a specific change, which cannot be done, does not prevent `apply_changes` for previously done changes, e.g., once first change is done, second cannot be done, but third should be done as well.
```
json.set a $ '{"x": {"y":[10,20], "z":[30,40]}}'
json.set b $ '{"x": 60}'

json.mset a $.x '{}' a '$.x.z[1]' '50' b $.x '70'
```

src/commands.rs
gkorland: I replaced it with a fold so if one fail an error will be returned

src/commands.rs
oshadmi: Maybe also on the same key
```
json.mset a $.x '{}' a '$.x.z[1]' '50' a $.u '70'
```

oshadmi: Since it is also fixed for `json_set` - now it may notify on a similar scenario with multi-value (using wildcards, etc.) , while it did not notify before - could affect RediSearch results for example, but this would be a progression and is not considered a breaking change.
Agree?

oshadmi: An example for the problem with a single `json.set` (using RediSearch as a subscriber for keyspace notifications)
```
127.0.0.1:6379> json.set k1 $ '{"x":{"x":{"y":[4,5,6]}}}'
OK
127.0.0.1:6379> ft.create idx on json schema '$..y[*]' as y numeric
OK
127.0.0.1:6379> ft.search idx '@y:[4 5]'
1) (integer) 1
2) "k1"
3) 1) "$"
   2) "{\"x\":{\"x\":{\"y\":[4,5,6]}}}"
```
The following `json.set` was failing in the middle and not notifying on changes, so the index was not being updated causing search to return a stale result:
```
127.0.0.1:6379> json.set k1 '$..x' '1'
(nil)
127.0.0.1:6379> json.get k1 $
"[{\"x\":1}]"
127.0.0.1:6379> ft.search idx '@y:[4 5]'
1) (integer) 1
2) "k1"
3) 1) "$"
   2) "{\"x\":1}"
```
With this PR it should be fixed ☝🏼 (search should return an empty result)
And once RediSearch/RediSearch#3507 is merged, it should also work with `json.mset`, e.g., 
```
json.mset k1 $ '{"x":{"x":{"y":[4,5,6]}}}' k1 '$..x' '1'
```
And then `ft.search idx '@y:[4 5]'` should return an empty result
And with
```
json.mset k1 $ '{"x":{"x":{"y":[4,5,6]}}}' k1 '$..x' '1' k1 '$.y' '[4, 5, 6]'
```
The same search should return k1.

<br>
I suppose such tests should be added to RediSearch/RediSearch#3507

oshadmi: @gkorland Keyspace notification can also be tested in this PR (without RediSearch) as in https://github.com/RedisJSON/RedisJSON/blob/0b898bef64912f7d5837c2a2495969c92f2c2c77/tests/pytest/test_keyspace_notifications.py#L14

oshadmi (approved):

PR #1004, opened 2023-05-15
cleanup and deps

gkorland:


json_path/Cargo.toml
```diff
@@ -1,16 +1,16 @@
 [package]
 name = "json_path"
 version = "0.1.0"
-edition = "2021"
+edition.workspace = true
```
oshadmi, line 4: Don't we also need to add `license` here as well?
(and maybe also description, repository, keywords, etc.)

oshadmi (dismissed):

PR #1018, opened 2023-06-04
remove serde_json manager

gkorland:


oshadmi:
@gkorland It seems we can use `ijson::IValue` in more places. There are more such opportunities.
Why not even in the sub crate `json_path`?

gkorland:
> `json_path`

gkorland:
@oshadmi we're using serde to parse the JSON, and for short living Values that we don't store I think serde_json is more efficient CPU wise.

PR #1046, opened 2023-07-10
Adapt command to RESP3

gkorland:
1. JSON.ARRPOP
2. JSON.MGET
3. Adapt commands with optional `path` to us `$` as the default path


MeirShpilraien:
@gkorland can you please update the top comment with the details about what was changed in this PR?

redis_json/src/commands.rs
```diff
@@ -608,16 +629,27 @@ pub fn json_mget<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>)
...
-                                        to_string(doc)
+                            value.map_or(Ok(RedisValue::Null), |doc| {
+                                let key_value = KeyValue::new(doc);
+
+                                // let value = match key.get_value()? {
+                                //     Some(doc) => KeyValue::new(doc).to_json(&mut paths, &format_options)?,
+                                //     None => RedisValue::Null,
+                                // };
+
+                                // value_to_resp3
```
MeirShpilraien, line 640: Remove?

redis_json/src/commands.rs
```diff
@@ -588,17 +594,32 @@ pub fn json_mget<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>)
...
 
-        let format_options = FormatOptions {
-            resp3: is_resp3(ctx),
-            ..Default::default()
+        let is_resp3 = is_resp3(ctx);
+        let mut format_options = FormatOptions::new(is_resp3);
+
+        // Check if the optional argument [FORMAT STRING|JSON|EXPAND] is present
+        let keys = if is_resp3 {
+            let format = &args[args.len() - 3..args.len() - 1];
```
MeirShpilraien, line 603: Wouldn't this raise a panic if out of range?

redis_json/src/commands.rs
```diff
@@ -588,17 +594,32 @@ pub fn json_mget<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>)
...
+            if let Ok(s) = format[0].try_as_str() {
+                if s.eq_ignore_ascii_case("FORMAT") {
+                    format_options.format = Format::from_str(format[1].try_as_str()?)?;
```
MeirShpilraien, line 606: Same comment as above, where do you check the `format[1]` exists?

redis_json/src/formatter.rs
```diff
@@ -46,27 +46,38 @@ pub struct FormatOptions<'a> {
...
+        self.indent.is_none() && self.space.is_none() && self.newline.is_none()
     }
 }
 
 impl Default for FormatOptions<'_> {
+    /// Creates a new FormatOptions with the default values matching RESP2
     fn default() -> Self {
         Self {
-            format: Format::EXPAND,
+            format: Format::STRING,
```
MeirShpilraien, line 80: You change the default here? Is this interntional?

MeirShpilraien (commented):
Some comments, probably the most important one are the JSON.MGET command format.

redis_json/src/commands.rs
gkorland: it will never fall since we check that the command has more than 3 arguments

redis_json/src/commands.rs
gkorland: yes it has to exist since the command has more than 3 arguments

redis_json/src/formatter.rs
gkorland: yes, String is the default for RESP2

redis_json/src/commands.rs
```diff
@@ -579,7 +585,7 @@ pub fn json_del<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>) -
 }
 
 ///
-/// JSON.MGET <key> [key ...] <path>
+/// JSON.MGET <key> [key ...] [FORMAT STRING|JSON|EXPAND] <path>
```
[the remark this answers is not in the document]
gkorland: >This API could be very confusing, for example what if the keys I want to work on are FORMAT STRING, this will obligate 
> me to specify the FORMAT as mandatory.
Yes @LiorKogan 's decision....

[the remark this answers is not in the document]
MeirShpilraien: :face_with_head_bandage:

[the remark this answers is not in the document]
MeirShpilraien: And what about the command definition which is wrong?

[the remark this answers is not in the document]
gkorland: > ["json.mget", json_command!(json_mget), "readonly", 1,1,1],

Yes, I'm not trying to protect from cross on the Redis level.
This means that you must catch cross slot errors in the command implementation itself.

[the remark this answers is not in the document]
oshadmi: @gkorland It was decided for `JSON.MGET` to
* Avoid adding the `FORMAT` arg
* Use default `EXPAND` in resp3 and default `STRING` in resp2
* Not to fix the command key position (from 1,1,1 to 1,-2,1) to avoid breaking change in resp2
  * @LiorKogan will document this existing problem
* Need a new command that follows the redis supported key positions

redis_json/src/commands.rs
```diff
@@ -1361,15 +1367,42 @@ pub fn json_arr_len<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString
 }
 
 ///
-/// JSON.ARRPOP <key> [path [index]]
+/// JSON.ARRPOP <key>
+///         [FORMAT {STRING|JSON|EXPAND}]
```
MeirShpilraien, line 1371: So here also the path can not be FORMAT .. very ugly api. I would remove it and match the behaviour with `json_mget`..

MeirShpilraien (dismissed):
With respect to code it look good.
I do not like this optional format argument on `JSON.ARRPOP`...

PR #1047, opened 2023-07-12
MOD-5390: Add C API to iterate a JSON Object

oshadmi:
Test will be in RediSearch/RediSearch#3707 (until MOD 4159 is done)

### DIFF
--- a/redis_json/src/c_api.rs
+++ b/redis_json/src/c_api.rs
@@ -301,7 +301,9 @@ where
 {
     let iter = unsafe { &mut *(iter.cast::<KeyValuesIterator<M::V>>()) };
     if iter.pos < iter.results.len() {
-        create_rmstring(ctx, iter.results[iter.pos].0, str);
+        if !str.is_null() {
+            create_rmstring(ctx, iter.results[iter.pos].0, str);
+        }
         let res = (iter.results[iter.pos].1 as *const M::V).cast::<c_void>();
         iter.pos += 1;
         res
@@ -561,6 +563,8 @@ macro_rules! redis_json_module_export_shared_api {
             )
         }
 
+        static REDISJSON_GETAPI_V1: &str = concat!("RedisJSON_V1", "\0");
+        static REDISJSON_GETAPI_V2: &str = concat!("RedisJSON_V2", "\0");
         static REDISJSON_GETAPI_V3: &str = concat!("RedisJSON_V3", "\0");
         static REDISJSON_GETAPI_V4: &str = concat!("RedisJSON_V4", "\0");
 
@@ -569,6 +573,18 @@ macro_rules! redis_json_module_export_shared_api {
                 LLAPI_CTX = Some(rawmod::RedisModule_GetThreadSafeContext.unwrap()(
                     std::ptr::null_mut(),
                 ));
+                ctx.export_shared_api(
+                    (&JSONAPI_CURRENT as *const RedisJSONAPI_CURRENT).cast::<c_void>(),
+                    REDISJSON_GETAPI_V1.as_ptr().cast::<c_char>(),
+                );
+                ctx.log_notice("Exported RedisJSON_V1 API");
+
+                ctx.export_shared_api(
+                    (&JSONAPI_CURRENT as *const RedisJSONAPI_CURRENT).cast::<c_void>(),
+                    REDISJSON_GETAPI_V2.as_ptr().cast::<c_char>(),
+                );
+                ctx.log_notice("Exported RedisJSON_V2 API");
+
                 ctx.export_shared_api(
                     (&JSONAPI_CURRENT as *const RedisJSONAPI_CURRENT).cast::<c_void>(),
                     REDISJSON_GETAPI_V3.as_ptr().cast::<c_char>(),
--- a/redis_json/src/c_api.rs
+++ b/redis_json/src/c_api.rs
@@ -293,7 +293,6 @@ where
 pub fn json_api_next_key_value<'a, M: Manager>(
     _: M,
     iter: *mut c_void,
-    ctx: *mut rawmod::RedisModuleCtx,
     str: *mut *mut rawmod::RedisModuleString,
 ) -> *const c_void
 where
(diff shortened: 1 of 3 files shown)


redis_json/src/c_api.rs
```diff
@@ -45,6 +45,11 @@ struct ResultsIterator<'a, V: SelectValue> {
     pos: usize,
 }
 
+struct KeyValuesIterator<'a, V: SelectValue> {
+    results: Vec<(&'a str, &'a V)>,
```
MeirShpilraien, line 49: Can we keep the iterator over the iterms here instead of collecting it into a vector?

redis_json/src/c_api.rs
```diff
@@ -267,6 +272,54 @@ pub fn json_api_is_json<M: Manager>(m: M, key: *mut rawmod::RedisModuleKey) -> c
...
 }
 
+pub fn json_api_get_key_value<'a, M: Manager>(_: M, val: *const c_void) -> *const c_void
+where
+    M::V: 'a,
+{
+    let json = unsafe { &*(val.cast::<M::V>()) };
+    match json.get_type() {
+        SelectValueType::Object => {
+            let vec = json.items().unwrap().collect::<Vec<(&'a str, &'a M::V)>>();
```
MeirShpilraien, line 282: Same comment as above

redis_json/src/c_api.rs
```diff
@@ -267,6 +272,54 @@ pub fn json_api_is_json<M: Manager>(m: M, key: *mut rawmod::RedisModuleKey) -> c
...
+    _: M,
+    iter: *mut c_void,
+    ctx: *mut rawmod::RedisModuleCtx,
+    str: *mut *mut rawmod::RedisModuleString,
+) -> *const c_void
+where
+    M::V: 'a,
+{
+    let iter = unsafe { &mut *(iter.cast::<KeyValuesIterator<M::V>>()) };
+    if iter.pos < iter.results.len() {
```
MeirShpilraien, line 303: I would prefer:
```rust
if iter.pos >= iter.results.len() {
    return null();
}
```

redis_json/src/c_api.rs
```diff
@@ -267,6 +272,54 @@ pub fn json_api_is_json<M: Manager>(m: M, key: *mut rawmod::RedisModuleKey) -> c
...
+        create_rmstring(ctx, iter.results[iter.pos].0, str);
```
MeirShpilraien, line 304: I believe you can create a String with NULL context and so we can avoid the ctx parameter to the function. Also I believe we should check if `str` is not NULL so we will allow the user pass NULL  pointer as `str` indicating that he do not need the key.

redis_json/src/c_api.rs
```diff
@@ -491,21 +571,15 @@ macro_rules! redis_json_module_export_shared_api {
                 ));
                 ctx.export_shared_api(
                     (&JSONAPI_CURRENT as *const RedisJSONAPI_CURRENT).cast::<c_void>(),
-                    REDISJSON_GETAPI_V1.as_ptr().cast::<c_char>(),
-                );
-                ctx.log_notice("Exported RedisJSON_V1 API");
```
MeirShpilraien, line 496: Why removing v1 and v2?

MeirShpilraien (commented):
Some comments

redis_json/src/c_api.rs
oshadmi: I tried and it crashed and I wanted to look at it later due to other priorities
https://github.com/RedisJSON/RedisJSON/pull/1047/commits/61570c41f13fc7c330103c74a9fd8a1f678a1e05

redis_json/src/c_api.rs
oshadmi: Even if this if will usually be not-taken?

redis_json/src/c_api.rs
oshadmi: Regarding using NULL context - we already pass context for creating strings to other APIs,
So it would be inconsistent.
Agree on allowing `null` str when key name is not needed.

redis_json/src/c_api.rs
oshadmi: Will revert.
Should be in a separate PR if/when we bundle JSON and Search together and can drop support for various version combinations.

redis_json/src/c_api.rs
MeirShpilraien: So let understand why it crashes

redis_json/src/c_api.rs
MeirShpilraien: Why is it different? In both cases you are doing an if statement.

redis_json/src/c_api.rs
MeirShpilraien: When/if this will happened, we can also remove the unneeded ctx you mentioned here: https://github.com/RedisJSON/RedisJSON/pull/1047#discussion_r1260775975

redis_json/src/c_api.rs
oshadmi: But most times the if will not be taken? Shouldn't we prefer when we have some info, to put the more-taken branch in the `if`, and the less-taken branch in the `else`?

redis_json/src/c_api.rs
oshadmi: So you prefer new APIs to avoid context? (and keep it for older APIs only which we cannot change)

MeirShpilraien: I believe so

redis_json/src/c_api.rs
MeirShpilraien: Why do we prefer this? Are you afraid of performance because of branch prediction? Maybe we can use [unlikely](https://docs.rs/likely_stable/latest/likely_stable/fn.unlikely.html)?

oshadmi: Avoiding the else

redis_json/src/c_api.rs
oshadmi: Was missing a `Box::new` in `json_api_get_key_value`

redis_json/src/c_api.rs
oshadmi: Was missing a `Box::new` in `json_api_get_key_value`

gkorland (approved):

PR #1095, opened 2023-09-21
[MOD-5769] Open RedisJSON key with flags

DvirDukhan:
This PR adds LLAPI functionality for opening a RedisJSON key with flags
this is mainly to be used in the RediSearch `getKeyCommonJSON`  where `NO_EFFECTS` flag should be send, to abvoid key expiration (write path) while quering (read path).


DvirDukhan:
@oshadmi do we need a new LLAPI version here?

alonre24 (approved):

redis_json/Cargo.toml
```diff
@@ -25,9 +25,11 @@ ijson.workspace = true
 serde_json.workspace = true
 serde.workspace = true
 libc = "0.2"
-redis-module = "2.0"
+redis-module ={ version = "^2.0.7", default-features = false, features = ["min-redis-compatibility-version-7-2"] }
```
MeirShpilraien, line 28: No need to work on version less than 7.2?

redis_json/src/c_api.rs
```diff
@@ -543,39 +579,24 @@ macro_rules! redis_json_module_export_shared_api {
             )
         }
 
-        static REDISJSON_GETAPI_V1: &str = concat!("RedisJSON_V1", "\0");
-        static REDISJSON_GETAPI_V2: &str = concat!("RedisJSON_V2", "\0");
-        static REDISJSON_GETAPI_V3: &str = concat!("RedisJSON_V3", "\0");
-        static REDISJSON_GETAPI_V4: &str = concat!("RedisJSON_V4", "\0");
+        // The apiname argument of export_shared_api should be a string literal with static lifetime
+        static mut VEC_EXPORT_SHARED_API_NAME : Vec<CString> = Vec::new();
```
MeirShpilraien, line 583: We can make it a static slice:
static VEC_EXPORT_SHARED_API_NAME: &'static str = &["RedisJSON_V1", ...];

MeirShpilraien: Also why does it need to be static lifetime?

MeirShpilraien (commented):
Few more comments.

redis_json/src/c_api.rs
oshadmi: https://github.com/redis/redis/blob/unstable/src/module.c#L10498-L10500

> IMPORTANT: the apiname argument should be a string literal with static
> lifetime. The API relies on the fact that it will always be valid in
> the future.
>  int RM_ExportSharedAPI(RedisModuleCtx *ctx, const char *apiname, void *func) {

redis_json/Cargo.toml
oshadmi: I think not.
We need to update ramp to "7.2"
> min_redis_version: "7.1"

Maybe also add a min version check on module load for OSS?

@adrianoamaral

adrianoamaral: we should consider attainment with the min grouped feature set. So, should follow 7, in this case, and be back-ported also for the preview version, but it won't be JSON 2.6, but 2.4

PR #1118, opened 2023-10-23
Fix zip filename for macos

alonre24:


oshadmi (approved):
(If not x86 and not ventura and not monterey - will not be monterey)

PR #1132, opened 2023-11-09
upload snapshots for integration branches as well [MOD-6102]

alonre24:
This temporary fix will allow testing RediSearch against the proper RedisJSON artifacts. 
RediSearch OSS expects to find RedisJSON up to date module in `rejson-oss/snapshots/` path, where the name of the zip file is `rejson-oss.<os>-<platform>-<arch>.<branch_name>.zip`. 
This had been broken due to incompatibility between the file name that was packed (arm arch name was changed to `aarch64` and macos for x86 was changed to catalina) and the expected filename to upload - hence the fix in the `upload-artifacts` script.

In addition, the snapshots were uploaded into a different path ("staging") at some point - and hence the change in`upload-artifacts` step in cirecle-ci.

Also, I added macOS builds to run on the integration branch as well, so we will have the updated artifacts for the master and 2.4 branches after we merge this PR (this can be reverted later on).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -256,6 +256,12 @@ commands:
       - save-tests-logs
       - early-return-for-forked-pull-requests
       - run:
+          name: Upload artifacts to S3
+          command: |
+            if [[ -n $CIRCLE_BRANCH ]]; then
+                make upload-artifacts OSNICK=<<parameters.platform>> SHOW=1
+            fi
+      - run:
           name: Publish container
           command: |
             docker login -u redisfab -p $DOCKER_REDISFAB_PWD
@@ -325,6 +331,7 @@ jobs:
       - early-returns
       - build-steps
       - test-steps
+      - persist-artifacts
       - run:
           name: Upload artifacts to S3
           command: |
@@ -333,7 +340,6 @@ jobs:
             else
                 make upload-release SHOW=1
             fi
-      - persist-artifacts
 
   build-macos-m1:
     macos:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -330,13 +330,15 @@ jobs:
       - early-returns
       - build-steps
       - test-steps
+      - persist-artifacts
       - run:
           name: Upload artifacts to S3
           command: |
-            if [[ -n $CIRCLE_BRANCH && "<<parameters.upload>>" == "yes" ]]; then
+            if [[ -n $CIRCLE_BRANCH ]]; then
```
oshadmi, line 337: Would it be easier to add  `upload: "yes"` when calling the job? (from `upload-artifacts`)

.circleci/config.yml
```diff
@@ -410,10 +414,10 @@ jobs:
           command: |
             mkdir -p bin
             ln -s ~/workspace/artifacts bin/artifacts
-            if [[ -n $CIRCLE_TAG && "<<parameters.staging-lab>>" != 1 ]]; then
+            if [[ -n $CIRCLE_TAG ]]; then
                 make upload-release SHOW=1
             else
-                make upload-release SHOW=1 STAGING=1
+                make upload-artifacts SHOW=1
```
oshadmi, line 420: No need for `STAGING` anymore?

oshadmi: Decided to cleanup later

.circleci/config.yml
oshadmi: Uploaded by default or specifically set to "yes" - so this parameter is redundant.
Decided to cleanup later.

oshadmi (approved):

PR #1178, opened 2024-01-30
[2.4] Fix mget test failures with redis unstable [MOD-6285, MOD-5943]

ephraimfeldblum:
Backport of #1110 to 2.4

Only the non RESP-3 parts are relevant.

### DIFF
--- a/tests/pytest/test.py
+++ b/tests/pytest/test.py
@@ -10,7 +10,6 @@ from RLTest import Env
 from includes import *
 from redis.client import NEVER_DECODE
 from RLTest import Defaults
-from packaging import version
 
 Defaults.decode_responses = True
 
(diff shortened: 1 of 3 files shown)


tests/pytest/test.py
```diff
@@ -381,44 +381,38 @@ def testMgetCommand(env):
     """Test REJSON.MGET command"""
     r = env
 
-    # Skip on Redis Unstable
-    _version = "99"
-    res = r.con.execute_command('INFO')
-    if(version.parse(res['redis_version']) >= version.parse(_version)):
```
nafraf, line 387: If we don't need to check the version, then I think we can remove the import:
``from packaging import version``

nafraf (commented):
Should we continue skipping the ``test_multi:testMGetCommand``?
Is it possible to use the same fix?

nafraf (approved):

PR #1184, opened 2024-02-20
Prepare the project for working on reducing the RAM consumption.

iddm:
1. Adds a temporary feature gate which allows switching between the JSON parser and the new custom JSON parser.
2. Abstracts away the critical interfaces, following the SRP and simply encapsulation/isolation principles.
3. Implements the interfaces for the storage types.
4. Due to the new abstractions, it allows for change of the storage backend without changing the data manager itself, which is helpful for the unit tests with mocks, and allows the replacement of the storage implementation from one to another without pain, given that all required interfaces are realised (traits).
5. Adds a few more narrow test cases to measure the memory consumption of the ijson and new custom json parser storages.
6. The new `MemoryConsumption` trait isn't optimised for the new custom JSON parser due to it being difficult to measure the memory allocated for the hashmap for the reasons: 1) it varies across different hash map implementations 2) across different hash algorithms and the way they are implemented in/for the target hash map 3) implementation dependent and violates encapsulation. The best way to do this is either for the hash map implementation to provide this, or to use a custom, targeted allocator that counts the allocated memory per this storage. This is possible right now with `rust nightly`, or if we fork the needed hashmaps (and then, vectors too, perhaps, or list-packs, whatever) and change the allocator API to our own, which we then also implement for our own custom type, which internally would count the memory allocated and delegate the allocation/deallocation requests to the global (Redis) allocator.


iddm:
Please note that I won't address the test issues in this PR - those are happening due to the Python tests not asserting correctly - those assume the order of values where it isn't supposed to be. We have already addressed this in RedisGears a while ago, and it should be done here as well, but this isn't our current goal as of now.

MeirShpilraien (approved):

PR #1190, opened 2024-02-28
MOD-6501, MOD-4551, MOD-4856, MOD-5714: fix integers over i64max to be promoted to f64

ephraimfeldblum:
**Description**

Indexing `NUMERIC` fields has unchanged behavior.
* A `NUMERIC` field with value greater than `i64::min` (`-2^63`), but less than `i64::max` (`2^63 - 1`) is indexed as `i64`. It does not lose precision. Its type is `integer`.
* A `NUMERIC` field with value greater than `i64::max` (`2^63 - 1`), but less than `u64::max` (`2^64 - 1`) is indexed as `u64`. It does not lose precision. Its type is `integer`.
* A `NUMERIC` field with value greater than `u64::max` (`2^64 - 1`), or lower than `i64::min` (`-2^63`), or containing a decimal point, is indexed as `f64`. Its type is `number`. This may implicate losing precision, but this is not a breaking change.

`NUMINCRBY` behavior changed to prevent crash.
* `i64 + i64` whose result is between `i64::min` and `i64::max` has unchanged behavior. Its result is stored as `i64`, its type is `integer`.
* `i64 + i64` whose result is greater than `i64::max` or less than `i64::min` explicitly wraps instead of invoking undefined behavior. Its result is stored as `i64`, its type is `integer`.
* `i64 + u64` and `u64 + u64` behavior is unchanged. Its result is stored as `f64`, its type is `number`. This may implicate losing precision, but this is not a breaking change.
* `u64 + i64` used to crash. Now it behaves consistently with the prior case.
* If either argument is `f64`, the result is stored as `f64`, with type `number`.

`DROP` of `integer`s stored as `u64` no longer crashes.

This PR only affects `NUMERIC` fields using vanilla JSON, not those using CRDT.

**Which issues this PR fixes**
1. [MOD-6501](https://redislabs.atlassian.net/browse/MOD-6501)
2. [MOD-4551](https://redislabs.atlassian.net/browse/MOD-4551)
3. [MOD-4856](https://redislabs.atlassian.net/browse/MOD-4856)
4. [MOD-5714](https://redislabs.atlassian.net/browse/MOD-5714)
5. [#875](https://github.com/RedisJSON/RedisJSON/issues/875)
6. [#3726 (Search)](https://github.com/RediSearch/RediSearch/issues/3726)

**Main objects this PR modified**
1. IValue Manager

[MOD-6501]: https://redislabs.atlassian.net/browse/MOD-6501?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MOD-4551]: https://redislabs.atlassian.net/browse/MOD-4551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MOD-4856]: https://redislabs.atlassian.net/browse/MOD-4856?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MOD-5714]: https://redislabs.atlassian.net/browse/MOD-5714?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/pytest/test.py
```diff
@@ -1420,6 +1420,36 @@ def testRDBUnboundedDepth(env):
...
+    # set u64 is promoted to f64
+    nn = random.randint(i64max + 1, u64max)
+    while nn == float(nn):
+        nn = random.randint(i64max + 1, u64max)
+    r.expect('JSON.SET', 'num', '$', nn).ok()
+    res = r.execute_command('JSON.GET', 'num', '$')
+    val = json.loads(res)[0]
+    r.assertNotEqual(val, nn)     # u64
+    r.assertEqual(val, float(nn)) # f64
+
```
iddm, line 1452: Random values in unit tests cause them to be not repeatable. As soon as one test will pass and another will fail without any change, people lose confidence in them, undermining their value. Printing a reproduction script is not enough. That said, randomized edge case testing and fuzz testing can provide value.

Can we make the test here a little simpler, providing the correct `nn` value straight away?

.gitignore
```diff
@@ -33,5 +33,6 @@
 .settings/
 
 wordlist.dic
+config.txt
 
 venv/
```
iddm, line 38: Once we are here, can you please add an empty line at the end of the file? :-)

.gitignore
```diff
@@ -33,5 +33,6 @@
 .settings/
 
 wordlist.dic
+config.txt
```
iddm, line 36: What's this `config.txt` about? :-)

tests/pytest/test.py
ephraimfeldblum: The point of the random values was that it should work for *every* value within the ranges provided. But looping over all 2^63 values in that range is a huge waste of time.

.gitignore
ephraimfeldblum: An automation file for RLTest. @MeirShpilraien can explain more in detail.

tests/pytest/test.py
iddm: This should be testable through the corner values: the min and close to min, somewhere in the middle, the max and close to the max. However, randomness in the tests may bring randomness in the test results. Indeed, looping isn't a good option.

.gitignore
iddm: Got you. The name is too generic. I think RLTest should have a different configuration name.

MeirShpilraien: @iddm We can deprecate this name and come up with another one more specific to RLTest. maybe `rltest_config.txt`? Personally I think maybe RLTest should also look for this configuration file under some user directory (maybe something like `~/.rltest/rltest_config.txt`) and not just under the current directory.

tests/pytest/test.py
oshadmi: Randomness can be used and still allow for repeatable test execution, such as by logging the random seed, so a specific test failure can be reproduced. For example, https://github.com/RedisJSON/RedisJSON/blob/6ea15d7e795046f6170466ca232a790261718131/tests/pytest/test_short_read.py#L437

.gitignore
iddm: Yes, indeed, much better than just `config.txt` :-) But even better, just as you said, a config directory, I think. However, we should use something like `XDG_CONFIG_DIR` or something if it is available; if not, then just use `$HOME/.config/rltest` for it.

ephraimfeldblum:
replaced by #1192

PR #1191, opened 2024-02-29
Add test to verify the unicode characters parse fine.

iddm:
There were reports that Redis json modules couldn't handle the unicode characters in strings. This commit adds two tests to make sure those are handled fine for RedisJSON.

P.S. I apologise for the "irrelevant" changes, my IDE removes all the free-standing spaces on lines, as well as the trailing spaces. This actually helps to reduce the 'char noise' and reduces the size of the files :-)

### DIFF
--- a/redis_json/src/ivalue_manager.rs
+++ b/redis_json/src/ivalue_manager.rs
@@ -716,8 +716,12 @@ impl<'a> Manager for RedisIValueJsonKeyManager<'a> {
 mod tests {
     use super::*;
 
+    static single_thread_test_mutex: std::sync::Mutex<()> = std::sync::Mutex::new(());
+
     #[test]
     fn test_get_memory() {
+        let _guard = single_thread_test_mutex.lock();
+
         let manager = RedisIValueJsonKeyManager {
             phantom: PhantomData,
         };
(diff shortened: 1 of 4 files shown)


tests/pytest/test.py
```diff
@@ -1414,12 +1414,19 @@ def testRDBUnboundedDepth(env):
...
     r.expect('JSON.SET', 'doc', '$..__leaf', json_value).ok()
-    
+
     # RDB dump and restore the key 'doc' and check that the key is still valid
     dump = env.execute_command('dump', 'doc', **{NEVER_DECODE: []})
     r.expect('RESTORE', 'doc1', 0, dump).ok()
     r.expect('JSON.GET', 'doc1', '$..__leaf..__deep_leaf').equal('[420]')
 
+def testUnicodeCharacters(env):
+    # Test JSONPath filter with unicode characters
```
oshadmi, line 1424: Should we test JSONPath filter?

tests/pytest/test.py
```diff
@@ -1414,12 +1414,19 @@ def testRDBUnboundedDepth(env):
...
+    r = env
+    # Test with some weird unicode characters.
+    r.assertOk(r.execute_command('JSON.SET', 'test', '$', '{"\u00a0": "⍢∪⇰"}'))
+    r.assertOk(r.execute_command('JSON.SET', 'test', '$', '{"\u00a0": "\u00a0\u00a0"}'))
```
oshadmi, line 1428: Can also confirm we can get it back with `JSON.GET`

tests/pytest/test_multi.py
```diff
@@ -227,6 +227,8 @@ def testSetAndGetCommands(env):
     res = r.execute_command('JSON.GET', 'doc1')
     r.assertEqual(res, '"inizio"')
 
+    # Test unicode characters
+    r.assertOk(r.execute_command('JSON.SET', 'doc1', '$', '"\u00a0"'))
```
oshadmi, line 231: Can also confirm we can get it back with `JSON.GET`

redis_json/src/ivalue_manager.rs
```diff
@@ -740,4 +744,32 @@ mod tests {
         let res = manager.get_memory(&value).unwrap();
         assert_eq!(res, 903);
     }
+
+    /// Tests the deserialiser of IValue for a string with unicode
+    /// characters, to ensure that the deserialiser can handle
+    /// unicode characters well.
+    #[test]
+    fn test_unicode_characters() {
+        let _guard = single_thread_test_mutex.lock();
```
oshadmi, line 753: What resource needs to be mutual exclusive here?

oshadmi (commented):
👍🏼 Few comments

tests/pytest/test.py
iddm: We can and probably should; it is just that this particular problem wasn't related to the JSON filter at all, and it also was according to the original ticket description. This particular problem happens very early and isn't a part of JSON path filter processing or any actual processing at all. It is only about the deserialisation of a string into a Rust object, so it happens before everything else.

tests/pytest/test.py
iddm: Sure.

redis_json/src/ivalue_manager.rs
iddm: The `ijson` string interning implementation. It is hidden deep inside the implementation. Guy Korland patched it so that we don't use a thread-safe string interning simply because we don't need it, as it happens all within a single thread anyway. But the tests run concurrently, and so there is a race condition. There are different ways to solve this, one is just to run the tests with single thread, another is this. I picked this one because no one else needs to know about it from the outside, as it is only in the tests, and we actually don't have unit tests at all. The less people know and the less they are required to know about "how the project needs to be tested" - the better. Otherwise, we will have to change all our makefiles and everything else to run `cargo test` with just one thread, list it everywhere for everyone.... The people who are familiar with Rust, and people in general, in my opinion, should be able to run the tests with just `cargo test`. Everything else is hidden behind the implementation. By the way, we also do it this way in RedisGears, for example. So, let's consider this a just a unit-test implementation detail, if you don't mind :-)

tests/pytest/test_multi.py
iddm: I'll actually remove it from here, this is not needed there.

iddm: I apologise for this one. I just quickly looked for the first test file I could open in my editor and added a line there to confirm it happens/doesn't happen. This test should be a separate thing, so I removed these two lines from here.

oshadmi (approved):

PR #1198, opened 2024-03-27
Fix the tests.sh script.

iddm:
If the status file does not exist and we are creating it, don't fail with an error if the file doesn't exist; it obviously won't exist until we create it. This change creates the file in case it didn't exist, resulting in the proper continuation of the script and a proper exit code.

This change is also required for the repeatability of the CI steps locally. Namely in https://github.com/RedisJSON/RedisJsonHdt/pull/226


ephraimfeldblum (approved):
LGTM

PR #1211, opened 2024-06-23
[MOD-7289,   MOD-7306, MOD-7305] Update base RDBs URL

ephraimfeldblum:
update s3 bucket url

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -374,6 +374,7 @@ jobs:
           name: Build & Test
           shell: /bin/bash -l -eo pipefail
           command: |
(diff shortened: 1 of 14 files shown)


raz-mon (approved):

iddm:
The provided solution for the coverage calculation in CircleCI is a hotfix, it shouldn't be fixed this way. But since the priority is to make the CI green first, a proper solution might be done later. Not to mention that we should get rid of the "readies", makefiles and other magical things as soon as possible.

eranhd (approved):

PR #1212, opened 2024-06-24
MOD-7293 fix AOF commands duplicated in JSON.MSET

eranhd:
1. Write to AOF outside the iterator
2. Added notify_keyspace_event API to manager

### DIFF
--- a/redis_json/src/commands.rs
+++ b/redis_json/src/commands.rs
@@ -193,7 +193,6 @@ pub fn json_set<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>) -
                 if op != SetOptions::NotExists {
                     redis_key.set_value(Vec::new(), val)?;
                     redis_key.apply_changes(ctx, "json.set")?;
-                    ctx.replicate_verbatim();
                     REDIS_OK
                 } else {
                     Ok(RedisValue::Null)
@@ -204,7 +203,6 @@ pub fn json_set<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>) -
                     let updated = apply_updates::<M>(&mut redis_key, val, update_info);
                     if updated {
                         redis_key.apply_changes(ctx, "json.set")?;
-                        ctx.replicate_verbatim();
                         REDIS_OK
                     } else {
                         Ok(RedisValue::Null)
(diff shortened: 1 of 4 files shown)


redis_json/src/commands.rs
```diff
@@ -354,10 +354,13 @@ pub fn json_mset<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>)
                 redis_key.set_value(Vec::new(), value)?
             };
             if updated {
-                redis_key.apply_changes(ctx, "json.mset")?
+                redis_key.notify_keyspace_event(ctx, "json.mset")?
```
MeirShpilraien, line 357: I think the separation you are doing here is right, manager should probably not handle the replication. But I believe we should make this separation all over the code, WDYT?
In addition, maybe we should find a more general name to it? Because on HDT it also need to apply the changes on the key and fire the effects.

tests/pytest/test.py
```diff
@@ -1487,6 +1488,48 @@ def test_promote_u64_to_f64(env):
     r.assertEqual(val, float(2 * i64max + 3))            # u64 + u64 promotes to f64. as prior, not breaking
     r.expect('JSON.TYPE', 'num', '$').equal(['number'])  # promoted
 
+
+def test_mset_replication_in_aof(env):
+    if not env.useAof:
+        env.skip()
```
MeirShpilraien, line 1494:
```suggestion
def test_mset_replication_in_aof():
    env = Env(useAof=True)
```
We probably want to create the env here with aof enabled

MeirShpilraien (commented):
:+1: 
Few suggestions

redis_json/src/commands.rs
eranhd: how it's managed in redis? there are calling to replication and notify in the commands? or it's something that happened in the behind the scenes?

MeirShpilraien: Notify is per command.
Replication is a totally different mechanism.

redis_json/src/ivalue_manager.rs
```diff
@@ -290,6 +290,14 @@ impl<'a> WriteHolder<IValue, IValue> for IValueKeyHolderWrite<'a> {
         }
     }
 
+    fn notify_keyspace_event(&mut self, ctx: &Context, command: &str) -> Result<(), RedisError> {
```
ephraimfeldblum, line 293: If we're splitting the functionality of `apply_changes` then we should have it call `notify_keyspace_event` to reduce duplication. eg something like
```rs
fn apply_changes(&mut self, ctx: &Context, command: &str) -> Result<(), RedisError> {
    self
        .notify_keyspace_event(ctx, command)
        .inspect(|_| ctx.replicate_verbatim())
}
```

MeirShpilraien (approved):

MeirShpilraien:
Let make sure to do the relevant changes on HDT as well.

MeirShpilraien (approved):
:+1:

PR #1214, opened 2024-07-04
MOD-7342 fix centos7&amzn2 build

eranhd:
readies updated, and now coentos7, amzn2 are using vault instead mirror


.circleci/config.yml
```diff
@@ -207,6 +207,10 @@ commands:
           command: |
             ROOT=$PWD
             cd build/docker
+            if [[ "<<parameters.platform>>" == "centos7" ]]; then
+                sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-*
+                sed -i 's|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-*
+            else
```
iddm, line 213: Shouldn't we do this in "readies" instead? https://github.com/RedisLabsModules/readies/blob/master/bin/getclang#L114

DvirDukhan (approved):

PR #1217, opened 2024-07-10
MOD-7368, make c_api public function

eranhd:
we use them in HDT lib (in redis_json_module_export_shared_api macro) so we need to keep it public.


MeirShpilraien (approved):

ephraimfeldblum (approved):
LGTM

PR #1220, opened 2024-07-14
Fix upload artifacts on macOS-x86

ephraimfeldblum:
The `macos.x86.medium.gen2` resource class was deprecated on June 28, 2024.

By default, Xcode 12.2+ will create universal binaries, compiling to a single executable that supports both x86_64 and ARM64 based CPUs.


sbin/upload-artifacts
```diff
@@ -113,8 +114,10 @@ s3_upload() {
...
 	if [[ $SNAPSHOT == 1 ]]; then
-		for file in `ls ${prefix}.*${PLATFORM}*.zip ${prefix}.*${PLATFORM}*.tgz 2> /dev/null`; do
-			s3_upload_file $file $upload_dir
+		for platform in ${PLATFORMS[@]}; do
+			for file in `ls ${prefix}.*${platform}*.zip ${prefix}.*${platform}*.tgz 2> /dev/null`; do
+				s3_upload_file $file $upload_dir
+			done
 		done
 	else
 		for file in `ls ${prefix}.*.zip ${prefix}.*.tgz 2> /dev/null`; do
```
eranhd, line 123: update here as well

ephraimfeldblum: not really sure what needs updating here. `prefix` is just `"rejson-oss"`

eranhd (approved):

iddm:
This PR introduced too much noise in the commit tree. Please squash the next time.

PR #1226, opened 2024-08-08
MOD-7264: Fix JSONPath filter with recursive descent processing elements twice

ephraimfeldblum:


redis_json/src/commands.rs
```diff
@@ -1740,22 +1696,11 @@ where
 pub fn json_clear<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>) -> RedisResult {
     let mut args = args.into_iter().skip(1);
     let key = args.next_arg()?;
-    let paths = args.try_fold::<_, _, Result<Vec<Path>, RedisError>>(
```
ephraimfeldblum, line 1743: @iddm I believe this to be a bug. This `try_fold` implies that multiple paths may be passed to `JSON.CLEAR`, and that if, for *any* path arg, if `arg.try_as_str().is_err()` then the function shall return early. Even if there was a legal path that was passed. Whereas the [documentation](https://redis.io/docs/latest/commands/json.clear/) clearly states that, at most, one singular path may be passed.

If anything, I believe the args length should be checked, and if it is not correct, then return `Err(WrongArity)`.

MeirShpilraien:
@ephraimfeldblum as we discussed. Please try to think how we can split this work to multiple smaller PR's which is easier to review. Make sure to have a detailed message that explain the changes. If there are some API changes, make sure to specify them and explain why it was done.

Last, make sure to split bugs to separate PR's so it will be easier to cherrypick (if needed).

ephraimfeldblum:
> @ephraimfeldblum as we discussed. Please try to think how we can split this work to multiple smaller PR's which is easier to review.
Sure thing.
> Make sure to have a detailed message that explain the changes. If there are some API changes, make sure to specify them and explain why it was done.
Almost none to external APIs. iirc, just regarding ownership to prevent having to clone unnecessarily.
> Last, make sure to split bugs to separate PR's so it will be easier to cherrypick (if needed).
Yep. Already split removed from here.

PR #1230, opened 2024-08-16
Add alpine CI workflow for x86.

iddm:


ephraimfeldblum (commented):
Is there a working example of this flow?

iddm:
> Is there a working example of this flow? I'd appreciate seeing it building and running before signing off on it.



> Is there a working example of this flow? I'd appreciate seeing it building and running before signing off on it.

Just as the others, like "mariner" the code is written to only support the branch name, not pull requests. So no, it will only work until `make pack`. You can find the pipeline which failed at this step.

iddm:
@ephraimfeldblum 
screenshot

ephraimfeldblum (approved):

PR #1241, opened 2024-08-27
Fix ci 2.4

iddm:


ephraimfeldblum (approved):

iddm:
The coverage isn't related to this PR, it is the first run over a long period of time. If needs to be address, then separately.

PR #1243, opened 2024-08-27
Remove the release-qa-tests

iddm:
The opereto is long gone and is not used by all other projects, not to mention it has been inaccessible for a long time and only prevents from following the release pipeline.


iddm:
@tezc

tezc (approved):

PR #1254, opened 2024-08-29
Always use the cargo package version as the module version.

iddm:
Changes the way the module version is specified to always get the crate version from the Cargo.toml. This greatly reduces the chances of accidentally forgetting to update the module version.


ephraimfeldblum (approved):
LGTM

ephraimfeldblum (approved):

PR #1265, opened 2024-09-10
MOD-7644 Support Alpine ARM64

eranhd:
This PR build and tests Alpine arm64 and upload to s3 alpine's module artifacts

https://github.com/RedisJSON/RedisJSON/actions/runs/10901675977/job/30251947716
https://github.com/RedisJSON/RedisJSON/actions/runs/10901604582/job/30251708038


ephraimfeldblum (dismissed):

.github/actions/upload-artifacts-to-s3/action.yml
```diff
@@ -0,0 +1,48 @@
...
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v3
+      with: # todo: use role instead of access key
+        aws-access-key-id: ${{ inputs.aws-access-key-id }}
+        aws-secret-access-key: ${{ inputs.aws-secret-access-key }}
+        aws-region: "us-east-1"
+    - name: Upload artifacts to S3 - staging
+      shell: bash
+      run: |
+        echo ::group::install aws cli
```
iddm, line 32: Do we need these "echo" ?

sbin/pack.sh
```diff
@@ -72,6 +72,7 @@ OSNICK=$($READIES/bin/platform --osnick)
 [[ $OSNICK == ol8 ]]     && OSNICK=rhel8
 [[ $OSNICK == rocky8 ]]  && OSNICK=rhel8
 [[ $OSNICK == rocky9 ]]  && OSNICK=rhel9
+[[ $OSNICK == NotpineForGHA3.20.3 ]]  && OSNICK=alpine3
```
iddm, line 75: It can be removed in favour of changing the `/etc/os-release` file. Check how it is done in RedisTimeSeries in the labs.

iddm (commented):
Good if it works as expected, but I don't see why we need the `echo` everywhere as this is not CircleCI and we can clearly see the steps in the GitHub Actions and whether they have started execution or not and for what reason (compared to CircleCI).

Also, we don't need hacks in the pack.sh scripts, we just need to revert the real alpine name there, this will be a little more correct for the scripts. The changes we are doing to the `/etc/os-release` are only needed for the GitHub Actions to work, not for the scripts.

.github/actions/upload-artifacts-to-s3/action.yml
eranhd: It collapses all logs under this name when we have a couple of actions in same step it is best practice to use it

sbin/pack.sh
eranhd: If I change the ID before this step it can't upload the artifacts to s3 because it use aws-cli

eranhd: in timeseriese currently we not upload artifacts

iddm:
```yaml
      - name: Pack module
        run: |
          git config --global --add safe.directory $GITHUB_WORKSPACE
          mv /etc/os-release.bak /etc/os-release
          make pack BRANCH=$TAG_OR_BRANCH SHOW=1
      - name: Workaround alpine-arm64 GHA issues (one more time)
        shell: sh
        run: |
          sed -i 's/ID=alpine/ID=NotpineForGHA/g' /etc/os-release
 ```

Comment (eranhd) [reply sbin/pack.sh:75]: but after that in uploading artifacts it should change back to NotpineForGHA

**Comment by iddm** [comment]
Even though it passes now, I think we should address this:
```
  RELEASE=1 ./sbin/upload-artifacts
  du: unrecognized option: apparent-size
  ```

ephraimfeldblum (dismissed):

eranhd:
> Even though it passes now, I think we should address this:
> 
> ```
>   RELEASE=1 ./sbin/upload-artifacts
>   du: unrecognized option: apparent-size
> ```

[[ $VERBOSE == 1 ]] && du -ah --apparent-size * || du -ah -s * I handled it here, and left a comment wht

iddm (approved):

ephraimfeldblum (approved):

iddm (approved):

PR #1267, opened 2024-09-12
Enabling the compiler to better see through `do_op`

ephraimfeldblum:
Split `remove` from `update`
* Callback passed to `update` would return `Option<()>` not because it had some value to return, but as a signaling mechanism to the caller that a separate branch should be taken in the case of `None`.
* This branch was taken only in the case of `delete_path`.
* `delete_path` now calls `remove`, which does not expect any callback.
* All other instances of calls to `update` do not require their callbacks to return an `Option`.

Merged `replace` with `update`
* Both functions were nearly identical in behavior.
* Only the behavior of their respective callbacks were different.
* Can offload the differences in behavior to their callers and the callbacks they choose to pass.

Callbacks to `update` are pure
* They do not save state from their source location to be updated from another location.
* They instead return the data back to the caller via the normal return path.


iddm:
I really like the compression here and splitting into single-responsibility clean functions. Let's finish this one. The code is unnecessarily overcomplicated in many places and the less such places we have, the better.

PR #1271, opened 2024-09-22
Fix module lib name

ephraimfeldblum:
Addendum to #1270, updating uploaded module library name from `librjson.so`/`.dylib` to `rejson.so`


eranhd (approved):
please verify it works against s3

PR #1274, opened 2024-09-26
GHA step: Sanitizer

ephraimfeldblum:


.github/actions/build-json-module-and-redis/action.yml
```diff
@@ -7,6 +7,8 @@ inputs:
     description: 'Redis version to build'
     required: true
     default: '7.2.1'
+  san:
```
iddm, line 10: It is a little difficult to guess what "san" may stand for. Can we make it longer yet fully self-explanatory?

.github/actions/run-tests/action.yml
```diff
@@ -0,0 +1,120 @@
...
+      uses: ./.github/actions/build-json-module-and-redis
+      with:
+        san: ${{ inputs.san }}
+
+    - name: Set Artifact Names
+      shell: bash
+      # Artifact names have to be unique, so we base them on the environment.
+      # We also remove invalid characters from the name.
+      id: artifact-names
+      run: | # Invalid characters include: Double quote ", Colon :, Less than <, Greater than >, Vertical bar |, Asterisk *, Question mark ?
```
iddm, line 76: Can the comment be on the next line?

.github/workflows/flow-sanitizer.yml
```diff
@@ -0,0 +1,25 @@
...
+    container:
+      image: "ubuntu:jammy"
+    steps:
+      - name: checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - uses: ./.github/actions/run-tests
+        with:
+          container: "ubuntu:jammy"
```
iddm, line 23: I wonder if we could just pass it like `${{ container.image }}` here instead of the string directly like that?

iddm (approved):
A couple of small questions, but it is good if it works. I don't like our approach in general, but that's a topic for another time, not related to this PR as it simply follows.

iddm (approved):

PR #1291, opened 2024-10-31
MOD-7499 Add JSON ACL category

ephraimfeldblum:
[MOD-7499](https://redislabs.atlassian.net/browse/MOD-7499)

[MOD-7499]: https://redislabs.atlassian.net/browse/MOD-7499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.circleci/config.yml
```diff
@@ -82,7 +82,7 @@ commands:
     parameters:
       redis_version:
         type: string
-        default: "7.2"
```
eranhd, line 85: why do we need to change it in circleci?

iddm: I think we should delete it, I agree. As a separate commit should be fine.

.github/actions/build-json-module-and-redis-with-cargo/action.yml
```diff
@@ -6,7 +6,7 @@ inputs:
   redis-ref:
     description: 'Redis version to build'
     required: true
-    default: '7.2.1'
```
iddm, line 9: Excuse me, do we really need to be that precise to specify `7.4.1` and not just `7.4`?

redis_json/Cargo.toml
```diff
@@ -22,8 +22,8 @@ ijson.workspace = true
 serde_json.workspace = true
 serde.workspace = true
 libc = "0.2"
-redis-module ={ git="https://github.com/RedisLabsModules/redismodule-rs", tag="v2.0.8", default-features = false, features = ["min-redis-compatibility-version-7-2"] }
-redis-module-macros = { git="https://github.com/RedisLabsModules/redismodule-rs", tag="v2.0.8" }
```
iddm, line 26: We should release the `redismodule-rs` version and use the new tag here.

iddm (changes requested):
We need to release the redismodule-rs version and depend on it here, instead of depending on the master branch.

.github/actions/build-json-module-and-redis-with-cargo/action.yml
ephraimfeldblum: these were the result of a `sed` from 7.2 to 7.4 :shrug:

iddm (approved):

eranhd:
Please update the description to include detailed information.

eranhd (approved):

PR #1299, opened 2024-11-20
Refactor shared string cache initialization into a separate function

eranhd:


redis_json/src/lib.rs
```diff
@@ -173,17 +173,7 @@ macro_rules! redis_json_module_create {(
...
-                .map_or(false, |res| match res {
-                    RedisValue::Array(a) => !a.is_empty(),
-                    _ => false,
-                });
-            ctx.log_notice(&format!("Initialized shared string cache, thread safe: {is_bigredis}."));
-            if let Err(e) = ijson::init_shared_string_cache(is_bigredis) {
-                ctx.log(RedisLogLevel::Warning, &format!("Failed initializing shared string cache, {e}."));
-                return Status::Err;
-            }
+            init_ijson_shared_string_cache(ctx);
```
MeirShpilraien, line 176:
```suggestion
            $crate::init_ijson_shared_string_cache(ctx);
```

redis_json/src/lib.rs
```diff
@@ -249,6 +239,21 @@ const fn dummy_init(_ctx: &Context, _args: &[RedisString]) -> Status {
     Status::Ok
 }
 
+fn init_ijson_shared_string_cache(ctx: &Context) -> Status {
```
MeirShpilraien, line 242:
```suggestion
pub fn init_ijson_shared_string_cache(ctx: &Context) -> Status {
```

redis_json/src/lib.rs
MeirShpilraien: When used on hdt we want it to take it from the json crate

redis_json/src/lib.rs
MeirShpilraien: Must be public so it can be used from hdt

MeirShpilraien (commented):
What about the missing use on of the ACL?

redis_json/src/lib.rs
```diff
@@ -249,6 +239,20 @@ const fn dummy_init(_ctx: &Context, _args: &[RedisString]) -> Status {
...
+    let is_bigredis =
+                ctx.call("config", &["get", "bigredis-enabled"])
+                .map_or(false, |res| match res {
+                    RedisValue::Array(a) => !a.is_empty(),
+                    _ => false,
+                });
+    ctx.log_notice(&format!("Initialized shared string cache, thread safe: {is_bigredis}."));
+    if let Err(e) = ijson::init_shared_string_cache(is_bigredis) {
+        ctx.log(RedisLogLevel::Warning, &format!("Failed initializing shared string cache, {e}."));
+        panic!("Failed initializing shared string cache, {e}.");
```
MeirShpilraien, line 252: Lets return error and check if we did not got an error above

MeirShpilraien (approved):

PR #1375, opened 2025-08-05
MOD-10709 pin nightly version

ephraimfeldblum:
[MOD-10709](https://redislabs.atlassian.net/browse/MOD-10709)

### DIFF
--- a/.install/getrust.sh
+++ b/.install/getrust.sh
@@ -9,12 +9,28 @@ $MODE curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
 echo "source $HOME/.cargo/env" >> $HOME/.bashrc
 source $HOME/.cargo/env
 
-# Update rustup and install nightly toolchain
+# Update rustup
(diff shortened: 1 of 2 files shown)


eranhd (approved):

gdesmott:
I think you also need to specify the pinned nightly version in `Makefile`: https://github.com/RedisJSON/RedisJSON/blob/master/Makefile#L112

I workaround it in the RediSearch CI by using the `RUST_GOOD_NIGHTLY` env variable, see https://github.com/RediSearch/RediSearch/pull/6513/files , but it would be good to fix that there at some point as well.

eranhd:
> I think you also need to specify the pinned nightly version in `Makefile`: https://github.com/RedisJSON/RedisJSON/blob/master/Makefile#L112
> 
> I workaround it in the RediSearch CI by using the `RUST_GOOD_NIGHTLY` env variable, see https://github.com/RediSearch/RediSearch/pull/6513/files , but it would be good to fix that there at some point as well.

@ephraimfeldblum

Issue #1374, opened 2025-08-05
JSON.DEL deletes only the first 10 matching object elements


LiorKogan:
```
> JSON.SET j $ '[{"x":1},{"x":2},{"x":3},{"x":4},{"x":5},{"x":6},{"x":7},{"x":8},{"x":9,"y":1},{"x":10},{"x":11},{"x":12}]' 
OK
> JSON.DEL j '$[?(@.x>0)]'
(integer) 10
> JSON.GET j 
"[{\"x\":11},{\"x\":12}]"
```

Source: https://stackoverflow.com/questions/79725743

luca-bassoricci:
The same applies also with array contains primitive elements
```
JSON.DEL test2
JSON.SET test2 "$" '{"localIndex":[]}'
JSON.ARRAPPEND test2 "$.localIndex" '1'
JSON.ARRAPPEND test2 "$.localIndex" '1'
JSON.ARRAPPEND test2 "$.localIndex" '1'
JSON.ARRAPPEND test2 "$.localIndex" '1'
JSON.ARRAPPEND test2 "$.localIndex" '1'
JSON.ARRAPPEND test2 "$.localIndex" '1'
JSON.ARRAPPEND test2 "$.localIndex" '1'
JSON.ARRAPPEND test2 "$.localIndex" '1'
JSON.ARRAPPEND test2 "$.localIndex" '1'
JSON.ARRAPPEND test2 "$.localIndex" '1'
JSON.ARRAPPEND test2 "$.localIndex" '1'
JSON.ARRAPPEND test2 "$.localIndex" '1'
JSON.ARRAPPEND test2 "$.localIndex" '1'
JSON.ARRAPPEND test2 "$.localIndex" '14'
JSON.ARRAPPEND test2 "$.localIndex" '15'
JSON.DEL test2 "$.localIndex[?(@ == 1)]"
JSON.GET test2 "$.localIndex[?(@ == 1)]"
```

luca-bassoricci:
Today I faced the same problem not with arrays but with objects
```
json.set test $ '{"root" : {"1" : {"value":1},"2" : {"value":2},"3" : {"value":3},"4" : {"value":4},"5" : {"value":5}}}' -> OK
json.del test '$.root[?(@.value > 2)]' (or json.del test '$.root..[?(@.value > 2)]') -> 3 (correct)
JSON.GET test '$.root' -> [{"root": {"1": {"value": 1},"5": {"value": 5}}}]
```

[...]

PR #1378, opened 2025-09-01
MOD-11067 Fix object path findings with filters

AvivDavid23:
https://github.com/RedisJSON/RedisJSON/issues/1374#issuecomment-3182658773

### DIFF
--- a/json_path/src/json_path.rs
+++ b/json_path/src/json_path.rs
@@ -990,23 +990,53 @@ impl<'i, UPTG: UserPathTrackerGenerator> PathCalculator<'i, UPTG> {
                             /* lets expend the array, this is how most json path engines work.
                              * Personally, I think this if should not exists. */
                             let values = json.values().unwrap();
+                            let items_iter = if json.get_type() == SelectValueType::Object {
+                                Some(json.items().unwrap())
+                            } else {
+                                None
+                            };
                             if let Some(pt) = path_tracker {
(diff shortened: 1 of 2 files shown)


json_path/src/json_path.rs
```diff
@@ -990,23 +990,53 @@ impl<'i, UPTG: UserPathTrackerGenerator> PathCalculator<'i, UPTG> {
...
                                 );
-                                for (i, v) in values.enumerate() {
-                                    trace!("calc_internal v {:?}", &v);
-                                    if self.evaluate_filter(curr.clone().into_inner(), v, calc_data)
-                                    {
-                                        let new_tracker = Some(create_index_tracker(i, &pt));
-                                        self.calc_internal(
-                                            pairs.clone(),
+                                // if the json is an object, then we need to iterate over the items and create a tracker for each key
+                                if let Some(items) = items_iter {
```
ephraimfeldblum: Would it be possible to partially deduplicate the code in this if-else by implementing an enum that abstracts over the iterator type? eg
```
enum Iter<'a> {
  Arr(Box<dyn Iterator<Item = (usize, &'a IValue)> + 'a>),
  Obj(Box<dyn Iterator<Item = (&'a str, &'a IValue)> + 'a>),
}
impl Iter {
  fn type_(&self) -> SelectValueType;
}
```

PR #1385, opened 2025-09-16
Remove alpine job configuration from event workflows

eranhd:


gabsow (approved):
Looks good, can be merged.

PR #1394, opened 2025-09-29
MOD-11358 integrate ijson to the json module

AvivDavid23:
- Support returning either Owned or Borrowed values from iterators, leading to whole changes in code flow.
- Changing c_api, introducing a new api to allow to always return a reference to the underlying data(since we can't always return a reference to the SelectValue from the iterators themselves

### DIFF
--- a/json_path/src/json_node.rs
+++ b/json_path/src/json_node.rs
@@ -12,6 +12,15 @@ use crate::select_value::{SelectValue, SelectValueType, ValueRef};
 use ijson::{array::ArrayIterItem, DestructuredRef, IString, IValue, ValueType};
 use serde_json::Value;
 
+impl<'a> From<ArrayIterItem<'a>> for ValueRef<'a, IValue> {
+    fn from(item: ArrayIterItem<'a>) -> Self {
+        match item {
+            ArrayIterItem::Borrowed(val) => ValueRef::Borrowed(val),
+            ArrayIterItem::Owned(val) => ValueRef::Owned(val),
+        }
+    }
+}
+
 impl SelectValue for Value {
     fn get_type(&self) -> SelectValueType {
         match self {
@@ -158,10 +167,7 @@ impl SelectValue for IValue {
 
     fn values<'a>(&'a self) -> Option<Box<dyn Iterator<Item = ValueRef<'a, Self>> + 'a>> {
         match self.destructure_ref() {
-            DestructuredRef::Array(arr) => Some(Box::new(arr.iter().map(|item| match item {
-                ArrayIterItem::Borrowed(val) => ValueRef::Borrowed(val),
-                ArrayIterItem::Owned(val) => ValueRef::Owned(val),
-            }))),
+            DestructuredRef::Array(arr) => Some(Box::new(arr.iter().map(|item| item.into()))),
             DestructuredRef::Object(o) => Some(Box::new(o.values().map(|v| ValueRef::Borrowed(v)))),
             _ => None,
         }
@@ -198,12 +204,8 @@ impl SelectValue for IValue {
     }
 
     fn get_index<'a>(&'a self, index: usize) -> Option<ValueRef<'a, Self>> {
-        self.as_array().and_then(|arr| {
-            arr.iter().nth(index).map(|item| match item {
-                ArrayIterItem::Borrowed(val) => ValueRef::Borrowed(val),
-                ArrayIterItem::Owned(val) => ValueRef::Owned(val),
-            })
-        })
+        self.as_array()
+            .and_then(|arr| arr.iter().nth(index).map(|item| item.into()))
     }
 
     fn is_array(&self) -> bool {
--- a/json_path/src/json_node.rs
+++ b/json_path/src/json_node.rs
@@ -48,9 +48,9 @@ impl SelectValue for Value {
         }
     }
 
-    fn items<'a>(&'a self) -> Option<Box<dyn Iterator<Item = (&'a str, &'a Self)> + 'a>> {
+    fn items<'a>(&'a self) -> Option<Box<dyn Iterator<Item = (&'a str, ValueRef<'a, Self>)> + 'a>> {
         match self {
-            Self::Object(o) => Some(Box::new(o.iter().map(|(k, v)| (&k[..], v)))),
+            Self::Object(o) => Some(Box::new(o.iter().map(|(k, v)| (&k[..], ValueRef::Borrowed(v))))),
             _ => None,
         }
     }
@@ -71,9 +71,9 @@ impl SelectValue for Value {
         }
     }
 
-    fn get_key<'a>(&'a self, key: &str) -> Option<&'a Self> {
+    fn get_key<'a>(&'a self, key: &str) -> Option<ValueRef<'a, Self>> {
         match self {
-            Self::Object(o) => o.get(key),
+            Self::Object(o) => o.get(key).map(|v| ValueRef::Borrowed(v)),
             _ => None,
         }
     }
@@ -174,9 +174,9 @@ impl SelectValue for IValue {
         }
     }
 
-    fn items<'a>(&'a self) -> Option<Box<dyn Iterator<Item = (&'a str, &'a Self)> + 'a>> {
+    fn items<'a>(&'a self) -> Option<Box<dyn Iterator<Item = (&'a str, ValueRef<'a, Self>)> + 'a>> {
         match self.destructure_ref() {
-            DestructuredRef::Object(o) => Some(Box::new(o.iter().map(|(k, v)| (k.as_str(), v)))),
+            DestructuredRef::Object(o) => Some(Box::new(o.iter().map(|(k, v)| (k.as_str(), ValueRef::Borrowed(v))))),
             _ => None,
         }
     }
@@ -193,8 +193,8 @@ impl SelectValue for IValue {
         self.is_empty()
     }
 
-    fn get_key<'a>(&'a self, key: &str) -> Option<&'a Self> {
-        self.as_object().and_then(|o| o.get(key))
+    fn get_key<'a>(&'a self, key: &str) -> Option<ValueRef<'a, Self>> {
+        self.as_object().and_then(|o| o.get(key).map(|v| ValueRef::Borrowed(v)))
     }
 
     fn get_index<'a>(&'a self, index: usize) -> Option<ValueRef<'a, Self>> {
--- a/json_path/src/json_node.rs
+++ b/json_path/src/json_node.rs
@@ -8,8 +8,8 @@
  */
 
 /// Use `SelectValue`
-use crate::select_value::{SelectValue, SelectValueType};
-use ijson::{DestructuredRef, IString, IValue, ValueType};
+use crate::select_value::{SelectValue, SelectValueType, ValueRef};
+use ijson::{array::ArrayIterItem, DestructuredRef, IString, IValue, ValueType};
 use serde_json::Value;
 
 impl SelectValue for Value {
@@ -33,10 +33,10 @@ impl SelectValue for Value {
         }
     }
 
-    fn values<'a>(&'a self) -> Option<Box<dyn Iterator<Item = &'a Self> + 'a>> {
+    fn values<'a>(&'a self) -> Option<Box<dyn Iterator<Item = ValueRef<'a, Self>> + 'a>> {
         match self {
-            Self::Array(arr) => Some(Box::new(arr.iter())),
-            Self::Object(o) => Some(Box::new(o.values())),
+            Self::Array(arr) => Some(Box::new(arr.iter().map(|v| ValueRef::Borrowed(v)))),
+            Self::Object(o) => Some(Box::new(o.values().map(|v| ValueRef::Borrowed(v)))),
             _ => None,
         }
     }
@@ -78,9 +78,9 @@ impl SelectValue for Value {
         }
     }
 
-    fn get_index(&self, index: usize) -> Option<&Self> {
+    fn get_index<'a>(&'a self, index: usize) -> Option<ValueRef<'a, Self>> {
         match self {
-            Self::Array(arr) => arr.get(index),
+            Self::Array(arr) => arr.get(index).map(|v| ValueRef::Borrowed(v)),
             _ => None,
         }
     }
@@ -156,10 +156,13 @@ impl SelectValue for IValue {
         self.as_object().map_or(false, |o| o.contains_key(key))
     }
 
-    fn values<'a>(&'a self) -> Option<Box<dyn Iterator<Item = &'a Self> + 'a>> {
+    fn values<'a>(&'a self) -> Option<Box<dyn Iterator<Item = ValueRef<'a, Self>> + 'a>> {
         match self.destructure_ref() {
-            DestructuredRef::Array(arr) => Some(Box::new(arr.iter())),
-            DestructuredRef::Object(o) => Some(Box::new(o.values())),
+            DestructuredRef::Array(arr) => Some(Box::new(arr.iter().map(|item| match item {
+                ArrayIterItem::Borrowed(val) => ValueRef::Borrowed(val),
+                ArrayIterItem::Owned(val) => ValueRef::Owned(val),
+            }))),
+            DestructuredRef::Object(o) => Some(Box::new(o.values().map(|v| ValueRef::Borrowed(v)))),
             _ => None,
         }
     }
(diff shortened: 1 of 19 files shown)


eranhd:
@AvivDavid23 what happens to the coverage?

redis_json/src/commands.rs
```diff
@@ -1716,7 +1722,7 @@ where
         Some(root) => find_all_values(path, root, |v| v.get_type() == SelectValueType::Object)?
             .iter()
             .map(|v| {
-                v.map_or(RedisValue::Null, |v| {
+                v.clone().map_or(RedisValue::Null, |v| {
```
ephraimfeldblum: Should this really be cloned?

redis_json/src/commands.rs
```diff
@@ -395,7 +395,7 @@ fn find_paths<T: SelectValue, F: FnMut(&T) -> bool>(
     let res = calc_once_with_paths(query, doc);
     Ok(res
         .into_iter()
-        .filter(|e| f(e.res))
+        .filter(|e| f(e.res.clone()))
```
ephraimfeldblum: Should this be cloned?
Should this `.filter().map()` not be a single `.filter_map()`?

redis_json/src/c_api.rs
```diff
@@ -278,6 +284,30 @@ pub fn json_api_get<M: Manager>(_: M, val: *const c_void, path: *const c_char) -
...
+    _: M,
+    json: *const c_void,
+    index: size_t,
+    value: *mut c_void,
+) -> c_int {
+    let json = unsafe { &*(json.cast::<&mut M::V>()) };
+    match json.get_type() {
+        SelectValueType::Array => {
+            if let Some(element) = json.get_index(index) {
+                let cloned_value = element.inner_cloned();
```
ephraimfeldblum: Should we truly be doing a deep copy here? What if the inner value is another Array or Object? I think we should check the type and do only a shallow copy.

redis_json/src/c_api.rs
```diff
@@ -311,12 +343,24 @@ pub fn json_api_free_key_values_iter<'a, M: Manager>(_: M, iter: *mut c_void)
...
 }
 
+pub fn json_api_alloc_json<M: Manager>(_: M) -> *mut c_void {
+    Box::into_raw(Box::new(M::V::default())).cast::<c_void>()
+}
+
+pub fn json_api_free_json<M: Manager>(_: M, json: *mut c_void) {
+    let json = json.cast::<M::V>();
+    let json = unsafe { Box::from_raw(json) };
+    drop(json);
```
ephraimfeldblum: And here we should also do a type check to determine if we truly need to free.

json_path/src/json_node.rs
```diff
@@ -190,12 +197,18 @@ impl SelectValue for IValue {
...
+    fn get_key<'a>(&'a self, key: &str) -> Option<ValueRef<'a, Self>> {
+        self.as_object()
+            .and_then(|o| o.get(key).map(|v| ValueRef::Borrowed(v)))
     }
 
-    fn get_index(&self, index: usize) -> Option<&Self> {
-        self.as_array().and_then(|arr| arr.get(index))
+    fn get_index<'a>(&'a self, index: usize) -> Option<ValueRef<'a, Self>> {
+        self.as_array().and_then(|arr| {
+            arr.iter().nth(index).map(|item| match item {
```
ephraimfeldblum: Could `impl From<ArrayIterItem> for ValueRef` to make this a little cleaner. I think it's useful to have it anyway because they are performing the same role and it should be seamless to convert from one to the other.

json_path/src/json_node.rs
```diff
@@ -33,10 +33,10 @@ impl SelectValue for Value {
         }
     }
 
-    fn values<'a>(&'a self) -> Option<Box<dyn Iterator<Item = &'a Self> + 'a>> {
+    fn values<'a>(&'a self) -> Option<Box<dyn Iterator<Item = ValueRef<'a, Self>> + 'a>> {
         match self {
-            Self::Array(arr) => Some(Box::new(arr.iter())),
-            Self::Object(o) => Some(Box::new(o.values())),
+            Self::Array(arr) => Some(Box::new(arr.iter().map(|v| ValueRef::Borrowed(v)))),
```
ephraimfeldblum: Stylistically, would it be possible to make this point-free? ie `.map(ValueRef::Borrowed)` instead of `.map(|v| ValueRef::Borrowed(v))`. It's not a big deal either way.

redis_json/src/ivalue_manager.rs
```diff
@@ -36,13 +37,83 @@ pub struct IValueKeyHolderWrite<'a> {
...
+                ArrayTag::U16 => Some(PathValue::U16(array, index)),
+                ArrayTag::F16 => Some(PathValue::F16(array, index)),
+                ArrayTag::BF16 => Some(PathValue::BF16(array, index)),
+                ArrayTag::I32 => Some(PathValue::I32(array, index)),
+                ArrayTag::U32 => Some(PathValue::U32(array, index)),
+                ArrayTag::F32 => Some(PathValue::F32(array, index)),
+                ArrayTag::I64 => Some(PathValue::I64(array, index)),
+                ArrayTag::U64 => Some(PathValue::U64(array, index)),
+                ArrayTag::F64 => Some(PathValue::F64(array, index)),
+                ArrayTag::Heterogeneous => unreachable!(),
```
ephraimfeldblum: This unreachable should not be here. this `match` is itself capable of entirely replacing the outer `if`-`else`

redis_json/src/ivalue_manager.rs
```diff
@@ -36,13 +37,83 @@ pub struct IValueKeyHolderWrite<'a> {
...
+            }
+        }
+    }
+}
+
+fn follow_path(path: Vec<String>, root: &mut IValue) -> Option<PathValue<'_, '_>> {
+    path.into_iter().try_fold(
+        PathValue::IValue(root),
+        |target: PathValue<'_, '_>, token: String| {
```
ephraimfeldblum: Do we need to explicitly name the types here?

redis_json/src/ivalue_manager.rs
```diff
@@ -96,29 +179,225 @@ impl<'a> IValueKeyHolderWrite<'a> {
         F2: FnOnce(f64, f64) -> f64,
     {
         let in_value = &serde_json::from_str(num)?;
+        use half::{bf16, f16};
+        // TODO: macro for this
```
ephraimfeldblum: Yes.

redis_json/src/ivalue_manager.rs
```diff
@@ -171,17 +450,101 @@ impl<'a> WriteHolder<IValue, IValue> for IValueKeyHolderWrite<'a> {
             self.set_root(v)
         } else {
             let root = self.get_value()?.unwrap();
-            Ok(update(path, root, |val| Ok(*val = v.take())).is_ok())
+            Ok(update(path, root, |val| match val {
+                PathValue::IValue(val) => Ok(*val = v.take()),
+                PathValue::I8(iarray, index) => {
```
ephraimfeldblum: Macro for this as well.

json_path/src/json_path.rs
```diff
@@ -13,11 +13,76 @@ use pest::Parser;
...
+    }};
+}
+
+macro_rules! value_ref_get_key {
+    ($value_ref:expr, $curr:expr) => {{
+        match &$value_ref {
+            ValueRef::Borrowed(v) => v.get_key($curr),
+            ValueRef::Owned(v) => v
+                .get_key($curr)
+                .and_then(|v| Some(ValueRef::Owned(v.inner_cloned()))),
```
ephraimfeldblum: Why `.and_then(Some(...))` instead of simply `.map(...)`?

json_path/src/json_path.rs
```diff
@@ -13,11 +13,76 @@ use pest::Parser;
...
+    }};
+}
+
+macro_rules! value_ref_get_index {
+    ($value_ref:expr, $i:expr) => {{
+        match &$value_ref {
+            ValueRef::Borrowed(v) => v.get_index($i),
+            ValueRef::Owned(v) => v
+                .get_index($i)
+                .and_then(|v| Some(ValueRef::Owned(v.inner_cloned()))),
```
ephraimfeldblum: Why `.and_then(Some(...))` instead of simply `.map(...)`?

json_path/src/json_path.rs
```diff
@@ -482,32 +547,32 @@ pub struct PathCalculator<'i, UPTG: UserPathTrackerGenerator> {
...
+    root: ValueRef<'i, S>,
 }
 
 // The following block of code is used to create a unified iterator for arrays and objects.
 // This can be used in places where we need to iterate over both arrays and objects, create a path tracker from them.
 enum Item<'a, S: SelectValue> {
-    ArrayItem(usize, &'a S),
-    ObjectItem(&'a str, &'a S),
+    ArrayItem(usize, ValueRef<'a, S>),
+    ObjectItem(String, ValueRef<'a, S>),
```
ephraimfeldblum: Does the Key need to be owned? What was wrong with `&'a str`?

json_path/src/json_path.rs
```diff
@@ -860,7 +925,7 @@ impl<'i, UPTG: UserPathTrackerGenerator> PathCalculator<'i, UPTG> {
                 Some(term) => {
                     let mut calc_data = PathCalculatorData {
                         results: Vec::new(),
-                        root: json,
+                        root: json.clone(),
```
ephraimfeldblum, line 926: Can these be handled without potentially cloning the entire document? Consider adding a TODO here and in all the subsequent `.clone()`s in this file. There is quite a lot of cloning occurring and I'm not certain that all of it is desirable or necessary.

ephraimfeldblum (changes requested):

json_path/src/json_node.rs
```diff
@@ -156,10 +175,13 @@ impl SelectValue for IValue {
...
 
-    fn values<'a>(&'a self) -> Option<Box<dyn Iterator<Item = &'a Self> + 'a>> {
+    fn values<'a>(&'a self) -> Option<Box<dyn Iterator<Item = ValueRef<'a, Self>> + 'a>> {
         match self.destructure_ref() {
-            DestructuredRef::Array(arr) => Some(Box::new(arr.iter())),
-            DestructuredRef::Object(o) => Some(Box::new(o.values())),
+            DestructuredRef::Array(arr) => Some(Box::new(arr.iter().map(|item| match item {
+                ArrayIterItem::Borrowed(val) => ValueRef::Borrowed(val),
+                ArrayIterItem::Owned(val) => ValueRef::Owned(val),
+            }))),
```
ephraimfeldblum: this can be `Array(arr) => Some(Box::new(arr.iter().map(Into::into)))`

json_path/src/json_node.rs
```diff
@@ -71,16 +73,16 @@ impl SelectValue for Value {
         }
     }
 
-    fn get_key<'a>(&'a self, key: &str) -> Option<&'a Self> {
+    fn get_key<'a>(&'a self, key: &str) -> Option<ValueRef<'a, Self>> {
         match self {
-            Self::Object(o) => o.get(key),
+            Self::Object(o) => o.get(key).map(|v| ValueRef::Borrowed(v)),
```
ephraimfeldblum: this can be `Object(o) => o.get(key).map(ValueRef::Borrowed)`

json_path/src/json_node.rs
```diff
@@ -71,16 +73,16 @@ impl SelectValue for Value {
...
             _ => None,
         }
     }
 
-    fn get_index(&self, index: usize) -> Option<&Self> {
+    fn get_index<'a>(&'a self, index: usize) -> Option<ValueRef<'a, Self>> {
         match self {
-            Self::Array(arr) => arr.get(index),
+            Self::Array(arr) => arr.get(index).map(|v| ValueRef::Borrowed(v)),
```
ephraimfeldblum: this can be `Array(arr) => arr.get(index).map(ValueRef::Borrowed)`

json_path/src/json_node.rs
```diff
@@ -131,6 +133,23 @@ impl SelectValue for Value {
...
         }
     }
+
+    fn shallow_clone(&self) -> Self {
+        self.clone()
+    }
+
+    fn shallow_drop(&mut self) {
+        unsafe { std::ptr::drop_in_place(self as *mut Self) };
+    }
```
ephraimfeldblum: Fine, I guess. Since `Value` is more or less a historical artifact. But this is technically incorrect. ;p

json_path/src/json_node.rs
```diff
@@ -190,12 +214,14 @@ impl SelectValue for IValue {
         self.is_empty()
     }
 
-    fn get_key<'a>(&'a self, key: &str) -> Option<&'a Self> {
-        self.as_object().and_then(|o| o.get(key))
+    fn get_key<'a>(&'a self, key: &str) -> Option<ValueRef<'a, Self>> {
+        self.as_object()
+            .and_then(|o| o.get(key).map(|v| ValueRef::Borrowed(v)))
```
ephraimfeldblum: This can be `|o| o.get(key).map(ValueRef::Borrowed)`

json_path/src/json_node.rs
```diff
@@ -190,12 +214,14 @@ impl SelectValue for IValue {
...
     }
 
-    fn get_index(&self, index: usize) -> Option<&Self> {
-        self.as_array().and_then(|arr| arr.get(index))
+    fn get_index<'a>(&'a self, index: usize) -> Option<ValueRef<'a, Self>> {
+        self.as_array()
+            .and_then(|arr| arr.iter().nth(index).map(From::from))
```
ephraimfeldblum: In my opinion, I'd prefer to use `Into::into` than `From::from` when the type can be inferred. Into is typically more encompassing because From synthesizes Into, but the reverse is not true.

json_path/src/json_node.rs
```diff
@@ -228,4 +254,26 @@ impl SelectValue for IValue {
...
     }
+
+    fn shallow_clone(&self) -> Self {
+        match self.get_type() {
+            SelectValueType::Null
+            | SelectValueType::Bool
+            | SelectValueType::Long
+            | SelectValueType::Double => self.clone(),
+            SelectValueType::Array | SelectValueType::Object | SelectValueType::String => unsafe {
+                std::mem::transmute_copy(self)
```
ephraimfeldblum: Do we still need to `transmute_copy` here? We know the concrete type in this context.

AvivDavid23: Yes but I cannot really create the IValue myself since I will need to expose some private functions

redis_json/src/include/rejson_api.h
```diff
@@ -42,13 +42,12 @@ typedef struct RedisJSONAPI {
   RedisJSON (*openKeyFromStr)(RedisModuleCtx *ctx, const char *path);
 
   JSONResultsIterator (*get)(RedisJSON json, const char *path);
-  
+  int (*getAt)(RedisJSON json, size_t index, RedisJSON value);
```
oshadmi: @AvivDavid23 Why not move it to be with the V6 entries? (would be clearer that it is a new API in V6)

gabsow:
Whats our status?
Can we merge this PR ?

json_path/src/select_value.rs
```diff
@@ -21,16 +21,62 @@ pub enum SelectValueType {
...
+    fn deref(&self) -> &Self::Target {
+        self.as_ref()
+    }
+}
+
+impl<'a, T: SelectValue> PartialEq<&T> for ValueRef<'a, T> {
+    fn eq(&self, other: &&T) -> bool {
+        self.as_ref() == *other
+    }
+}
```
ephraimfeldblum, line 68: I think this is comparing refs, not objects. See the implementation of `KeyValue::is_equal` for how to compare equality of `SelectValue` objects.
To be honest, the fact that function is in KeyValue and not SelectValue is a mistake. It should be brought here.

redis_json/src/include/rejson_api.h
```diff
@@ -124,10 +120,22 @@ typedef struct RedisJSONAPI {
...
+  // V6 entries //
+  ////////////////
+  RedisJSONPtr (*allocJson)();
+  // The caller must pass 'ptr' which was allocated with allocJson
+  int (*getAt)(RedisJSON json, size_t index, RedisJSONPtr ptr);
+  // Get the next key-value pair
+  // The caller gains ownership of `key_name`
+  // The caller must pass 'ptr' which was allocated with allocJson
+  int (*nextKeyValue)(JSONKeyValuesIterator iter, RedisModuleString **key_name, RedisJSONPtr ptr);
+
```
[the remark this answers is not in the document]
AvivDavid23: you dont say

redis_json/src/c_api.rs
```diff
@@ -653,7 +781,7 @@ macro_rules! redis_json_module_export_shared_api {
                     std::ptr::null_mut(),
                 ));
 
-                for v in 1..6 {
+                for v in 1..7 {
```
GuyAv46, line 784: We should not register API versions we can't supply
```suggestion
                for v in 6..=6 {
```

ephraimfeldblum: i don't understand. we're not supplying v7 nor are we claiming to. ranges are defined to be half open. `1..7` has 6 elements; `{1, 2, 3, 4, 5, 6}`

GuyAv46: The `=6` is a nitpick, I think it's clearer than `7`
But I'm talking about the minimal API - the module cannot supply v1-v5 anymore.
Currently, a module can register to any of 1-5 and crash the first time it uses it. If you only expose v6 now, outdated modules will simply not link to JSON and behave as if it's missing.
The range in this for loop should only include 1 element - `{6}`

PR #1425, opened 2025-10-20
Update to the latest `ijson`

chesedo:
This version of `ijson` does not need nightly Rust to be compiled because it removes this use of `#[feature]` which only exists in the `d226058` revision. Therefore this updates the dependency away from the `d226058` to the latest one.

https://github.com/RedisJSON/ijson/pull/10/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759L21


chesedo:
Closed by #1429

PR #1429, opened 2025-10-22
RED-173025 stable channel with error handling

ephraimfeldblum:


redis_json/src/commands.rs
```diff
@@ -1347,18 +1347,16 @@ where
...
-            res = Some(redis_key.arr_insert(p, &args, index)?);
-        }
-        redis_key.notify_keyspace_event(ctx, "json.arrinsert")?;
-        manager.apply_changes(ctx);
-        Ok(res.unwrap().into())
+        ));
     }
+    let res = paths
+        .into_iter()
+        .try_fold(0, |_, p| redis_key.arr_insert(p, &args, index))?;
```
AvivDavid23, line 1356: why not `try_for_each`?

ephraimfeldblum: because `try_for_each` expects `F: FnMut(Item) -> R, R: Try<Output = ()>`, and here we have `R: Try<Output = usize>`

PR #1464, opened 2025-11-25
RED-177836 Adjust test to the eventual consistency nature of oss cluster configs

galcohen-redislabs:


gabsow:
I see ci is still failing

PR #1473, opened 2025-12-09
MOD-8137 Fix adding escapes to already-escaped characters

### LINKED ISSUE #1298: Inconsistent Path Evaluation Between JSON Commands
TLDR: Setting a value for a key of a single backslash using JSON.SET with a path seems impossible

# Inconsistent Behavior Seemingly Making It Impossible

I encountered inconsistent path evaluation across JSON commands:

```sh
127.0.0.1:6379> JSON.SET mydoc $ '{}'
OK
127.0.0.1:6379> JSON.SET mydoc '$["\\"]' '"x"'
OK
127.0.0.1:6379> JSON.GET mydoc '$["\\"]'
"[]"
127.0.0.1:6379> JSON.DEL mydoc '$["\\"]'
(integer) 0
127.0.0.1:6379> JSON.GET mydoc
"{\"\\\\\\\\\":\"x\"}"
127.0.0.1:6379> JSON.OBJKEYS mydoc
1) "\\\\"
```

Is setting a key of a single backslash using a path possible with the set command? 

I have been able to SET and GET a key of a single backslash like so:

```sh
127.0.0.1:6379> JSON.SET mydoc '$' '{"\\": "x"}'
OK
127.0.0.1:6379> JSON.OBJKEYS mydoc
1) "\\"
127.0.0.1:6379> JSON.GET mydoc
"{\"\\\\\":\"x\"}"
127.0.0.1:6379> JSON.GET mydoc '$["\\"]'
"[\"x\"]"
```

# Motivation

I'm trying to store and patch some valid JSON documents that include backslashes as parts of JSON keys inside of RedisJSON. It seems like doing so might not be possible. Is there something I'm missing?

ephraimfeldblum:
[MOD-8137](https://redislabs.atlassian.net/browse/MOD-8137)
fixes #1298

[MOD-8137]: https://redislabs.atlassian.net/browse/MOD-8137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/json_path/src/json_path.rs
+++ b/json_path/src/json_path.rs
@@ -569,10 +569,7 @@ impl<'a, S: SelectValue> Item<'a, S> {
         }
     }
 
-    fn create_tracker<'i, 'j>(&'i self, parent: &'j PathTracker<'i, 'j>) -> PathTracker<'i, 'j>
-    where
-        'a: 'i,
-    {
+    fn create_tracker<'i>(&'a self, parent: &'i PathTracker<'a, 'i>) -> PathTracker<'a, 'i> {
         match self {
             Item::ArrayItem(index, _) => create_index_tracker(*index, parent),
             Item::ObjectItem(key, _) => create_str_tracker(key.clone(), parent),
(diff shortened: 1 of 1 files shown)


json_path/src/json_path.rs
```diff
@@ -744,41 +688,24 @@ impl<'i, UPTG: UserPathTrackerGenerator> PathCalculator<'i, UPTG> {
...
-                    }
-                    Rule::string_value_escape_2 => {
-                        value_ref_get_key!(json, &(s.replace("\\\\", "\\").replace("\\'", "'")))
-                    }
-                    _ => panic!("{c:?}"),
-                };
-                if let Some(e) = curr_val {
-                    self.calc_internal(pairs.clone(), e, None, calc_data);
+                Rule::string_value_escape_2 => {
+                    Cow::Owned(s.replace("\\\\", "\\").replace("\\'", "'"))
```
galcohen-redislabs, line 699: What about other sanitizing other special chars? (e.g., `\n`, `\r`, `\t`, `\uXXXX`)

ephraimfeldblum: not needed.

we do not sanitize them because we do not sanitize in general. it's only quotes that need sanitization because only quotes required the extra `\\`s to be correctly parsed and passed to JSON in the first place.

parsing arbitrary `\uXXXX` will invariably create bugs from illegal utf-8 chars.

galcohen-redislabs (approved):

PR #1530, opened 2026-03-31
MOD-14632 Eliminate panics across codebase

AvivDavid23:
Trying to eliminate as much a possible from all the panic and unwanted `unwraps` in. the codebase.

- Currently, for `SelectValue` trait getters, returning an `Option` to align we current solution of `get_key`, `get_values`, etc..
- Add cargo clippy to the CI

---

> [!NOTE]
> **Medium Risk**
> Moderate risk due to wide-reaching signature changes (`SelectValue` getters now return `Option`) and many unwrap/panic removals across JSONPath evaluation and RedisJSON command/C-API paths, which could subtly change error/NULL behavior.
> 
> **Overview**
> Adds a new GitHub Actions `clippy` workflow and wires it into PR CI, running `cargo clippy` with warnings (and `panic`/`todo`/`unimplemented`) treated as failures.
> 
> Refactors the JSONPath/RedisJSON Rust code to eliminate many `panic!`/`unwrap()` paths by making `SelectValue` primitive getters return `Option` and updating JSONPath evaluation, C API helpers, serialization (`KeyValue::serialize_object` now returns `RedisResult<String>`), RDB backward loading, and command implementations to propagate errors or return `NULL`/`Err` defensively. Adds targeted tests to ensure `calc_once_paths`/`calc_paths` maintain the invariant that tracked paths align 1:1 with value matches.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 91c155dd46f5746170e957dfcafb892f76b0c970. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/json_path/src/json_node.rs
+++ b/json_path/src/json_node.rs
@@ -100,39 +100,41 @@ impl SelectValue for Value {
         }
     }
 
-    fn get_str(&self) -> String {
+    fn get_str(&self) -> Option<String> {
         match self {
-            Self::String(s) => s.to_string(),
-            _ => panic!("not a string"),
+            Self::String(s) => Some(s.to_string()),
+            _ => None,
         }
     }
 
-    fn as_str(&self) -> &str {
+    fn as_str(&self) -> Option<&str> {
         match self {
-            Self::String(s) => s.as_str(),
-            _ => panic!("not a string"),
+            Self::String(s) => Some(s.as_str()),
+            _ => None,
         }
     }
 
-    fn get_bool(&self) -> bool {
+    fn get_bool(&self) -> Option<bool> {
         match self {
-            Self::Bool(b) => *b,
-            _ => panic!("not a bool"),
+            Self::Bool(b) => Some(*b),
+            _ => None,
         }
     }
 
-    fn get_long(&self) -> i64 {
+    fn get_long(&self) -> Option<i64> {
         match self {
-            Self::Number(n) if n.is_i64() => n.as_i64().unwrap(),
-            _ => panic!("not a long"),
+            Self::Number(n) if n.is_i64() => n.as_i64(),
+            Self::Number(_) => None,
+            _ => None,
         }
     }
 
-    fn get_double(&self) -> f64 {
+    fn get_double(&self) -> Option<f64> {
         match self {
-            Self::Number(n) if n.is_f64() => n.as_f64().unwrap(),
-            Self::Number(n) if n.is_u64() => n.as_u64().unwrap() as _,
-            _ => panic!("not a double"),
+            Self::Number(n) if n.is_f64() => n.as_f64(),
+            Self::Number(n) if n.is_u64() => n.as_u64().map(|u| u as f64),
+            Self::Number(_) => None,
+            _ => None,
         }
     }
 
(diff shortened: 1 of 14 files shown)


json_path/src/json_node.rs
TalBarYakar, line 1: Please double-check that every SelectValue call site handles None from :
* get_str 
* as_str 
* get_bool 
* get_long 
* get_double

json_path/src/lib.rs
TalBarYakar, line 1: Please validate find_paths / find_add_paths still behave correctly when the returned path list is shorter or empty.

json_path/src/select_value.rs
```diff
@@ -115,18 +116,26 @@ pub trait SelectValue: Debug + Eq + PartialEq + Default + Clone + Serialize {
             SelectValueType::Array => {
```
TalBarYakar, line 116: add comment about values() is none treat as no children (depth 0)...

json_path/src/select_value.rs
```diff
@@ -115,18 +116,26 @@ pub trait SelectValue: Debug + Eq + PartialEq + Default + Clone + Serialize {
...
                 1 + self
                     .values()
-                    .unwrap()
-                    .map(|v| v.calculate_value_depth())
-                    .max()
-                    .unwrap_or_default()
+                    .map(|vals| vals.map(|v| v.calculate_value_depth()).max().unwrap_or(0))
+                    .unwrap_or(0)
             }
             SelectValueType::Object => {
```
TalBarYakar, line 122: I’d rather avoid debug-only behavior (debug_assert!, etc.) so debug and release match. Can we use the same Option/fallback (or tests) everywhere instead?

json_path/src/select_value.rs
```diff
@@ -136,26 +145,54 @@ pub fn is_equal<T1: SelectValue, T2: SelectValue>(a: &T1, b: &T2) -> bool {
...
-            SelectValueType::Long => a.get_long() == b.get_long(),
-            SelectValueType::Double => a.get_double() == b.get_double(),
-            SelectValueType::String => a.get_str() == b.get_str(),
-            SelectValueType::Array => {
-                a.len().unwrap() == b.len().unwrap()
-                    && a.values()
-                        .unwrap()
-                        .zip(b.values().unwrap())
-                        .all(|(a, b)| is_equal(a.as_ref(), b.as_ref()))
+            SelectValueType::Bool => {
```
TalBarYakar: don't you think unreachable is more suitable in here ?
`SelectValueType::Bool => match (a.get_bool(), b.get_bool()) {
    (Some(x), Some(y)) => x == y,
    _ => unreachable!("get_type/getter mismatch in is_equal for Bool"),
},`

In my experience, having different behavior in debug and release modes causes issues and makes things more complicated to analyze

TalBarYakar (commented):
The comments apply to the entire section. Let me know what you think

json_path/src/lib.rs
AvivDavid23: Can you elaborate more? These function have not changed

json_path/src/select_value.rs
AvivDavid23: will try

json_path/src/select_value.rs
AvivDavid23: I really think that's not necessary, it's well shown in the code

json_path/src/select_value.rs
AvivDavid23: In general, you are right. In theory, get_x api might fail(currently returns None for simplicity).
That's why I can't say it is really unreachable

json_path/src/lib.rs
TalBarYakar: The find_paths / find_add_paths implementations themselves didn’t change, but they both call calc_once_paths (json_path::calc_once_paths), and that function did change: it used to do path_tracker.unwrap() and now uses filter_map so entries with path_tracker == None are dropped instead of panicking.

json_path/src/select_value.rs
TalBarYakar: Let’s align on one approach:

If None from get_*() is allowed (even when get_type() matches), remove debug_assert! and keep the same zip / false logic in debug and release—no panic! either.

If None means a broken SelectValue impl, debug_assert! isn’t enough (it’s stripped in release). Use panic! / unreachable! in all builds, or enforce with tests—not debug-only checks.

@AvivDavid23 WDY ?

TalBarYakar (commented):
.

json_path/src/select_value.rs
AvivDavid23: sounds good, will go with the first option where possible

json_path/src/json_node.rs
```diff
@@ -24,6 +24,8 @@ impl SelectValue for Value {
             Self::Object(_) => SelectValueType::Object,
             Self::Number(n) if n.is_i64() => SelectValueType::Long,
             Self::Number(n) if n.is_f64() | n.is_u64() => SelectValueType::Double,
+            #[allow(clippy::panic)]
```
TalBarYakar: Switched the Number catch-all to unreachable! instead of panic! so it’s explicit that this arm exists only for match exhaustiveness and should never run if serde_json::Number stays i64/u64/f64—unreachable! documents that contract better than a generic panic.

json_path/src/json_path.rs
```diff
@@ -756,38 +800,60 @@ impl<'i, UPTG: UserPathTrackerGenerator> PathCalculator<'i, UPTG> {
         if json.get_type() != SelectValueType::Array {
             return;
         }
-        let n = json.len().unwrap();
```
TalBarYakar, line 759: for the new early-return paths (e.g. calc_range), did you verify callers treat “no recursive calc_internal / no pushes” correctly (empty vs Invalid subpath), and that nothing relied on unwrap for impossible states?

TalBarYakar (commented):
.

json_path/src/json_path.rs
AvivDavid23: yes. calc_range just calculate the next expression. if it cant calculate, no results will be appended. And of course its better than panicking

TalBarYakar (approved):

Issue #17, opened 2017-05-03
master-slave mode data inconsistency 


leonchen83:
run below command in master  
and data inconsistency in slave  

```java  
JSON.NUMINCRBY
JSON.ARRAPPEND
JSON.ARRINSERT
JSON.ARRTRIM
JSON.STRAPPEND
JSON.ARRPOP
```
but there are 2 commands data consistency in slave

```java  
JSON.SET
JSON.DEL
```

env

```
redis : redis-4.0-rc3
loaded module : both master and slave loaded rejson module
kernal : Linux version 4.8.13-100.fc23.x86_64 (mockbuild@bkernel02.phx2.fedoraproject.org) (gcc version 5.3.1 20160406 (Red Hat 5.3.1-6) (GCC) )
```

itamarhaber:
Thanks for reporting - I have a good notion of why this could be happening (i.e. my bug, my bad).

leonchen83:
seems these commands miss `RedisModule_ReplicateVerbatim` or `RedisModule_Replicate`

itamarhaber:
That's exactly the reason - I'd appreciate if you can verify that 8e11106 fixes this issue.

leonchen83:
verified. could close this issue.

itamarhaber:
@leonchen83 thanks!

It would be good to add master/slave (and perhaps cluster too) to https://github.com/RedisLabs/rmtest - @dvirsky wdut?

PR #1595, opened 2026-05-17
MOD-15731 - modules modernization  alignment

TalBarYakar:
> [!NOTE]
> **Medium Risk**
> Large cross-platform CI and dependency-install refactor; failures would block builds/tests rather than changing runtime RedisJSON behavior, but coverage across many OS images increases regression risk.
> 
> **Overview**
> This PR **modernizes build and CI bootstrap** (MOD-15731) by replacing scattered per-distro `.install/*.sh` scripts and `common_installations.sh` with a single pipeline: **`install_script.sh`** detects **OSNICK**, installs OS packages via **`lib/pm.sh`** + **`os/<osnick>.sh`**, then **`lib/setup-python.sh`** (uv, repo **`venv/`**, pinned pip/RLTest deps), and **`getrust.sh`** on non-Alpine hosts.
> 
> **`make bootstrap`** now runs only that installer (no Readies include in the same invocation) and wipes **`venv`** first. CI composite actions (**`run-tests`**, **`san-run-tests`**) **activate the pre-built `venv`** instead of recreating it or calling **`common_installations.sh`**; sanitizer flow drops redundant checkouts/getrust steps and documents the install contract. **Dockerfiles** delegate system deps to **`install_script.sh`** and fix **`requirements_docker.txt`** naming; several Debian/Ubuntu images add apt retry/timeouts.
> 
> **`getrust.sh`** symlinks Rust tools into **`/usr/local/bin`** so **`cargo`** works in the same shell as bootstrap. **`tests/pytest/tests.sh`** picks a Python with RLTest (prefers **`$ROOT/venv`**, supports **`RLTEST_PYTHON`**). **`module.conf`** adds a reserved **redis-gen-conf private** block; **`.gitignore`** ignores **`.prepared`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit caea8412af87c3e02d8482c24fd88d731d6b45d1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.install/lib/detect-osnick.sh
```diff
@@ -0,0 +1,61 @@
...
+# Map the host to a canonical OSNICK that names a file in ../os/.
+#
+# Sourced by install_script.sh. Provides one function: detect_osnick.
+# Result is printed on stdout; empty string on unrecognised hosts.
+#
+# Caller may pre-set $OSNICK in the environment (e.g. Docker ARG/ENV) to
+# bypass detection entirely; we still normalise rhel*->rocky* so callers
+# don't have to remember which file we picked.
+
+detect_osnick() {
```
AvivDavid23, line 11: so we dont rely on readis anymore, right?

module.conf
```diff
@@ -1,2 +1,11 @@
 ############################## JSON ENGINE CONFIG ############################
 
+# >>> BEGIN redis-gen-conf:private <<<
+# Internal/advanced RedisJSON tunables — intentionally NOT exported to
+# redis-gen.conf by `make sync-redis-conf`. Set via CLI override or by
+# editing this module.conf directly when running redis-server against it.
+# See modules/sync-redis-conf.mk for the marker contract.
+
+
+# <<< END redis-gen-conf:private <<<
```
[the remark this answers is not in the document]
AvivDavid23: @TalBarYakar is it resolved?

.install/lib/detect-osnick.sh
TalBarYakar: We do , in our CI. 
But When you run bootstrap locally, it’s automatically detected by the script itself.

AvivDavid23 (dismissed):

.install/os/amazonlinux2.sh
```diff
@@ -0,0 +1,23 @@
...
+rhel_default_install
+
+# Amazon Linux 2's base `openssl` is 1.0.2 — too old for Redis/Rust. Layer
+# openssl 1.1 from amzn2 and symlink /usr/bin/openssl so the build picks it up.
+$SUDO yum -y install openssl11 openssl11-devel
+if command -v openssl11 >/dev/null 2>&1; then
+    $SUDO ln -sf "$(command -v openssl11)" /usr/bin/openssl
+fi
+
+$SUDO ln -sf "$(command -v cmake3)" /usr/bin/cmake
```
gabsow: This can blow up under `set -e`. `cmake3` is installed via `yum_install` a few lines above, which uses `--skip-broken`; if it fails to resolve, `command -v cmake3` returns empty and you end up running `ln -sf "" /usr/bin/cmake`, which aborts the whole script. Wrap it in the same guard used for `openssl11` just above:

```bash
if command -v cmake3 >/dev/null 2>&1; then
    $SUDO ln -sf "$(command -v cmake3)" /usr/bin/cmake
fi
```

gabsow (commented):
Nice refactor — the single-source-of-truth library and per-OSNICK dispatcher cut a lot of duplication. Comments below are mostly small bugs and inconsistencies; a couple (amazonlinux2 cmake3 guard, the misleading safe.directory comment in install_script.sh) are worth fixing before merge.

.install/install_script.sh
```diff
@@ -1,22 +1,69 @@
...
-source ${OS}.sh $MODE
+echo "==> [redisjson] OSNICK=$OSNICK PM=$PM"
+
+# shellcheck disable=SC1090
+. "$osfile"
+
+# Allow git operations on the checked-out source even when its uid doesn't
+# match the current user (common in CI containers). Scoped to this repo
+# (--local), not the host's global git config.
+git config --global --add safe.directory '*' || true
```
gabsow, line 50: The comment above this block says "Scoped to this repo (`--local`), not the host's global git config" but the actual command is `--global` with wildcard `*` — that's the broadest possible scope, not the narrowest. The follow-up `git -C "$ROOT" config --local --add safe.directory '*'` is also redundant since the global `*` already matches everything. Either trim down to just the local config (safer, matches the comment) or fix the comment to reflect the code.

.install/os/azurelinux3.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+# Azure Linux 3 (tdnf).
+
+. "$LIB/packages.sh"
+
+tdnf_default_install
+git config --global --add safe.directory '*' || true
```
gabsow: Duplicate of `install_script.sh`'s `git config --global --add safe.directory '*'`. Same line is also in `os/mariner2.sh`. Remove from the OS scripts and keep just the one in `install_script.sh`.

.install/os/azurelinux3.sh
```diff
@@ -0,0 +1,10 @@
...
+
+# Install AWS CLI for uploading to S3
+pip3 install awscli --upgrade
```
gabsow: This runs against the host's `pip3` outside any venv. On modern distros with PEP 668 markers, this fails with `error: externally-managed-environment`. Prefer the tdnf package (`tdnf install -y awscli` if available) or the curl-based installer pattern used in `os/mariner2.sh`. If you keep pip, at least pass `--break-system-packages`.

.install/os/mariner2.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+# CBL-Mariner 2.0 (tdnf).
+
+. "$LIB/packages.sh"
+
+tdnf_default_install
+git config --global --add safe.directory '*' || true
+
+# Install aws-cli for uploading artifacts to s3
+curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
```
gabsow: AWS CLI download pollutes the caller's CWD. The old `common_base_linux_mariner_2.0.sh` did `cd /tmp` first; since this file is now sourced from `install_script.sh` with CWD at `.install/`, the zip + `./aws/` directory end up in the repo. Wrap in a subshell:

```bash
(cd /tmp && curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o awscliv2.zip && unzip -q awscliv2.zip && ./aws/install)
```

.install/os/mariner2.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+# CBL-Mariner 2.0 (tdnf).
+
+. "$LIB/packages.sh"
+
+tdnf_default_install
+git config --global --add safe.directory '*' || true
```
gabsow: Redundant with the same line in `install_script.sh` — remove.

tests/pytest/tests.sh
```diff
@@ -18,6 +18,29 @@ SAN_REDIS_SUFFIX=7.4
...
+		echo "$ROOT/venv/bin/python"
+		return
+	fi
+	for p in python3.12 python3.11 python3.10 python3.9 python3; do
+		if command -v "$p" &>/dev/null && "$p" -c 'import RLTest' &>/dev/null; then
+			command -v "$p"
+			return
+		fi
+	done
+	echo python3
```
gabsow: Final fallback is `echo python3`, but you've already established at this point that none of `python3.12/.11/.10/.9` had RLTest installed. If the base `python3` doesn't either, this returns a python that will fail downstream with `ModuleNotFoundError: No module named 'RLTest'`. Since you've done the lookup, emit a clearer error here instead — saves a debugging cycle later.

.install/lib/detect-osnick.sh
```diff
@@ -0,0 +1,61 @@
...
+
+    if [ -r /etc/os-release ]; then
+        # shellcheck disable=SC1091
+        . /etc/os-release
+        case "${ID:-}:${VERSION_ID:-}" in
+            ubuntu:18.04)             printf 'bionic\n'          ;;
+            ubuntu:20.04)             printf 'focal\n'           ;;
+            ubuntu:22.04)             printf 'jammy\n'           ;;
+            ubuntu:24.04)             printf 'noble\n'           ;;
+            ubuntu:25.*|ubuntu:26.*)  printf 'resolute\n'        ;;
```
gabsow: `ubuntu:25.*` maps to `resolute`, but `os/resolute.sh`'s header says "Ubuntu 26.04 (resolute)". Ubuntu 25.04 / 25.10 are interim releases — not resolute. Either rename the OSNICK or split into `questing`/`oracular`/etc.

.install/lib/detect-osnick.sh
```diff
@@ -0,0 +1,61 @@
...
+            debian:11*)               printf 'bullseye\n'        ;;
+            debian:12*)               printf 'bookworm\n'        ;;
+            debian:13*|debian:trixie) printf 'trixie\n'          ;;
```
gabsow: The `debian:trixie` arm is unreachable — Debian's `VERSION_ID` is always numeric (e.g. "13"), never the codename. If you want the codename match as a safety net (e.g. for pre-release sid), check `${VERSION_CODENAME:-}` separately.

.github/actions/run-tests/action.yml
```diff
@@ -9,17 +9,25 @@ inputs:
 runs:
   using: composite
   steps:
+    # Caller contract: `.install/install_script.sh` MUST have already run in this
```
gabsow, line 12: FYI: `grep -rn 'actions/run-tests' .github/workflows/` returns nothing — only `san-run-tests` is referenced (in `flow-sanitizer.yml`). This action appears to be orphaned. If that's intentional (kept for future use), please ignore; otherwise, removing it would simplify the PR. The new caller contract here only matters if anything actually calls this action.

.install/lib/pm.sh
```diff
@@ -0,0 +1,184 @@
...
+
+# tdnf has no --skip-broken and aborts the whole transaction if any single
+# package is missing from the repos (CBL-Mariner / AzureLinux ship a much
+# smaller subset than dnf). Install one at a time so a missing package
+# becomes a warning rather than a build failure.
+tdnf_install() {
+    [ "$#" -gt 0 ] || return 0
+    local pkg
+    for pkg in "$@"; do
+        if ! $SUDO tdnf -y install "$pkg" >/dev/null 2>&1; then
```
gabsow: Swallowing all of tdnf's stderr makes Mariner/AzureLinux failures painful to debug — the warning loses the actual error message. Consider letting stderr through, or capturing it and emitting it when a package fails:

```bash
local out
if ! out=$($SUDO tdnf -y install "$pkg" 2>&1); then
    echo "pm.sh: tdnf skipped unavailable package '$pkg': $(echo "$out" | tail -1)"
fi
```

.install/lib/setup-python.sh
```diff
@@ -0,0 +1,58 @@
...
+
+if ! command -v uv >/dev/null 2>&1; then
+    echo "==> [redisjson] installing uv"
+    curl -LsSf https://astral.sh/uv/install.sh | sh
+    export PATH="$HOME/.local/bin:$HOME/.cargo/bin:$PATH"
+fi
+
+if ! command -v uv >/dev/null 2>&1; then
+    echo "setup-python.sh: WARNING: uv installation failed; skipping venv setup" >&2
+    return 0 2>/dev/null || exit 0
```
gabsow: Silently returning leaves no venv, and downstream then fails with the unrelated `venv missing` error in `actions/run-tests/action.yml`. The original failure (uv install) gets buried. Prefer to fail fast here so the actual root cause is at the top of the failed CI step.

.install/lib/setup-python.sh
```diff
@@ -0,0 +1,58 @@
...
+fi
+
+# All pip work goes through `uv pip --python <venv>` (never --system, never
+# under sudo). Sourcing under sudo would otherwise resolve uv against /usr's
+# python3 (3.6 on EL8) and break rltest.
+uv_pip() {
+    uv pip install --python "$ROOT/venv/bin/python" "$@"
+}
+
+uv_pip --upgrade pip wheel "setuptools<81"
```
gabsow, line 63: `setuptools<81` pinned without a comment. A one-liner explaining the constraint (which package broke, when this can be relaxed) would prevent this from rotting into a mystery pin in six months.

.install/getrust.sh
```diff
@@ -29,6 +29,17 @@ $MODE rustup component add rust-src
...
 
+# Symlink the toolchain into /usr/local/bin so it's on every shell's PATH —
+# not just bash-login shells that source ~/.cargo/env. This is what makes
+# `make build` work in the same shell session that just ran `make bootstrap`
+# (and also covers direct `make -C modules/<name>` calls, CI runners, and
+# any future script that shells out to cargo/rustc). $MODE handles
+# sudo-vs-no-sudo identically to the rest of the script; if /usr/local/bin
+# isn't writable we still fall back to the ~/.cargo/env bashrc hook above.
+for bin in rustc cargo rustup rustfmt cargo-fmt clippy-driver cargo-clippy; do
+    [ -e "$HOME/.cargo/bin/$bin" ] && $MODE ln -sf "$HOME/.cargo/bin/$bin" "/usr/local/bin/$bin" 2>/dev/null || true
```
gabsow: Subtle precedence issue with `&& A || true`: the `|| true` triggers whenever the guard `[ -e ... ]` is false *or* when `ln -sf` fails. It works here because both branches are no-ops, but shellcheck would flag SC2015. The clearer form:

```bash
if [ -e "$HOME/.cargo/bin/$bin" ]; then
    $MODE ln -sf "$HOME/.cargo/bin/$bin" "/usr/local/bin/$bin" 2>/dev/null || true
fi
```

Also, the surrounding comment claims a fallback to the `~/.cargo/env` bashrc hook, but nothing here verifies that anything was actually symlinked — so it's not really a fallback path, just a best-effort attempt.

.install/os/alma8.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+# AlmaLinux 8. EL8 + gcc-toolset-11; clang-format; legacy pip dataclasses shim.
+
+. "$LIB/packages.sh"
+
+el8_default_install
+dnf_install clang-tools-extra
+
+if command -v python3 >/dev/null 2>&1 && ! python3 -c 'import dataclasses' 2>/dev/null; then
+    python3 -m pip install --disable-pip-version-check -q 'dataclasses>=0.8,<1' || true
```
gabsow, line 18: Worth a comment on what consumes this shim. `el8_default_install` installs python3.11 (which has `dataclasses` built-in) and `setup-python.sh` creates the venv against 3.11. The `python3` resolved here is the base EL8 3.6 — so the shim only matters if something later runs against system python3.6 (not the venv). If nothing does, this is dead code; if something does, the comment should say what.

(Same applies to `os/rocky8.sh:10`.)

.github/workflows/flow-macos.yml
```diff
@@ -99,7 +99,9 @@ jobs:
             . "${GITHUB_WORKSPACE}/venv/bin/activate"
           echo ::endgroup::
           echo ::group::Install python dependencies
-            ./.install/common_installations.sh
+            pip install --upgrade pip wheel "setuptools<81"
```
gabsow, line 102: Inconsistent with the rest of the PR's "one entrypoint" design — every other distro now bootstraps via `install_script.sh` (which delegates to `lib/setup-python.sh` for the uv venv + pip work), but macOS still does three explicit `pip install` lines. Consider switching this to `./.install/install_script.sh` so macOS goes through `os/macos.sh` + `setup-python.sh` like the rest.

module.conf
```diff
@@ -1,2 +1,11 @@
 ############################## JSON ENGINE CONFIG ############################
 
+# >>> BEGIN redis-gen-conf:private <<<
+# Internal/advanced RedisJSON tunables — intentionally NOT exported to
+# redis-gen.conf by `make sync-redis-conf`. Set via CLI override or by
+# editing this module.conf directly when running redis-server against it.
+# See modules/sync-redis-conf.mk for the marker contract.
```
gabsow: The markers reference `modules/sync-redis-conf.mk`, but `find . -name sync-redis-conf.mk` returns nothing — the consumer isn't in this repo (yet). Land the consumer in the same PR, or drop the empty markers until the contract is in place. Adding unused contract markers tends to confuse future readers.

Dockerfile.amazonlinux2023
```diff
@@ -2,25 +2,25 @@ FROM amazonlinux:2023
...
-    rsync unzip clang tar
+# Install build/test deps via the per-OSNICK installer (.install/install_script.sh
+# dispatches to .install/os/<osnick>.sh).
+COPY rust-toolchain.toml /workspace/rust-toolchain.toml
+COPY .install /workspace/.install
+RUN bash /workspace/.install/install_script.sh && dnf clean all
+# rustc/cargo from rustup (install_script.sh ran getrust.sh)
+ENV PATH="/root/.cargo/bin:${PATH}"
 
 # Install Redis (SANITIZER can be 'address' for ASAN builds)
```
gabsow, line 13: Nit: the `# Install Redis (SANITIZER…)` comment now floats above two `ARG` lines with a blank line between it and its associated `RUN` (line 17). Cosmetic — either move the comment down to line 17 or drop it.

module.conf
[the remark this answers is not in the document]
TalBarYakar: Won't fix — intentional marker format, parser matches by exact string.

.install/os/amazonlinux2.sh
TalBarYakar: Fixed — wrapped the cmake3 symlink in the same command -v guard used for openssl11 above. Matches the existing pattern in the file.

.install/install_script.sh
TalBarYakar: Fixed — dropped the redundant --local block and rewrote the comment to honestly describe the global scope. Kept the global line because git config --local itself fires the safe.directory check, so we need a global rule in place before any per-repo command can run.

.install/os/azurelinux3.sh
TalBarYakar: removed.

.install/os/azurelinux3.sh
TalBarYakar: Fixed — aligned azurelinux3.sh to match timeseries master's pattern (curl-based AWS CLI v2 installer instead of pip3 install awscli, which fails under PEP 668 on Azure Linux 3). File is now byte-identical to timeseries master, so the three tdnf-based modules stay consistent.

.install/os/mariner2.sh
TalBarYakar: Fixed — wrapped in a (cd /tmp && …) subshell exactly as you suggested

tests/pytest/tests.sh
TalBarYakar: Fixed — replaced the silent echo python3 fallback with a clear stderr error and exit 1. The message points at the two recovery paths (install RLTest or set RLTEST_PYTHON) so the operator doesn't have to retrace the lookup.

.install/os/mariner2.sh
TalBarYakar: done

.install/lib/detect-osnick.sh
TalBarYakar: Fixed — dropped ubuntu:25.* so 25.04 (plucky) / 25.10 (questing) fall through to the unsupported-OSNICK error instead of silently picking up resolute's (Ubuntu 26.04 LTS) package set. We don't have os/plucky.sh / os/questing.sh and interim Ubuntu releases aren't supported

.install/lib/detect-osnick.sh
TalBarYakar: Fixed — dropped the unreachable |debian:trixie alternation arm. Will mirror this and the #15 fix into bloom + timeseries master in follow-up PRs so the three repos stay aligned; the proper readies-style rewrite (read VERSION_CODENAME from /etc/os-release directly) we'll do as a separate cleanup that eliminates this class of bug in one shot.

.github/actions/run-tests/action.yml
TalBarYakar: The orphan predates this PR — actions/run-tests has 0 callers on master too (along with 5 other unused composite actions: build-json-module-and-redis*, make-pack, node20-supported, pack-module). What this PR changes is only the action body — it now assumes the new install_script.sh-managed venv instead of creating one inline. Wiring run-tests (and the others) up to a workflow is a real cleanup, but it spans multiple files (flow-linux, flow-macos, flow-memory-*) and is independent of the bootstrap refactor this PR is doing. I'd rather not expand scope here; happy to file a follow-up that rewires all six in one go.

.install/lib/pm.sh
TalBarYakar: Fixed — applied your pattern. Combined stderr+stdout is captured into $out; on failure, the warning now includes the last line of tdnf's output so a real network/GPG/conflict failure is distinguishable from the boring "package not in repo" case without changing the tolerate-the-miss behavior.

.install/lib/setup-python.sh
TalBarYakar: Fixed — flipped soft-success (return 0 / exit 0) to hard-fail (return 1 / exit 1) and upgraded WARNING → ERROR. The root cause (uv install failed) now lands at the top of the failed CI step instead of being buried under a misleading "venv missing" error later.

.install/lib/setup-python.sh
TalBarYakar: Added an inline comment explaining the pin: setuptools 81 (2025-07) removed the bundled distutils shim plus several legacy setup.py install paths; gevent and RLTest are the historical laggards on the install path. The comment also tells the next reader how to test whether the pin can be relaxed (drop the upper bound, run make bootstrap clean — if everything still installs, the constraint is obsolete).

.install/getrust.sh
TalBarYakar: Fixed both halves of your note. Rewrote the loop as if [ -e ... ]; then ln ... || true; fi (no SC2015, separate failure paths for "binary not present" and "ln failed"). Softened the comment from "we still fall back to the ~/.cargo/env bashrc hook" to "best-effort, silently no-ops if not writable; the bashrc hook only covers later login shells — the current shell would then not have cargo on PATH" so the limitation is honest.

.install/os/alma8.sh
TalBarYakar: Documented in both alma8.sh and rocky8.sh. The comment lays out the situation honestly: setup-python.sh runs against the venv's python3.11 which has dataclasses built-in, so nothing in this PR's call graph obviously consumes the shim — it's a safety net for any historical readies-driven script that still imports against bare python3. The comment also tells the next reader how to verify it's dead code (run make bootstrap && make build && make test on a clean EL8 image with the block removed). I'd rather not delete it without that real EL8 test run.

.github/workflows/flow-macos.yml
TalBarYakar: You're right that it's inconsistent. The straightforward "swap pip lines for install-python-deps" reading doesn't work cleanly though — json's flow-macos.yml uses a two-stage checkout (sparse → full) that invalidates the venv install_script.sh creates in the setup stage, and the macOS-26 SDK handling is inlined into each build step instead of being set once in $GITHUB_ENV like bloom/ts master do. Properly aligning means reshaping the job to match the bloom/ts pattern (full checkout first → install_script.sh → install-python-deps → build-module-and-redis → run-tests), which is ~30 lines moved around with macOS-26 SDK path implications I'd want to test on a real runner. Opening a focused follow-up PR for that — keeping this PR scoped to the cross-distro library work.

Linked: this is also why actions/run-tests shows as orphaned (#17) in this PR — it gets wired up as part of the same follow-up.

module.conf
TalBarYakar: Comment fixed. The consumer does exist — it's at scripts/sync-redis-conf.sh in the redis-core repo (not modules/sync-redis-conf.mk in this repo as the old comment claimed; that's a stale reference from before the script was rewritten + relocated). Updated the comment to point at the real consumer with the constants it matches against (PRIVATE_BEGIN / PRIVATE_END, exact-string match) so a reader can grep for the contract without ambiguity. Mirroring this same fix to bloom's module.conf:7 as part of the bloom follow-up.

Dockerfile.amazonlinux2023
TalBarYakar: Fixed — dropped the blank line between ARG SANITIZER= and the RUN so the comment now visually owns the whole block (comment + 2 ARGs + RUN as one contiguous group). Same nit exists in Dockerfile.alpine and Dockerfile.jammy; happy to mirror there too if you want consistency, but kept the change scoped to what was flagged.

gabsow (dismissed):
LGTM — followup commit `77b20d1` addresses 15 of 18 inline comments fully and 1 partially (module.conf comment now references the real consumer). The two skipped items are nits I'd flagged as optional: the orphaned `actions/run-tests/action.yml` (possibly intentional) and the inconsistent macOS bootstrap in `flow-macos.yml`. Both can land later.

Particularly nice touches in the followup: honest "this may be dead code, here's how to verify" comments on the EL8 dataclasses shim, and the rewritten `safe.directory` comment that actually explains the chicken-and-egg with `--local`.

.install/os/azurelinux3.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# shellcheck source=../lib/packages.sh
+. "$LIB/packages.sh"
+
+tdnf_default_install
+
+# Install aws-cli for uploading artifacts to s3
+curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
+unzip awscliv2.zip
+./aws/install
```
gabsow: Regression introduced by the followup: when you replaced `pip3 install awscli --upgrade` with the curl-based installer, you copied the *old* mariner2 pattern — not the fixed one. `os/mariner2.sh` was simultaneously updated to wrap this in `(cd /tmp && …)` to avoid leaving `awscliv2.zip` and an `./aws/` directory in the caller's CWD (which, when sourced from `install_script.sh`, is `.install/`).

The header comment here even says "Same shape as mariner2" — so let's actually make it the same shape:

```bash
(cd /tmp && curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o awscliv2.zip && unzip -q awscliv2.zip && ./aws/install)
```

Severity is cosmetic-to-minor (the image just ends up with stray files), but it's the exact bug the followup just fixed one file over.

gabsow (commented):
Spotted one regression on a re-read of the followup — flagging inline. Sorry for missing it the first time around.

.install/os/azurelinux3.sh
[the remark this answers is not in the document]
TalBarYakar: done

TalBarYakar: done

gabsow (approved):
Re-approving on `2261af0`. All my inline comments are addressed across the followup commits, including the `azurelinux3.sh` CWD regression I flagged on the second pass — it's now a verbatim subshell match with `mariner2.sh`.

Nice bonus catch from your side: the `pick_rltest_python` `exit 1` only killing the `$(…)` subshell — both call sites in `tests.sh` now propagate the failure with `|| exit 1`. That one was on me for not thinking through the consequence of my own "fail fast" suggestion. Good fix.

Two items deliberately left open from my original review (both nits I'd flagged as optional): the orphaned `actions/run-tests/action.yml` and the inconsistent macOS bootstrap in `flow-macos.yml`. Happy to leave both as follow-up.

PR #1600, opened 2026-06-02
MOD-15901 legacy rdb array - try reserve allocation

AvivDavid23:
> [!NOTE]
> **Medium Risk**
> Touches untrusted legacy RDB/RESTORE input paths; behavior change is safer allocation, with regression coverage for crash-class payloads.
> 
> **Overview**
> Fixes a **server crash** when loading **legacy RedisJSON RDB** (encver 0) arrays whose stored length is huge or invalid. Array loading no longer uses `Vec::with_capacity` on the untrusted length; it uses **`try_reserve_exact`** so oversize values (e.g. above `isize::MAX`) return a **recoverable error** instead of **SIGABRT**.
> 
> Adds a pytest that **forges a `RESTORE` payload** with a legacy array length of **2^62**, asserts **`RESTORE` fails with `ResponseError`**, and confirms the instance still answers **`PING`** and accepts normal **`JSON.SET`** afterward (VDP-4660 / MOD-15901).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2478692af845655165220d5e5ddd8ec5972b3541. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis_json/src/backward.rs
+++ b/redis_json/src/backward.rs
@@ -15,9 +15,6 @@ use serde_json::map::Map;
 use serde_json::Number;
 use serde_json::Value;
 
-/// Upper bound on element/entry counts accepted from the legacy (encver=0) RDB
-/// stream before allocation.
-const LEGACY_MAX_LEN: u64 = 1_048_576; // 1 << 20;
 
 #[derive(Debug, PartialEq)]
 enum NodeType {
@@ -76,11 +73,6 @@ pub fn json_rdb_load(rdb: *mut raw::RedisModuleIO) -> RedisResult<Value> {
         }
         NodeType::Dict => {
             let len = raw::load_unsigned(rdb)?;
-            if len > LEGACY_MAX_LEN {
-                return Err(RedisError::Str(
-                    "Can't load old RedisJSON RDB: dict length too large",
-                ));
-            }
             let mut m = Map::with_capacity(len as usize);
             for _ in 0..len {
                 let t = NodeType::try_from(raw::load_unsigned(rdb)?)?;
@@ -94,11 +86,6 @@ pub fn json_rdb_load(rdb: *mut raw::RedisModuleIO) -> RedisResult<Value> {
         }
         NodeType::Array => {
             let len = raw::load_unsigned(rdb)?;
-            if len > LEGACY_MAX_LEN {
-                return Err(RedisError::Str(
-                    "Can't load old RedisJSON RDB: array length too large",
-                ));
-            }
             let mut v = Vec::new();
             v.try_reserve_exact(len as usize)?;
             for _ in 0..len {
--- a/redis_json/src/backward.rs
+++ b/redis_json/src/backward.rs
@@ -15,7 +15,6 @@ use serde_json::map::Map;
 use serde_json::Number;
 use serde_json::Value;
 
-
 #[derive(Debug, PartialEq)]
 enum NodeType {
     Null,
--- a/redis_json/src/backward.rs
+++ b/redis_json/src/backward.rs
@@ -15,6 +15,10 @@ use serde_json::map::Map;
 use serde_json::Number;
 use serde_json::Value;
 
+/// Upper bound on element/entry counts accepted from the legacy (encver=0) RDB
+/// stream before allocation.
+const LEGACY_MAX_LEN: u64 = 1 << 20;
+
 #[derive(Debug, PartialEq)]
 enum NodeType {
     Null,
@@ -72,6 +76,11 @@ pub fn json_rdb_load(rdb: *mut raw::RedisModuleIO) -> RedisResult<Value> {
         }
         NodeType::Dict => {
             let len = raw::load_unsigned(rdb)?;
+            if len > LEGACY_MAX_LEN {
+                return Err(RedisError::Str(
+                    "Can't load old RedisJSON RDB: dict length too large",
+                ));
+            }
             let mut m = Map::with_capacity(len as usize);
             for _ in 0..len {
                 let t = NodeType::try_from(raw::load_unsigned(rdb)?)?;
@@ -85,7 +94,13 @@ pub fn json_rdb_load(rdb: *mut raw::RedisModuleIO) -> RedisResult<Value> {
         }
         NodeType::Array => {
             let len = raw::load_unsigned(rdb)?;
-            let mut v = Vec::with_capacity(len as usize);
+            if len > LEGACY_MAX_LEN {
+                return Err(RedisError::Str(
+                    "Can't load old RedisJSON RDB: array length too large",
+                ));
+            }
+            let mut v = Vec::<Value>::new();
+            v.try_reserve_exact(len as usize)?;
             for _ in 0..len {
                 let nested = json_rdb_load(rdb)?;
                 v.push(nested);
--- a/redis_json/src/backward.rs
+++ b/redis_json/src/backward.rs
@@ -17,7 +17,7 @@ use serde_json::Value;
 
 /// Upper bound on element/entry counts accepted from the legacy (encver=0) RDB
 /// stream before allocation.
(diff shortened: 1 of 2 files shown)


redis_json/src/backward.rs
```diff
@@ -15,6 +15,10 @@ use serde_json::map::Map;
 use serde_json::Number;
 use serde_json::Value;
 
+/// Upper bound on element/entry counts accepted from the legacy (encver=0) RDB
+/// stream before allocation.
+const LEGACY_MAX_LEN: u64 = 1 << 20;
```
gabsow: Nit: consider noting the literal value in the doc comment, e.g. `1_048_576 (1 << 20)`. Makes the cap discoverable without mental arithmetic when someone greps for tuning. Matches the value the VDP-4660 advisory recommended.

redis_json/src/backward.rs
```diff
@@ -72,6 +76,11 @@ pub fn json_rdb_load(rdb: *mut raw::RedisModuleIO) -> RedisResult<Value> {
         }
         NodeType::Dict => {
             let len = raw::load_unsigned(rdb)?;
+            if len > LEGACY_MAX_LEN {
+                return Err(RedisError::Str(
+                    "Can't load old RedisJSON RDB: dict length too large",
+                ));
+            }
             let mut m = Map::with_capacity(len as usize);
```
gabsow, line 75: Asymmetry vs. the Array branch: this still uses `with_capacity` without a `try_reserve`. Today that's safe because we don't enable `serde_json`'s `preserve_order` feature, so `Map` is `BTreeMap<String, Value>` and `with_capacity` is effectively a no-op. **But** if anyone ever flips that feature on (directly or via a transitive dep that unions it in), `Map` becomes `IndexMap` and `with_capacity(1<<20)` will eagerly allocate ~tens of MB and re-open a smaller version of this attack surface (no abort, but big amplification from a ~37-byte RESTORE).

Two options:
1. Add a one-line comment here explaining the invariant, e.g. `// Safe: !preserve_order means Map is BTreeMap and with_capacity is a no-op. Mirror the Array try_reserve dance if that ever changes.`
2. Just mirror the Array pattern: `let mut m = Map::new(); /* IndexMap has try_reserve; BTreeMap ignores capacity hints anyway */`.

I'd lean toward (1) given current deps.

redis_json/src/backward.rs
```diff
@@ -85,7 +94,13 @@ pub fn json_rdb_load(rdb: *mut raw::RedisModuleIO) -> RedisResult<Value> {
         }
         NodeType::Array => {
             let len = raw::load_unsigned(rdb)?;
-            let mut v = Vec::with_capacity(len as usize);
+            if len > LEGACY_MAX_LEN {
+                return Err(RedisError::Str(
+                    "Can't load old RedisJSON RDB: array length too large",
+                ));
+            }
+            let mut v = Vec::<Value>::new();
```
gabsow: Nit: turbofish is unneeded — the `v.push(nested)` below pins the element type, so `let mut v = Vec::new();` infers `Vec<Value>` just fine. Trivial.

tests/pytest/test.py
```diff
@@ -374,6 +375,113 @@ def testBackwardRDB(env):
...
+    if env.useAof:
+        env.skip()
+
+    conn = env.getConnection()
+    conn.execute_command('JSON.SET', 'tmp', '$', '[0]')
+    genuine_dump = conn.execute_command('DUMP', 'tmp', NEVER_DECODE=True)
+
+    # 2^62 overflows isize::MAX in Vec::with_capacity (and IndexMap if preserve_order is ever
+    # enabled); it also far exceeds LEGACY_MAX_LEN so the new bound check rejects it.
+    payload = _build_legacy_restore_payload(genuine_dump, node_discriminant, 2 ** 62)
```
gabsow: Coverage suggestion: only exercising `2 ** 62` proves the `isize::MAX`/`capacity_overflow` path is closed, but it doesn't pin the `LEGACY_MAX_LEN` bound itself. A future refactor that replaces `LEGACY_MAX_LEN` with something like `isize::MAX as u64` would still pass this test while quietly re-introducing the unbounded-`with_capacity` regression class.

Consider parametrizing this helper (or adding a second one) with a `(1 << 20) + 1` case — it's tiny, allocates almost nothing, and locks the bound in. The existing `2 ** 62` case stays as the smoke test for the abort path.

gabsow (commented):
Fix is correct and matches the VDP-4660 advisory recommendation (bound check + fallible alloc). A few small suggestions inline; none block merging.

redis_json/src/backward.rs
AvivDavid23: 1. We don't need a comment. The fact that even if it safe, I guard it is enough. Its future proof
2. There is no `try_reserve` for `serde` ```
Map
```

File: redis_json/src/backward.rs [hunk 5]
```diff
@@ -85,7 +94,13 @@ pub fn json_rdb_load(rdb: *mut raw::RedisModuleIO) -> RedisResult<Value> {
...
         NodeType::Array => {
             let len = raw::load_unsigned(rdb)?;
-            let mut v = Vec::with_capacity(len as usize);
+            if len > LEGACY_MAX_LEN {
+                return Err(RedisError::Str(
+                    "Can't load old RedisJSON RDB: array length too large",
+                ));
+            }
+            let mut v = Vec::new();
+            v.try_reserve_exact(len as usize)?;
```
gabsow, line 89: what is the user expirencsing in this case ?

redis_json/src/backward.rs
```diff
@@ -15,6 +15,10 @@ use serde_json::map::Map;
 use serde_json::Number;
 use serde_json::Value;
 
+/// Upper bound on element/entry counts accepted from the legacy (encver=0) RDB
+/// stream before allocation.
+const LEGACY_MAX_LEN: u64 = 1_048_576; // 1 << 20;
```
gabsow: i prefer not to have a fixed limit lets thing of some fallible allocation

redis_json/src/backward.rs
gabsow: Optional: drop the cap on this branch and let the no-preallocate `BTreeMap` (or future `IndexMap`-with-try_reserve) handle it.

```suggestion
            let mut m = Map::new();
```

redis_json/src/backward.rs
gabsow: Optional: drop the cap — `try_reserve_exact` on the next line already catches `len > isize::MAX` (the VDP-4660 PoC) and OOM.

```suggestion
            let mut v = Vec::new();
```

redis_json/src/backward.rs
gabsow: If you take the two suggestions below, this const is no longer referenced — delete lines 18-21.

gabsow (commented):
Optional follow-up — an alternative if you'd rather not keep the arbitrary `LEGACY_MAX_LEN` cap.

`Vec::try_reserve_exact` already returns `TryReserveErrorKind::CapacityOverflow` for `len > isize::MAX` — which is exactly the VDP-4660 PoC at `2^62`. So the bound check on the Array branch is redundant with the fallible-alloc layer.

On the Dict branch, `Map::with_capacity` is a no-op today because `serde_json` is built without `preserve_order` (so `Map` is `BTreeMap`, which doesn't preallocate). `Map::new()` is functionally identical right now and stays safe even if `preserve_order` is ever flipped on.

If you adopt the two suggestions below, the `LEGACY_MAX_LEN` const at the top of the file becomes unused (delete it), and my earlier `(1 << 20) + 1` test suggestion is moot. Just an option — happy with either direction.

gabsow (approved):

Issue #949, opened 2023-04-02
Add support form comparison with composite values


LiorKogan:
RedisJSON 2.4.6

```
JSON.SET j $ '{"a":[1,2,3,"a","b","c",false,true,["a",1],{"a":1}]}'
```

RedisJSON does not support comparison with an composite value.

The following are not supported:

```
JSON.GET j '$.a[?(@==["a",1])]'
JSON.GET j '$.a[?(@=={"a":1})]'
```

PR #1602, opened 2026-06-15
MOD-16274: Extended JSONPath - Literals, Functions, `in/nin` and Arithmetic operations

AvivDavid23:
RFC-9535-standard or widely-used filter capabilities we were missing. All additions live in the existing path dialect; 

- **JSONPath literals comparison** — array/object literals as `==` / `!=` operands (#949). Deep equality via a small internal `Literal` type compared through the existing cross-type `is_equal`.
- **Filter negation** — `!` operator, binding tighter than `&&` / `||`.
- **Function infra + RFC functions** — `length`, `count`, `value`, `match` (anchored), `search` (substring), usable inside filter expressions.
- **in / nin operators** — set membership; numbers coerce int/float (aligned with `==`), structured elements use deep equality.
- **Arithmetic operations** — `+ - * / %` and unary `+` / `-` with standard precedence. Operators must be space-separated (the path literal charset already includes `+ - / %`); division is float; div/mod-by-zero and non-numeric operands yield Nothing (comparison false) — a documented divergence from PostgreSQL, which errors.

Examples:

```text
$[?@ == [1]]  /  $[?@ == {"a":1}]    literal comparison
$[?!(@.a == 1)]                       negation
$[?length(@) > 2]                     function
$[?@.val in @.vals]                   membership
$[?@.a + @.b * 2 == 7]                arithmetic
```

---

> [!NOTE]
> **Medium Risk**
> Large changes to hot-path filter evaluation and a documented parse break for bare `in`/`nin` path segments; semantics are subtle (Nothing vs error, numeric coercion) but heavily tested.
> 
> **Overview**
> Extends JSONPath **filter expressions** toward RFC 9535: richer comparisons and logic without changing non-filter path syntax.
> 
> **Operands and comparisons:** Filters can use **array/object literals** in `==` / `!=` (deep equality via `is_equal` / a new `Literal` term type). **`in` / `nin`** test membership against array literals, document arrays, or path-resolved lists (with int/float coercion aligned to scalar `==`). Ordering ops against structured literals stay non-comparable (no matches).
> 
> **Logic and tests:** **`!`** negates a filter operand (tighter than `&&` / `||`). Bare filter terms are existence tests; **boolean** terms (including from `match` / `search`) use their value.
> 
> **Functions and regex:** **`length`**, **`count`**, **`value`**, **`match`** (anchored), and **`search`** (substring) are callable in filters; `=~` and `search` share a **per-query regex cache**.
> 
> **Arithmetic:** Space-separated **`+ - * / %`** and unary `+`/`-` with normal precedence; invalid ops (e.g. div/mod by zero, non-numeric operands) yield **Nothing** (filter false) instead of errors.
> 
> **Parsing caveat:** Bare first segments **`in` / `nin`** are reserved as operators, not field names—use `$.in` or `$["in"]` instead.
> 
> Coverage is expanded in **Rust unit/integration tests** and **Redis JSON module pytest** for the new filter features.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 204983692666ffe9bf41eb0f15baeda74c657019. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


json_path/src/json_path.rs
```diff
@@ -224,6 +224,291 @@ fn unescape_string_value<'a>(pair: Pair<'a, Rule>) -> Cow<'a, str> {
...
+    match pair.as_rule() {
+        Rule::array_literal => Literal::Array(pair.into_inner().map(build_literal).collect()),
+        Rule::object_literal => Literal::Object(
+            pair.into_inner()
+                .map(|member| {
+                    let mut it = member.into_inner();
+                    let key = it
+                        .next()
+                        .map(|k| unescape_string_value(k).into_owned())
+                        .unwrap_or_default();
```
gabsow, line 246: **Object-literal keys may be silently dropped.** The grammar rule is `object_member = {string ~ ":" ~ json_value}`, so the first child here is a `Rule::string` pair. But `unescape_string_value` only matches `string_value | string_value_escape_1 | string_value_escape_2` — if `string` is a non-silent wrapper rule, this falls through to its default and `unwrap_or_default()` yields `""`, turning `{"a":1}` into `{"":1}`.

Existing tests use simple unescaped keys, so this could pass CI yet break on escaped/quoted keys. Could you confirm `string`'s rule kind in `grammar.pest`? If it's not silent/atomic, `unescape_string_value` should recurse into `Rule::string` (or `build_literal` should descend one level before unescaping).

---

json_path/src/json_path.rs
```diff
@@ -224,6 +224,291 @@ fn unescape_string_value<'a>(pair: Pair<'a, Rule>) -> Cow<'a, str> {
...
+    Int(i64),
+    Float(f64),
+}
+
+/// Numeric equality matching `==`: exact for two integers, by `f64` value otherwise.
+fn numbers_equal(a: Num, b: Num) -> bool {
+    match (a, b) {
+        (Num::Int(x), Num::Int(y)) => x == y,
+        (Num::Float(x), Num::Float(y)) => x == y,
+        (Num::Int(x), Num::Float(y)) | (Num::Float(y), Num::Int(x)) => x as f64 == y,
```
gabsow, line 292: **int→f64 coercion can collide for large integers.** `x as f64 == y` loses precision above 2^53, so e.g. `9007199254740993` and `9007199254740992` compare equal. This matches the existing `is_equal` behavior, so it's consistent — but since `in`/`nin` now routes through here, large-integer membership could give surprising matches. Worth at least a comment noting the limitation.

---

json_path/src/json_path.rs
```diff
@@ -224,6 +224,291 @@ fn unescape_string_value<'a>(pair: Pair<'a, Rule>) -> Cow<'a, str> {
...
+}
+
+fn regex_is_match(regex: &str, s: &str) -> bool {
+    Regex::new(regex).is_ok_and(|re| re.is_match(s))
+}
+
+/// Anchored (full-string) regex match used by RFC 9535 `match()`.
+fn regex_is_full_match(regex: &str, s: &str) -> bool {
+    Regex::new(&format!("^(?:{regex})$")).is_ok_and(|re| re.is_match(s))
+}
```
gabsow: **Regex recompiled on every element (perf).** Both helpers call `Regex::new(...)` per invocation. For `$.a[?match(@, "pat")]` over an N-element array, the same pattern is compiled N times. The pattern is invariant across the evaluation — compile it once, ideally at query-compile time, or via a small thread-local cache.

Also: this duplicates the existing `TermEvaluationResult::re_is_match` (used by `=~`). Consolidating into a single shared helper would let the compile-once fix apply uniformly to `match`/`search`/`=~`.

---

json_path/src/json_path.rs
```diff
@@ -224,6 +224,291 @@ fn unescape_string_value<'a>(pair: Pair<'a, Rule>) -> Cow<'a, str> {
...
+            match (s, re) {
+                (Some(s), Some(re)) => TermEvaluationResult::Bool(if full {
+                    regex_is_full_match(&re, &s)
+                } else {
+                    regex_is_match(&re, &s)
+                }),
+                _ => TermEvaluationResult::Bool(false),
+            }
+        }
+        _ => TermEvaluationResult::Invalid,
```
gabsow: **Unknown function names fail silently.** A typo like `lenght(@)` lands here and returns `Invalid`, so the query just yields no results with no signal. Since the full set of valid function names is known at compile time, a parse-time error would be friendlier; at minimum a `trace!` here would aid debugging.

---

json_path/src/grammar.pest
```diff
@@ -26,30 +26,66 @@ from_root    = {"$" ~ root?}
...
     ge = @{">="}
     gt = @{">"}
     le = @{"<="}
     lt = @{"<"}
     eq = @{"=="}
     ne = @{"!="}
     re = @{"=~"}
+    // Set-membership operators (word operators; require a non-alphanumeric boundary)
+    in_op = @{"in" ~ !ASCII_ALPHANUMERIC}
+    nin_op = @{"nin" ~ !ASCII_ALPHANUMERIC}
```
gabsow: **Boundary inconsistent with `op_keyword`.** Here the boundary is `!ASCII_ALPHANUMERIC`, but `op_keyword` (line 103) uses `!(ASCII_ALPHANUMERIC | "_")`. So these rules would accept `in` followed by `_` (e.g. an `in_` prefix) while `op_keyword` correctly rejects it. Align both to `!(ASCII_ALPHANUMERIC | "_")` so the operator and the path-segment guard agree.

---

json_path/src/grammar.pest
```diff
@@ -26,30 +26,66 @@ from_root    = {"$" ~ root?}
...
+function_name = @{ASCII_ALPHA ~ (ASCII_ALPHANUMERIC | "_")*}
+function_call = {function_name ~ "(" ~ (term ~ ("," ~ term)*)? ~ ")"}
+
+term = _{select_term | decimal | string | boolean | null | array_literal | object_literal | function_call}
+
+// Arithmetic over filter operands. Operators must be space-separated, since the path
+// `literal` charset already includes + - / % (so `@.a+1` is a field name, while
+// `@.a + 1` is addition). Precedence: unary, then * / %, then + -.
+// Division/modulo by zero (and any non-numeric operand) yields Nothing, so the
+// comparison is false and the node is skipped (unlike PostgreSQL, which errors).
```
gabsow: The "operators must be space-separated" requirement is enforced only as a side effect of `literal`'s charset including `+ - / %` (so `@.a+1` parses as a field name). That's a subtle, implicit invariant — a future change to the `literal` charset could silently break arithmetic parsing. Worth a dedicated test that locks in both directions (e.g. `@.a + 1` is addition, `@.a+1` is a field name) so the behavior is regression-protected.

---

gabsow (commented):
Reviewed commit by commit — clean feature splitting, solid test coverage, good architecture overall. A few things worth addressing, inline comments attached:

**Worth resolving before merge**
- `match`/`search` recompile the regex per element (commit 3) — invariant pattern, should be compiled once.
- Object-literal key parsing may silently drop/blank keys depending on the `string` rule kind (commit 1).

**Smaller**
- `in_op`/`nin_op` boundary differs from `op_keyword` (`_` handling) (commit 4).
- int→f64 coercion precision for large ints in membership (commit 4).
- Unknown function names fail silently (commit 3).
- The "space-separated arithmetic" invariant is implicit — add a regression test (commit 5).

**Nice work**: the `evaluate_filter_operand` refactor (commit 2) de-duplicates the old dispatch cleanly; the integer-overflow→float fallback in `num_binop` and the documented div-by-zero→Nothing semantics (commit 5) are well thought through.

A couple of test gaps I'd suggest: invalid regex in `match`/`search`, empty/non-array RHS for `in`, and `length()` on scalar types.

---

json_path/src/json_path.rs
gabsow: Retracting this — I checked `grammar.pest`: `string` is a silent rule (`string = _{...}`), so `object_member.into_inner()` flattens it and yields `string_value`/`string_value_escape_*` directly, which `unescape_string_value` matches correctly. No bug here. Sorry for the noise.

---

json_path/src/grammar.pest
```diff
@@ -61,7 +97,12 @@ bracket = _{numbers_range | numbers_list | string_list | all | "?" ~ filter}
 
 element = _{"." ~ literal | full_scan ~ (literal | all | "[" ~ bracket ~ "]") | "." ~ all | ("."? ~ "[" ~ bracket ~ "]")}
 
-first_element = _{literal | all}
+// The bare word operators `in`/`nin` must not be swallowed as a bare path segment
+// (e.g. `@ in [...]`). A field literally named `in`/`nin` is still reachable via
+// `.in` / `["in"]`.
+op_keyword = @{("nin" | "in") ~ !(ASCII_ALPHANUMERIC | "_")}
+
+first_element = _{(!op_keyword ~ literal) | all}
```
gabsow, line 109: **Backward-compat regression: a bare first segment named `in`/`nin` no longer parses.** The `!op_keyword` guard rejects `in`/`nin` as the first un-dotted path segment. Verified against `master`: `compile("$in")` on `{"in":5}` returned `[5]` before, now errors (`expected EOI`). Same for `$nin`, and inside filters `@in == 1` (`expected arith_factor`).

The comment correctly notes `.in` / `["in"]` still work (those go through `element`, not `first_element`), but the bare-segment form was valid before and silently breaks now. Worth confirming whether any stored queries rely on it; if this is acceptable, a note in the changelog/docs would help.

---

json_path/src/json_path.rs
```diff
@@ -987,6 +1445,8 @@ impl<'i, UPTG: UserPathTrackerGenerator> PathCalculator<'i, UPTG> {
                 Rule::eq => term1_val.eq(&term2_val),
                 Rule::ne => term1_val.ne(&term2_val),
                 Rule::re => term1_val.re(&term2_val),
+                Rule::in_op => term1_val.member_of(&term2_val),
+                Rule::nin_op => !term1_val.member_of(&term2_val),
```
gabsow, line 1487: **`nin` selects every node when the RHS isn't an array.** `member_of` returns `false` for a non-array RHS — a scalar, or an absent/`Invalid` path — and `nin` negates that to `true`. So `$.items[?@.v nin @.missing]` (RHS path absent) matches **all** items, and `$.items[?@.v nin @.scalarField]` likewise. The `in` side has a test for the non-array RHS case (no match), but `nin` doesn't, so this slips through. Consider whether a non-collection RHS should make both `in` and `nin` false (or be a defined error), rather than `nin` becoming vacuously true.

---

json_path/src/json_path.rs
```diff
@@ -996,7 +1456,37 @@ impl<'i, UPTG: UserPathTrackerGenerator> PathCalculator<'i, UPTG> {
                 }
             }
         } else {
-            !matches!(term1_val, TermEvaluationResult::Invalid)
+            // A bare term is a test: a boolean result (e.g. `match(...)`) uses its
+            // value; any other present value is truthy (existence), `Invalid` is false.
+            match term1_val {
+                TermEvaluationResult::Bool(b) => b,
```
gabsow, line 1500: Minor / heads-up: this changes the bare-term test for `Bool` results. Previously a bare term passed iff it wasn't `Invalid` (pure existence), so `$[?false]` matched every node; now it returns none. This is arguably *more* correct (and is what makes `match()`/`search()` work as tests), so I'd keep it — just flagging it as an intentional semantic change worth a test asserting `$[?false]` → `[]` and `$[?true]` → all, so it's locked in deliberately rather than incidentally.

---

json_path/src/json_path.rs
```diff
@@ -224,6 +224,291 @@ fn unescape_string_value<'a>(pair: Pair<'a, Rule>) -> Cow<'a, str> {
     }
 }
 
+/// Recursively build an owned `Literal` value from an array/object literal
+/// parse subtree (e.g. `[1,{"a":2}]`). Used for structured filter operands.
+fn build_literal(pair: Pair<Rule>) -> Literal {
```
gabsow: **Perf: the literal is rebuilt from the parse tree on every element.** `build_literal` runs inside `evaluate_single_term`, which is called per array element / object member during filter evaluation. For `$.a[?@ in [1,2,…,1000]]` over an M-element array, the 1000-element `Literal::Array` (with all its `Vec`/`String` allocations) is reconstructed and dropped M times, even though it's constant across elements. Building literals once at `compile()` time and reusing them would make allocation independent of document size. (Larger refactor — fine as a follow-up, but worth tracking.)

---

json_path/src/select_value.rs
```diff
@@ -170,3 +170,302 @@ pub fn is_equal<T1: SelectValue, T2: SelectValue>(a: &T1, b: &T2) -> bool {
 
 #[allow(unused)]
 pub const MAX_DEPTH: usize = 128;
+
+/// A small owned JSON value used to represent array/object literals that appear
+/// as filter operands (e.g. `?@==[1]`, `?@=={"a":1}`). It implements `SelectValue`
+/// so it can be deep-compared against the document's value
+#[derive(Debug, Clone, Default)]
+pub(crate) enum Literal {
```
gabsow: Altitude question: `Literal` hand-rolls the full ~20-method `SelectValue` trait plus a manual `Serialize` impl. The crate already depends on owned-JSON value types (`serde_json::Value`, `ijson::IValue`) — implementing `SelectValue` for one of those (or reusing an existing impl) would give array/object-literal operands for free and avoid a third trait implementor that has to be updated every time `SelectValue` changes. The manual `Serialize` also risks diverging from serde's own on edge cases (float formatting, etc.). Not blocking, but worth considering whether this is the right layer.

---

gabsow (commented):
Follow-up from a deeper review pass (high-effort, multi-angle). Five net-new items, inline:

**Correctness**
- `$in`/`@in` bare path segments no longer parse — backward-compat regression (verified vs `master`).
- `nin` returns true for every node when the RHS isn't an array (scalar/absent path) — untested edge.
- Bare `Bool(false)` test now returns no nodes (e.g. `$[?false]`) — likely an intentional improvement, just worth a locking test.

**Perf / altitude**
- `build_literal` rebuilds the literal per element instead of once at compile time.
- `Literal` reimplements the whole `SelectValue` trait + `Serialize` rather than reusing an existing owned-JSON value type.

These are in addition to the earlier comments (literal-vs-`in` numeric coercion, regex recompilation, duplicate regex helper, `in_op` boundary). The numeric-coercion inconsistency remains the one I'd prioritize.

---

json_path/src/select_value.rs
AvivDavid23: ~~Right. The alternative, is adding some methods to the SelectValue trait, and each implementor will implement(from_array, from_map, etc) -> Will result every caller to add implementation, allocating extra(or less) memory, depending on the implementor overhead.~~

~~I prefer to go with one implementation which lives inside this crate, specifically for this case~~

update: will use serde value for now

gabsow (commented):
Re-reviewed after the latest commits (`CR`, `serde::Value` swap, comment rewording, and the `i64::MIN % -1` fix). Most of the earlier pass is resolved — thanks for the quick turnaround.

**Resolved ✅**
- Regex recompiled per element → cached (unified for `=~` / `match` / `search`)
- `in_op` / `op_keyword` `_`-boundary inconsistency → aligned
- Unknown function names → now emit a `trace!`
- Space-separated arithmetic invariant → regression test added
- `Literal` type → replaced with `serde_json::Value` (drops the hand-rolled `SelectValue` + `Serialize`)
- `i64::MIN % -1` modulo panic → `checked_rem` + reproducing test

**Still open — one blocker**
- **`nin` matches every node when the RHS isn't an array.** `nin = !member_of`, and `member_of` returns `false` for a non-array RHS (absent field, scalar, `Invalid`), so the negation makes `$.items[?@.v nin @.missing]` match **all** items — a typo'd field name silently returns the whole dataset. Cursor Bugbot independently flagged this HIGH. A non-collection RHS should make **both** `in` and `nin` return `false` (or be a defined error). If you and the team decide to accept the current semantics per the PRD's open TODO, that's a conscious call — but it should be explicit and tested, not incidental.

**Minor / non-blocking (fine as follow-ups)**
- `$in` / `@in` bare first-segment no longer parses — documented break; a CHANGELOG note would help.
- `build_literal` rebuilt per element (perf) — Bugbot flagged MEDIUM; worth a `compile()`-time materialization later.
- int→f64 precision >2^53 in membership — a one-line comment noting the limitation.
- Add a locking test for `$[?false]` → `[]` / `$[?true]` → all.

Happy to approve once the `nin` non-array case is settled. CI is green across Linux + most of macOS; a few macOS variants still finishing.

---

json_path/src/json_path.rs
gabsow: Revising this — **not a blocker.** Product confirmed we're aligning to Jayway JsonPath semantics, and I checked Jayway's evaluator source:

```java
// NotInEvaluator
return !evaluators.get(RelationalOperator.IN).evaluate(left, right, ctx);

// InEvaluator — non-array / undefined RHS:
if (vn.isUndefinedNode()) { return false; }
```

So in Jayway `nin = !in`, and `in` returns `false` for a non-array/undefined RHS, making `nin` return `true` — which is exactly what this PR does. The current behavior is therefore spec-conformant, not a bug. (Same for the Cursor Bugbot HIGH on this.)

Only ask: lock it in with a test (`@.v nin @.missing` / scalar RHS → matches) and a one-line comment noting it's intentional Jayway-aligned behavior, so it isn't "fixed" away later. No code change needed.

---

gabsow (approved):
LGTM ✅

Re-reviewed against the latest commit (`204983692`) with Product's confirmation that we're aligning to Jayway JsonPath semantics. All earlier findings are resolved:

- Regex recompile → cached (unified `=~` / `match` / `search`)
- `Literal` type → replaced with `serde_json::Value`
- `in_op` / `op_keyword` `_`-boundary → aligned
- Unknown function names → `trace!`
- `i64::MIN % -1` modulo panic → `checked_rem` + test
- Space-separated arithmetic → regression test

The `nin` non-array RHS case (which I and Cursor Bugbot had flagged HIGH) is **correct, not a bug** — Jayway defines `nin = !in`, and `in` returns false for a non-array/undefined RHS, so `nin` returns true. The PR matches that exactly, and it's now documented (json_path.rs:1485) and tested (`test_membership_nin_non_array_rhs`, lib.rs:807).

Remaining items are all optional polish, fine as follow-ups: CHANGELOG note for the `$in`/`@in` bare-segment parse break, `build_literal` per-element materialization at `compile()` time, an int→f64 >2^53 precision comment, and a `$[?false]`/`$[?true]` locking test.

Broader Jayway operator set (`subsetof`, `anyof`, `noneof`, `size`, `empty`) is tracked as follow-up subtasks under the initiative — out of scope for this slice.

Nice work. Approving.

---

PR #1603, opened 2026-06-16
MOD-16275: Extended JSONPath — set ops, `size`/`empty`, numeric/string/array functions (in filter expressions only)

AvivDavid23:
Follow-up to #1602 (MOD-16274). Implements the remaining "extended" JSONPath features.
Again, currently only for filter operations

## Features
- **Set relations:** `subsetof`, `anyof`, `noneof`
- **Size / emptiness:** `sizeof` (alias `size`), `empty <bool>`
- **Numeric functions:** `ceiling`, `floor`, `abs`
- **String function:** `concat`
- **Array functions:** `min`, `max`, `avg`, `sum`, `stddev`, `first`, `last`, `index`

Also, a follow-up from previous commit:
**Regex cache: add cap** — bound the per-query regex cache at 64 entries (a
  document-value pattern, e.g. `match(@.s, @.pattern)`, could otherwise retain one
  compiled `Regex` per element → unbounded memory); compile uncached past the cap.

---

> [!NOTE]
> **Medium Risk**
> Broad changes to query parsing and filter evaluation semantics (including documented bare-segment behavior); correctness and performance depend on new operators and uncached regex paths under heavy distinct-pattern loads.
> 
> **Overview**
> Extends **filter-only** JSONPath with RFC 9535-style **set relations** (`subsetof`, `anyof`, `noneof`), **length/emptiness** checks (`sizeof` / alias `size`, `empty <bool>`), and new **filter functions** (`ceiling`, `floor`, `abs`, `concat`, `sum`/`min`/`max`/`avg`/`stddev`, `first`/`last`/`index`).
> 
> The grammar and `op_keyword` guard are updated so these word operators are not parsed as bare first path segments (e.g. use `$.size` instead of `$size`). Set relations reuse `in`-style numeric coercion via shared `values_equal` / `value_in_array`; multi-node operands use the same any-of semantics as `==` and `in`. Invalid inputs and wrong arity yield **Nothing** (no match) rather than partial evaluation.
> 
> The per-query **regex cache is capped at 64** entries; additional distinct patterns compile uncached so document-driven `=~` / `match` cannot grow memory without bound. Rust unit tests and Redis JSON pytest coverage exercise the new behavior and edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eacd007c4651b15031f9cc0b164b4454dd67b2d6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/json_path/src/grammar.pest
+++ b/json_path/src/grammar.pest
@@ -1,4 +1,4 @@
-literal = @{('a' .. 'z' | 'A' .. 'Z' | '0' .. '9' | "-" | "_" | "`" | "+" | "#" | "%" | "$" | "^" | "/" | ":")+}
+literal = @{('a' .. 'z' | 'A' .. 'Z' | '0' .. '9' | "-" | "_" | "`" | "~" | "+" | "#" | "%" | "$" | "^" | "/" | ":")+}
 
 string = _{("'" ~ (string_value) ~ "'") | ("\"" ~ (string_value) ~ "\"") | ("\"" ~ (string_value_escape_1) ~ "\"") | ("'" ~ (string_value_escape_2) ~ "'")}
     string_escape = @{"\\"}
@@ -102,15 +102,9 @@ all = {"*"}
 
 full_scan = {".."}
 
-// Get-keys: `$.obj~` (terminal `~`) or its function alias `$.obj.keys()`. Both emit the
-// object's member names as string results. `~` was removed from the `literal` charset, so
-// a field literally named with `~` is now reachable only via `["a~b"]`.
-get_keys = {"~"}
-get_keys_fn = {"keys" ~ "(" ~ ")"}
-
 bracket = _{numbers_range | numbers_list | string_list | all | "?" ~ filter}
 
-element = _{"." ~ get_keys_fn | "." ~ literal | full_scan ~ (literal | all | "[" ~ bracket ~ "]") | "." ~ all | ("."? ~ "[" ~ bracket ~ "]") | get_keys}
+element = _{"." ~ literal | full_scan ~ (literal | all | "[" ~ bracket ~ "]") | "." ~ all | ("."? ~ "[" ~ bracket ~ "]")}
 
 // The bare word operators must not be swallowed as a bare first path segment
 // (e.g. `@ in [...]`). A field with one of these names is still reachable via
--- a/json_path/src/grammar.pest
+++ b/json_path/src/grammar.pest
@@ -112,7 +112,9 @@ element = _{"." ~ literal | full_scan ~ (literal | all | "[" ~ bracket ~ "]") | 
 // NOTE: this is a deliberate, documented behavior change: a bare first segment with one
 // of these names (e.g. `$in`, `@size`) no longer parses as a path to that field — use
 // `$.size` / `$["size"]` instead.
-op_keyword = @{("subsetof" | "noneof" | "anyof" | "sizeof" | "size" | "empty" | "nin" | "in") ~ !(ASCII_ALPHANUMERIC | "_")}
+// Order mirrors the `op` rule's word-operator alternation (with `sizeof` before its
+// prefix `size`), so the operator list and this path-segment guard stay in agreement.
+op_keyword = @{("in" | "nin" | "subsetof" | "anyof" | "noneof" | "sizeof" | "size" | "empty") ~ !(ASCII_ALPHANUMERIC | "_")}
 
 first_element = _{(!op_keyword ~ literal) | all}
 
--- a/json_path/src/grammar.pest
+++ b/json_path/src/grammar.pest
@@ -1,4 +1,4 @@
-literal = @{('a' .. 'z' | 'A' .. 'Z' | '0' .. '9' | "-" | "_" | "`" | "~" | "+" | "#" | "%" | "$" | "^" | "/" | ":")+}
+literal = @{('a' .. 'z' | 'A' .. 'Z' | '0' .. '9' | "-" | "_" | "`" | "+" | "#" | "%" | "$" | "^" | "/" | ":")+}
 
 string = _{("'" ~ (string_value) ~ "'") | ("\"" ~ (string_value) ~ "\"") | ("\"" ~ (string_value_escape_1) ~ "\"") | ("'" ~ (string_value_escape_2) ~ "'")}
     string_escape = @{"\\"}
@@ -102,9 +102,15 @@ all = {"*"}
 
 full_scan = {".."}
 
+// Get-keys: `$.obj~` (terminal `~`) or its function alias `$.obj.keys()`. Both emit the
+// object's member names as string results. `~` was removed from the `literal` charset, so
+// a field literally named with `~` is now reachable only via `["a~b"]`.
+get_keys = {"~"}
+get_keys_fn = {"keys" ~ "(" ~ ")"}
+
 bracket = _{numbers_range | numbers_list | string_list | all | "?" ~ filter}
 
-element = _{"." ~ literal | full_scan ~ (literal | all | "[" ~ bracket ~ "]") | "." ~ all | ("."? ~ "[" ~ bracket ~ "]")}
+element = _{"." ~ get_keys_fn | "." ~ literal | full_scan ~ (literal | all | "[" ~ bracket ~ "]") | "." ~ all | ("."? ~ "[" ~ bracket ~ "]") | get_keys}
 
 // The bare word operators must not be swallowed as a bare first path segment
 // (e.g. `@ in [...]`). A field with one of these names is still reachable via
--- a/json_path/src/grammar.pest
+++ b/json_path/src/grammar.pest
@@ -26,7 +26,7 @@ from_root    = {"$" ~ root?}
 
 select_term = _{from_current | from_root}
 
-op = _{ge | gt | le | lt | eq | ne | re | in_op | nin_op | subsetof_op | anyof_op | noneof_op}
+op = _{ge | gt | le | lt | eq | ne | re | in_op | nin_op | subsetof_op | anyof_op | noneof_op | size_op | empty_op}
     ge = @{">="}
     gt = @{">"}
     le = @{"<="}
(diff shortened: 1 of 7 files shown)


json_path/src/json_path.rs
```diff
@@ -302,6 +302,26 @@ fn value_as_number<V: SelectValue>(v: &V) -> Option<Num> {
...
+/// (`is_equal`): nested numbers do not coerce, so `1` ≠ `1.0`. Non-array `haystack` ⇒ false.
+fn value_in_array<'i, 'j, S: SelectValue, V: SelectValue>(
+    needle: &V,
+    haystack: &TermEvaluationResult<'i, 'j, S>,
+) -> bool {
+    match haystack {
+        TermEvaluationResult::Value(v) if v.as_ref().get_type() == SelectValueType::Array => v
+            .as_ref()
+            .values()
+            .is_some_and(|mut it| it.any(|e| is_equal(needle, e.as_ref()))),
```
gabsow: **Set ops are type-strict while `in`/`nin` coerce numbers — same data, opposite answers.** `value_in_array` compares with `is_equal`, which requires matching type (`1` is Long, `1.0` is Double, so `1` ≠ `1.0`). But `member_of`/`equals_value` behind `in`/`nin` coerce via `numbers_equal`. So `$.a[?2 in @.vals]` matches `vals:[1.0,2.0]` (existing behavior), yet `$.a[?@ subsetof @.vals]` for `[2]` against `[1.0,2.0]` does **not** match.

It's documented and pinned by `test_set_subsetof_type_strict`, so this looks intentional — flagging because two operator families that both mean "is a member of" disagreeing on int-vs-float is a contract decision worth an explicit confirm (and worth checking against the Jayway target, since that's what we aligned `in`/`nin` to). If it's deliberate, a one-line note in the user-facing docs would save confusion.

---

json_path/src/json_path.rs
```diff
@@ -945,6 +1152,103 @@ impl<'i, 'j, S: SelectValue> TermEvaluationResult<'i, 'j, S> {
...
+                SelectValueType::Array => v.len(),
+                _ => None,
+            }
+        }
+        match self {
+            TermEvaluationResult::Str(s) => Some(s.chars().count()),
+            TermEvaluationResult::String(s) => Some(s.chars().count()),
+            TermEvaluationResult::Value(v) => arr_or_str_len(v.as_ref()),
+            TermEvaluationResult::Literal(l) => arr_or_str_len(l),
+            TermEvaluationResult::NodeList(list) if list.len() == 1 => {
```
gabsow: **`sizeof`/`empty` silently evaluate to false for a multi-result left operand.** `seq_length` only sizes a `NodeList` when `list.len() == 1`; any other multi-node list falls through to `None`, so the operator is `false`. Other comparison ops (`==`, `<`, `in`) apply any-of-nodelist semantics over a multi-result left side, so `$[?(@.* sizeof 2)]` is always false even when the matched nodes are size-2 arrays/strings.

Either apply the same any-of semantics here, or document that `sizeof`/`empty` require a single-node left operand. A test over a multi-result left side (`@.*`) would lock in whichever behavior is intended.

---

json_path/src/grammar.pest
```diff
@@ -98,13 +106,13 @@ bracket = _{numbers_range | numbers_list | string_list | all | "?" ~ filter}
...
-// `in`/`nin` (e.g. `$in`, `@in`) no longer parses as a path to that field — use
-// `$.in` / `$["in"]` instead.
-op_keyword = @{("nin" | "in") ~ !(ASCII_ALPHANUMERIC | "_")}
+// The bare word operators must not be swallowed as a bare first path segment
+// (e.g. `@ in [...]`). A field with one of these names is still reachable via
+// `.name` / `["name"]`.
+// NOTE: this is a deliberate, documented behavior change: a bare first segment with one
+// of these names (e.g. `$in`, `@size`) no longer parses as a path to that field — use
+// `$.size` / `$["size"]` instead.
+op_keyword = @{("subsetof" | "noneof" | "anyof" | "sizeof" | "size" | "empty" | "nin" | "in") ~ !(ASCII_ALPHANUMERIC | "_")}
```
gabsow: **`op_keyword` now reserves `size` and `empty` as bare first segments — wider compat break than `in`/`nin`.** `size` and `empty` are very common JSON keys, so a bare first segment like `$size` / `@empty` that previously addressed that field now fails to parse. Scope is narrow (only the bare first segment; `$.size`, `$["size"]`, and `@.size` still work) and the comment documents it — but given how common these keys are, this is worth an explicit CHANGELOG/docs callout for the release, beyond the inline comment.

(Also a small consistency nit: the alternation order here differs from the `op` rule's order. Harmless today because every keyword has the `!(ASCII_ALPHANUMERIC | "_")` boundary and the only prefix pair `sizeof`/`size` is longest-first — but the comment says the two lists must "agree", so keeping them in the same order would prevent a future prefix-keyword from parsing as an op while the segment guard disagrees.)

---

json_path/src/json_path.rs
```diff
@@ -519,6 +703,29 @@ fn eval_function<'i, 'j, S: SelectValue>(
             TermEvaluationResult::Integer(function_count(a))
         }),
         "value" if args.len() == 1 => function_value(args.pop().unwrap()),
+        "ceiling" => function_round(args.first(), f64::ceil),
```
gabsow, line 733: **New functions don't validate arity.** `value`/`length` are gated with `args.len() == 1`, but `ceiling`/`floor`/`abs`/`concat`/`sum`/`min`/`max`/`avg`/`stddev`/`index` use `args.first()` / `into_iter().next()` with no arity guard, so extra args are silently ignored: `ceiling(@, 99)` rounds the first arg and discards `99`; `index(@.n, 1, 2)` ignores the third arg. A malformed query that should be Nothing/error instead "works" on the first operand. Consider rejecting wrong arity (→ `Invalid`) for consistency with `value`/`length`.

---

json_path/src/json_path.rs
```diff
@@ -490,22 +510,186 @@ type RegexCache = HashMap<String, Option<Regex>>;
...
+            .checked_abs()
+            .map_or(TermEvaluationResult::Invalid, TermEvaluationResult::Integer),
+        Some(Num::Float(f)) => TermEvaluationResult::Float(f.abs()),
+        None => TermEvaluationResult::Invalid,
+    }
+}
+
+/// `concat(s1, s2, ...)`: concatenate string arguments into one string. Any non-string
+/// argument yields Nothing.
+fn function_concat<'i, 'j, S: SelectValue>(
```
gabsow, line 589: **`concat()` with zero arguments returns `String("")` instead of Nothing.** The loop never runs for an empty arg list, so `concat()` yields an empty string and `$[?concat() == ""]` would match every node. Every other new function returns `Invalid` when the first arg is missing (`args.first() == None`); `concat` is the outlier because it folds over the slice. Minor, but worth an `if args.is_empty() { return Invalid }` guard for consistency.

---

gabsow (commented):
High-effort review pass (multi-angle: line-by-line, removed-behavior, cross-file, plus reuse/simplification/efficiency/altitude). Overall this is in good shape — clean function decomposition and genuinely thorough test coverage including the negative/edge cases (overflow→Nothing, non-numeric elements, empty arrays, type-strict set comparison).

**No hard blockers.** Inline comments cover five semantic/correctness items; the two I'd want resolved or consciously accepted before merge are both contract calls:
- **Type-strict set ops vs coercing `in`/`nin`** (inline) — confirm this is the intended user-facing contract.
- **`sizeof`/`empty` on a multi-result left operand** (inline) — silently false, asymmetric with other ops' nodelist semantics.

Plus `op_keyword` reserving `size`/`empty` deserves a CHANGELOG note, and two low-severity consistency nits (function arity, `concat()` zero-arg).

**Cleanup / efficiency (non-blocking, fine as follow-ups):**

1. **The 3-arm "array-shaped term" match is copy-pasted 4×** — `value_in_array`, `term_number_seq`, `set_relate`, and the pre-existing `member_of` each re-open the same `Value(Array)` / `Literal(Value::Array)` / `NodeList` skeleton. A single `array_elements(term) -> Option<impl Iterator<...>>` helper would collapse them; today a new variant or a `values()` fix is a 4-site edit (and that duplication is what makes the type-strict/nodelist issues easy to introduce).

2. **`seq_length` largely duplicates `value_length`/`function_length`** (the RFC `length` helper from #1602) — the only real difference is excluding objects. Consider wrapping the existing core rather than a parallel copy.

3. **`set_relate` is O(n·m), rescanning the invariant RHS per element** — `subsetof`/`anyof`/`noneof` re-walk the whole RHS for each LHS element, and the filter runs once per document node. Materializing the RHS element list once per evaluation avoids the repeated 3-arm dispatch.

4. **Float→i64 truncation is hand-rolled in three spots** with inconsistent overflow behavior — `function_round` (bounds-checked → `Invalid`) vs `size_of` and the `index` n-arg (bare `as i64`, which saturates). One `as_i64_truncating` helper would make the overflow/negative policy explicit and consistent.

5. **`term_number_seq` allocates a `Vec<f64>`** that `sum`/`min`/`max`/`avg` each fold over once — only `stddev` needs a buffer (or use Welford's one-pass). A streaming fold avoids a per-node allocation sized to the array.

Verified-and-fine (so nobody re-investigates): `function_round`'s `r < i64::MAX as f64` strict upper bound is correct (`i64::MAX as f64` rounds up to 2^63); `function_abs` maps `i64::MIN` to `Invalid` via `checked_abs`; `function_index`'s owned branch clones via `inner_cloned()` before drop; `resolve()`'s negative-index arithmetic can't overflow and `try_from` rejects negatives; the regex-cache cap is correctly bounded (the past-cap uncached path is the intended memory bound from #1602, not a regression); and the ±INFINITY agg seeds are safe because `term_number_seq` returns `None` for empty arrays.

Nice work.

---

gabsow (dismissed):
LGTM ✅ — re-reviewed after the latest CR commit. All five findings from the earlier pass are resolved:

- **Set ops type-strict vs `in`/`nin` coercion** → new `values_equal()` coerces numbers; `value_in_array` updated; `test_set_subsetof_numeric_coercion` locks it in.
- **`sizeof`/`empty` silent false on multi-result left operand** → any-of NodeList handling added to `size_of` and `empty_check`; two new tests.
- **Function arity not validated** → `arity_ok` block at top of `eval_function`; `test_function_wrong_arity_nothing` covers it.
- **`concat()` zero-arg returning `""`** → covered by the arity check.
- **`op_keyword` alternation order** → reordered to match `op` rule, comment updated.
- **Float→i64 truncation in 3 spots** → centralized in `f64_to_i64()`, used consistently.

One open Cursor Bugbot item: when `@.*` produces a NodeList as the LHS of `subsetof`/`anyof`/`noneof`, `set_relate` checks if each node is a scalar *member* of the RHS rather than applying any-of per-node set semantics — so `@.* subsetof [1,2,3]` on an object with array values doesn't behave symmetrically with `sizeof`/`empty` (which got the any-of fix). Corner case for sure, worth a follow-up ticket but not blocking.

Let CI finish green before merge. Nice work.

---

PR #1604, opened 2026-06-17
MOD-16274 | MOD-16275 Extended JSONPath: top-level projection express…

AvivDavid23:
…ions
Summary

Adds support for projection expressions at the top level of a JSONPath query, so JSON.GET can return a computed value rather than only selecting document nodes. The arithmetic/function evaluator built for filter expressions in #1602 (MOD-16274) is now reachable from the query root, plus new postfix method-call syntax.

```
JSON.GET key '$.a + 1'              -> [3]
JSON.GET key '$.a * $.b'           -> [8]
JSON.GET key '($.a + $.b) / 2'     -> [3.0]
JSON.GET key '-$.a'                -> [-2]
JSON.GET key '$.arr.length()'      -> [3]
JSON.GET key 'length($.arr)'       -> [3]

```
What's added

Broad capability
- Top-level projection in JSON.GET ($-mode): the query root may be a computed expression, not just a path. Works in single-path, multi-path, RESP2 and RESP3 replies.
- Three call forms: infix arithmetic, postfix/method (`$.arr.length()`), and prefix (`length($.arr)`).
- Projections are read-only: only JSON.GET evaluates them; every other command (write and node-based read) rejects a projection path with a clear error.

Specific operators / functions usable at the top level (all already implemented for filters; now reachable as projections)
- Arithmetic: `+` `-` `*` `/` `%`, unary `+`/`-`, parentheses, precedence.
- Numeric: `abs`, `ceiling`, `floor`.
- Array: `min`, `max`, `avg`, `sum`, `stddev`, `length`, `first`, `last`, `index`.
- String: `length`, `concat`, `match`, `search`.
- Nodelist: `count`, `value`.

---

> [!NOTE]
> **Medium Risk**
> Core query parsing and classification affect every JSON command path, though plain-path behavior is explicitly preserved; projections are gated to read commands only, reducing mutation risk.
> 
> **Overview**
> **JSONPath** now parses the query root as an `arith_expr` instead of only `$` + path segments. Lone `$.path` queries still classify as plain paths; anything with operators, unary signs, `@`, prefix calls, or **postfix** `.name(...)` is a read-only **projection** evaluated via `eval_projection` / `calc_once_projection`.
> 
> Grammar adds `postfix_method` / `method_chain` and stops `.func(` from being parsed as a field segment. `=~` can use computed string patterns via `term_as_str`. `compile()` enforces a **128**-level nesting cap (ignoring brackets inside quoted strings). Parenthesized lone paths like `($..x)` still classify as paths to avoid double-wrapped arrays.
> 
> **redis_json** routes projections through **JSON.GET**, **JSON.MGET**, and **JSON.RESP** (including legacy-normalized forms such as `a + 1`); all other path commands return `ERR computed/projection expressions are only supported by JSON.GET/JSON.MGET/JSON.RESP`. Legacy `Path::is_jsonpath` is extended so spaced/unary/prefix forms normalize correctly without breaking fields like `-$x` or `$.a+1`.
> 
> The JSON **LLAPI** rejects projection paths on parse and get.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0868dadd2b9287deffb348bbb05e0f2adbbba10e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/json_path/src/grammar.pest
+++ b/json_path/src/grammar.pest
@@ -80,7 +80,14 @@ mul_op = _{mul | div | rem}
 neg = @{"-"}
 pos = @{"+"}
 unary_op = _{neg | pos}
-arith_primary = _{"(" ~ arith_expr ~ ")" | term}
+// Postfix (method) function call applied to a primary, e.g. `$arr.length()`,
+// `$.a.index(2)`. Disambiguated from a `.field` path segment by the immediately
+// following `(` (which is not a `literal` char). Args are full expressions.
+postfix_method = {"." ~ function_name ~ "(" ~ (arith_expr ~ ("," ~ arith_expr)*)? ~ ")"}
+// A term followed by >=1 method call. The `+` keeps a bare term out of this rule so
+// existing (non-method) parse trees are unchanged.
+method_chain = {term ~ postfix_method+}
+arith_primary = _{"(" ~ arith_expr ~ ")" | method_chain | term}
 arith_factor = {unary_op? ~ arith_primary}
 arith_term = {arith_factor ~ (mul_op ~ arith_factor)*}
 arith_expr = {arith_term ~ (add_op ~ arith_term)*}
@@ -104,7 +111,11 @@ full_scan = {".."}
 
 bracket = _{numbers_range | numbers_list | string_list | all | "?" ~ filter}
 
-element = _{"." ~ literal | full_scan ~ (literal | all | "[" ~ bracket ~ "]") | "." ~ all | ("."? ~ "[" ~ bracket ~ "]")}
+// A `.`-field segment must NOT be a method call (a name immediately followed by `(`),
+// so a trailing `.func()` is left for `postfix_method` instead of being swallowed as a
+// field. `.field` with no `(` is unaffected, so a field literally named `length` still
+// parses as a path segment.
+element = _{"." ~ !(function_name ~ "(") ~ literal | full_scan ~ (literal | all | "[" ~ bracket ~ "]") | "." ~ all | ("."? ~ "[" ~ bracket ~ "]")}
 
 // The bare word operators must not be swallowed as a bare first path segment
 // (e.g. `@ in [...]`). A field with one of these names is still reachable via
@@ -122,7 +133,10 @@ root = {(element|first_element) ~ (element)*}
 
 simple_root = {"." ~ literal}
 
-query = _{SOI ~ "$" ~ root? ~ EOI}
+// Top level is an arithmetic expression. A lone `$.path` is a valid `arith_expr` (a single
+// `from_root` term, no operator/method) and reproduces the former `"$" ~ root?` exactly;
+// `compile()` classifies such a lone path as path-mode and anything else as a projection.
+query = _{SOI ~ arith_expr ~ EOI}
 
 simple_query = _{SOI ~ "$" ~ simple_root ~ EOI}
 
--- a/json_path/src/json_path.rs
+++ b/json_path/src/json_path.rs
@@ -734,8 +734,8 @@ fn eval_function<'i, 'j, S: SelectValue>(
     let arity_ok = match name {
         "concat" => !args.is_empty(),
         "index" => args.len() == 2,
-        "ceiling" | "floor" | "abs" | "sum" | "min" | "max" | "avg" | "stddev" | "first"
-        | "last" => args.len() == 1,
+        "length" | "count" | "ceiling" | "floor" | "abs" | "sum" | "min" | "max" | "avg"
+        | "stddev" | "first" | "last" => args.len() == 1,
         _ => true,
     };
     if !arity_ok {
@@ -858,7 +858,12 @@ fn classify_query<'i>(expr: &Pair<'i, Rule>, empty: &Pairs<'i, Rule>) -> QueryCl
             Some(root) => QueryClass::Path(root.into_inner()),
             None => QueryClass::Path(empty.clone()),
         },
-        // from_current (`@`), method_chain, function_call, literals, parenthesized expr.
+        // A fully-parenthesized expression with no surrounding operator/sign/method reduces to
+        // its inner expression: `($..x)` is the same query as `$..x`. Recurse so a wrapped lone
+        // path classifies as a path — otherwise the projection serializer collapses a multi-node
+        // result into one array and then wraps it again (`($..x)` -> `[[..]]` instead of `[..]`).
+        Rule::arith_expr => classify_query(&prim, empty),
+        // from_current (`@`), method_chain, function_call, literals.
         _ => QueryClass::Projection,
     }
 }
@@ -866,6 +871,24 @@ fn classify_query<'i>(expr: &Pair<'i, Rule>, empty: &Pairs<'i, Rule>) -> QueryCl
 /// Compile the given string query into a query object.
 /// Returns error on compilation error.
 pub(crate) fn compile(path: &str) -> Result<Query<'_>, QueryCompilationError> {
+    const MAX_NESTING_DEPTH: usize = 128;
+    let (mut depth, mut max_depth) = (0usize, 0usize);
+    for b in path.bytes() {
+        match b {
+            b'(' | b'[' => {
+                depth += 1;
+                max_depth = max_depth.max(depth);
+            }
+            b')' | b']' => depth = depth.saturating_sub(1),
+            _ => {}
+        }
+        if max_depth > MAX_NESTING_DEPTH {
+            return Err(QueryCompilationError {
+                location: 0,
+                message: format!("JSONPath nesting too deep (max depth {MAX_NESTING_DEPTH})"),
+            });
+        }
+    }
     let query = JsonPathParser::parse(Rule::query, path);
     match query {
         Ok(mut q) => {
--- a/json_path/src/json_path.rs
+++ b/json_path/src/json_path.rs
@@ -110,6 +110,12 @@ pub struct Query<'i> {
     pub root: Pairs<'i, Rule>,
     is_static: Option<bool>,
     size: Option<usize>,
+    /// For a projection query (e.g. `$.a + 1`, `$arr.length()`) this holds the top-level
+    /// `arith_expr` to evaluate; `None` for a plain path. A projection is read-only (it has
+    /// no document path) and is evaluated via `PathCalculator::eval_projection` — `root` is
+    /// left empty and unused.
+    #[allow(dead_code)]
+    projection: Option<Pair<'i, Rule>>,
 }
 
 #[derive(Debug)]
@@ -184,6 +190,19 @@ impl<'i> Query<'i> {
         self.size = Some(size);
         self.is_static = Some(is_static);
         is_static
+    }
+
+    /// Whether this query is a projection (a computed expression such as `$.a + 1` or
+    /// `$arr.length()`) rather than a plain path. Projections are read-only.
+    #[allow(dead_code)]
+    pub fn is_projection(&self) -> bool {
+        self.projection.is_some()
+    }
+
+    /// The top-level `arith_expr` of a projection query, if any (for `eval_projection`).
+    #[allow(dead_code)]
+    pub(crate) fn projection_expr(&self) -> Option<&Pair<'i, Rule>> {
+        self.projection.as_ref()
     }
 }
 
@@ -771,21 +790,109 @@ fn eval_function<'i, 'j, S: SelectValue>(
     }
 }
 
+/// Convert a projection's evaluated `TermEvaluationResult` into the single output value as an
+/// impl-independent `serde_json::Value` (or `None` for Nothing, which becomes an empty result —
+/// like a non-matching path).
+#[allow(dead_code)]
+fn term_to_output<'i, 'j, S: SelectValue>(term: TermEvaluationResult<'i, 'j, S>) -> Option<Value> {
+    match term {
+        TermEvaluationResult::Integer(n) => Some(Value::from(n)),
+        TermEvaluationResult::Float(f) => serde_json::Number::from_f64(f).map(Value::Number),
+        TermEvaluationResult::Str(s) => Some(Value::from(s)),
+        TermEvaluationResult::String(s) => Some(Value::from(s)),
+        TermEvaluationResult::Bool(b) => Some(Value::from(b)),
+        TermEvaluationResult::Null => Some(Value::Null),
+        TermEvaluationResult::Literal(v) => Some(v),
+        // A real document node (e.g. `$.a.first()`, `value($.a)`): serialize it to a Value.
+        TermEvaluationResult::Value(vref) => {
+            Some(serde_json::to_value(&vref).unwrap_or(Value::Null))
+        }
+        // Multiple nodes (e.g. `($..x)`): render as a JSON array.
+        TermEvaluationResult::NodeList(list) => Some(Value::Array(
+            list.iter()
+                .map(|v| serde_json::to_value(v).unwrap_or(Value::Null))
+                .collect(),
+        )),
+        // Nothing -> empty result.
+        TermEvaluationResult::Invalid => None,
+    }
+}
+
+/// Result of classifying a top-level `arith_expr`.
+enum QueryClass<'i> {
+    /// A plain path; the inner value is the `root` segments to walk.
+    Path(Pairs<'i, Rule>),
+    /// A computed projection; evaluated via `eval_projection`.
+    Projection,
+}
+
+/// Classify a top-level `arith_expr`. It is a plain *path* iff it is a single bare
+/// `from_root` term: no arithmetic operator, no unary sign, no method call, not a function
+/// call, not parenthesized, and not `@`-rooted. Anything else is a projection. `empty` is a
+/// guaranteed-empty `Pairs` (the EOI subtree) reused as the `root` for a bare `$`.
+fn classify_query<'i>(expr: &Pair<'i, Rule>, empty: &Pairs<'i, Rule>) -> QueryClass<'i> {
+    let mut terms = expr.clone().into_inner(); // arith_term (add/sub arith_term)*
+    let Some(term) = terms.next() else {
+        return QueryClass::Projection;
+    };
+    if terms.next().is_some() {
+        return QueryClass::Projection; // a top-level + / - operator
+    }
+    let mut factors = term.into_inner(); // arith_factor (mul/div/rem arith_factor)*
+    let Some(factor) = factors.next() else {
+        return QueryClass::Projection;
+    };
+    if factors.next().is_some() {
+        return QueryClass::Projection; // a * / % operator
+    }
+    let mut inner = factor.into_inner(); // unary_op? arith_primary
+    let Some(prim) = inner.next() else {
+        return QueryClass::Projection;
+    };
+    if matches!(prim.as_rule(), Rule::neg | Rule::pos) {
+        return QueryClass::Projection; // unary +/-
+    }
+    match prim.as_rule() {
+        // A lone `$` / `$.path`: walk its `root` segments (empty for a bare `$`).
+        Rule::from_root => match prim.into_inner().next() {
+            Some(root) => QueryClass::Path(root.into_inner()),
+            None => QueryClass::Path(empty.clone()),
+        },
+        // from_current (`@`), method_chain, function_call, literals, parenthesized expr.
+        _ => QueryClass::Projection,
+    }
+}
+
 /// Compile the given string query into a query object.
 /// Returns error on compilation error.
 pub(crate) fn compile(path: &str) -> Result<Query<'_>, QueryCompilationError> {
     let query = JsonPathParser::parse(Rule::query, path);
     match query {
         Ok(mut q) => {
-            let root = q.next().ok_or_else(|| QueryCompilationError {
+            let expr = q.next().ok_or_else(|| QueryCompilationError {
                 location: 0,
                 message: "internal: empty JSONPath parse result".to_string(),
             })?;
-            Ok(Query {
-                root: root.into_inner(),
-                is_static: None,
-                size: None,
-            })
+            // EOI follows `arith_expr`; its (empty) inner is reused as the empty `root` for
+            // a bare `$` and for projection queries.
+            let empty = q
+                .next()
+                .map(Pair::into_inner)
+                .unwrap_or_else(|| expr.clone().into_inner());
+            match classify_query(&expr, &empty) {
+                QueryClass::Path(root) => Ok(Query {
+                    root,
+                    is_static: None,
+                    size: None,
+                    projection: None,
+                }),
+                QueryClass::Projection => Ok(Query {
+                    root: empty,
+                    is_static: None,
+                    size: None,
+                    projection: Some(expr),
+                }),
+            }
         }
         // pest::error::Error
         Err(e) => {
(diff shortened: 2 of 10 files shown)


json_path/src/json_path.rs
```diff
@@ -771,21 +790,109 @@ fn eval_function<'i, 'j, S: SelectValue>(
...
-                root: root.into_inner(),
-                is_static: None,
-                size: None,
-            })
+            // EOI follows `arith_expr`; its (empty) inner is reused as the empty `root` for
+            // a bare `$` and for projection queries.
+            let empty = q
+                .next()
+                .map(Pair::into_inner)
+                .unwrap_or_else(|| expr.clone().into_inner());
```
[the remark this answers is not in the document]
AvivDavid23: in practice, `unwrap_or_else(|| expr.clone().into_inner())` fallback is never reached. (If EOI weren't emitted, empty would be the fallback = an arith_term, and root_only would already fail — it doesn't.
But will make such case return a valid error

json_path/src/json_path.rs
```diff
@@ -1316,25 +1428,12 @@ impl<'i, 'j, S: SelectValue> TermEvaluationResult<'i, 'j, S> {
...
-                    (_, _) => false,
-                }
-            }
-            (_, _) => false,
+        // Both operands must be strings. `term_as_str` normalizes every string-shaped term
+        // (document string, `Str`/`String` literal, single-string nodelist), so a computed
+        // pattern such as `@.s =~ concat(@.a, @.b)` matches like a literal one.
+        match (term_as_str(self), term_as_str(s)) {
+            (Some(subject), Some(regex)) => Self::re_is_match(cache, regex, subject),
+            _ => false,
```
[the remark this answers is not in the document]
AvivDavid23: Not a bug — the operand roles aren't swapped. =~ is positional: LHS = subject, RHS = regex, and that's preserved exactly from the pre-refactor code:

- Old (Value, Str) → subject = LHS value, regex = RHS literal.
- Old (Value, Value) → re_is_match(cache, s2, s1), i.e. regex = RHS (v2), subject = LHS (v1).

The refactor keeps the same assignment — re_is_match(cache, regex /*RHS*/, subject /*LHS*/) — and just lets each operand be any string-shaped term via term_as_str (so a computed pattern like `@.s =~ concat(@.a, @.b) `works, which is the point of the change)

json_path/src/grammar.pest
```diff
@@ -122,7 +133,10 @@ root = {(element|first_element) ~ (element)*}
 
 simple_root = {"." ~ literal}
 
-query = _{SOI ~ "$" ~ root? ~ EOI}
+// Top level is an arithmetic expression. A lone `$.path` is a valid `arith_expr` (a single
+// `from_root` term, no operator/method) and reproduces the former `"$" ~ root?` exactly;
+// `compile()` classifies such a lone path as path-mode and anything else as a projection.
+query = _{SOI ~ arith_expr ~ EOI}
```
gabsow, line 139: **Correctness (high) — top-level deep nesting can crash the server.** Changing `query` to `SOI ~ arith_expr ~ EOI` exposes the recursive `arith_primary = "(" ~ arith_expr ~ ")"` rule at the query top level. `JSON.GET key '(((((…$.a…)))))'` with a few thousand nested parens overflows the parser stack and aborts the process (SIGABRT). Reachable through `Path::new` from the primary path of every JSON command (`is_jsonpath` returns true on a leading `(`, so it is compiled verbatim). The unbounded recursion pre-dates this PR (filters already recurse on `[?((( … )))]`), but this change makes it reachable without a filter — worth a parse depth/length guard on untrusted input.

redis_json/src/redisjson.rs
```diff
@@ -132,6 +130,36 @@ impl<'a> Path<'a> {
...
+            // Unary `-`/`+` on a rooted or parenthesized operand (e.g. `-$.a`, `+($.a)`),
+            // distinct from a legacy field name that happens to contain a sign.
+            Some(b'-' | b'+') => matches!(b.get(1), Some(b'$' | b'(')),
+            // Prefix function call: an identifier immediately followed by `(`.
+            Some(c) if c.is_ascii_alphabetic() || *c == b'_' => {
+                let name_len = b
+                    .iter()
+                    .take_while(|x| x.is_ascii_alphanumeric() || **x == b'_')
+                    .count();
+                b.get(name_len) == Some(&b'(')
```
gabsow: **Correctness (high) — `is_jsonpath` is whitespace-blind, the grammar is not → silent wrong answer.** This checks the byte *immediately* after a function name for `(`, but `WHITESPACE = _{" "}` lets the grammar accept spaces. So `JSON.GET doc 'length ( $.arr )'` is misclassified as legacy, normalized to `$.length ( $.arr )`, and parsed as a *method chain on the whole document* — returning the document key-count (e.g. `4`) instead of the array length `3`. `- $.a` likewise becomes the unparseable `$.- $.a`. The no-space form `length($.arr)` is fine, so the trigger is whitespace inside the call / after a unary sign.

redis_json/src/redisjson.rs
```diff
@@ -132,6 +130,36 @@ impl<'a> Path<'a> {
...
+    fn is_jsonpath(path: &str) -> bool {
+        let b = path.as_bytes();
+        match b.first() {
+            // Rooted JSONPath (legacy keeps `$<name>` as a field).
+            Some(b'$') => path.len() < 2 || matches!(b.get(1), Some(b'.' | b'[')),
+            // Parenthesized projection, e.g. `($.a + $.b) / 2`.
+            Some(b'(') => true,
+            // Unary `-`/`+` on a rooted or parenthesized operand (e.g. `-$.a`, `+($.a)`),
+            // distinct from a legacy field name that happens to contain a sign.
+            Some(b'-' | b'+') => matches!(b.get(1), Some(b'$' | b'(')),
```
gabsow: **Correctness (high) — legacy fields named `-$…`/`+$…` are silently reinterpreted.** `Some(b'-' | b'+') => matches!(b.get(1), Some(b'$' | b'('))` reclassifies a path like `-$x` from a legacy field access into the projection `-($.x)`. On `{"-$x":111,"x":5}`: `JSON.GET key '-$x'` returns `[-5]` (was `[111]`), and every write (`JSON.SET`/`JSON.DEL` `key '-$x' …`) now fails with `err_projection_readonly` instead of mutating that field. Niche key names, but a silent data-access regression vs 1.x.

json_path/src/json_path.rs
```diff
@@ -771,21 +790,114 @@ fn eval_function<'i, 'j, S: SelectValue>(
...
 }
 
+/// Convert a projection's evaluated `TermEvaluationResult` into the single output value as an
+/// impl-independent `serde_json::Value` (or `None` for Nothing, which becomes an empty result —
+/// like a non-matching path).
+#[allow(dead_code)]
+fn term_to_output<'i, 'j, S: SelectValue>(term: TermEvaluationResult<'i, 'j, S>) -> Option<Value> {
+    match term {
+        TermEvaluationResult::Integer(n) => Some(Value::from(n)),
+        TermEvaluationResult::Float(f) => serde_json::Number::from_f64(f).map(Value::Number),
```
gabsow, line 800: **Correctness (medium) — projection arithmetic loses precision on integers above i64::MAX.** *(Root cause is the number typing at `json_path/src/json_node.rs:26`, which isn't in this diff — flagging at the projection output here.)* A `u64 > i64::MAX` is typed `Double`, so the arithmetic evaluator reads it as `u as f64`. On `{"big":18446744073709551615}`: `JSON.GET doc '$.big'` returns the exact integer, but `JSON.GET doc '$.big + 0'` returns `1.8446744073709552e19`. The projection output and the plain path disagree for the same stored number.

json_path/src/json_path.rs
```diff
@@ -771,21 +790,114 @@ fn eval_function<'i, 'j, S: SelectValue>(
...
+        TermEvaluationResult::String(s) => Some(Value::from(s)),
+        TermEvaluationResult::Bool(b) => Some(Value::from(b)),
+        TermEvaluationResult::Null => Some(Value::Null),
+        TermEvaluationResult::Literal(v) => Some(v),
+        // A real document node (e.g. `$.a.first()`, `value($.a)`): serialize it to a Value.
+        TermEvaluationResult::Value(vref) => {
+            Some(serde_json::to_value(&vref).unwrap_or(Value::Null))
+        }
+        // Multiple nodes (e.g. `($..x)`): render as a JSON array.
+        TermEvaluationResult::NodeList(list) => Some(Value::Array(
```
gabsow, line 811: **Correctness (medium) — a parenthesized multi-node projection double-wraps.** The `NodeList` arm collapses a multi-match into one `Value::Array`, then the projection serializer wraps it again. On `{"a":{"x":1},"b":{"x":2}}`: `$..x` → `[1,2]` but `($..x)` → `[[1,2]]`. RESP3 diverges further — `($..x)` returns a single bulk-string `"[1,2]"` (via `value_to_resp3`'s catch-all) instead of two integers. No test covers a parenthesized multi-node projection.

json_path/src/json_path.rs
```diff
@@ -1316,25 +1428,12 @@ impl<'i, 'j, S: SelectValue> TermEvaluationResult<'i, 'j, S> {
...
-                        .zip(v2.as_str())
-                        .is_some_and(|(s1, s2)| Self::re_is_match(cache, s2, s1)),
-                    (_, _) => false,
-                }
-            }
-            (_, _) => false,
+        // Both operands must be strings. `term_as_str` normalizes every string-shaped term
+        // (document string, `Str`/`String` literal, single-string nodelist), so a computed
+        // pattern such as `@.s =~ concat(@.a, @.b)` matches like a literal one.
+        match (term_as_str(self), term_as_str(s)) {
```
gabsow, line 1474: **Correctness (medium) — `=~` now matches a string-literal left operand.** The rewrite runs `term_as_str` on *both* operands; the old code only matched when the LHS was a document `Value` (and returned false otherwise). So `$.items[?"abc" =~ @.pat]` goes from matching 0 nodes to matching wherever `@.pat` is a substring of `"abc"`. Subject/regex order is correctly preserved — this is purely the additive literal-LHS case, a behavior change to existing filters.

json_path/src/json_path.rs
```diff
@@ -1811,10 +1910,39 @@ impl<'i, UPTG: UserPathTrackerGenerator> PathCalculator<'i, UPTG> {
...
+        let mut acc = self.evaluate_single_term(recv, json.clone(), calc_data);
+        for method in it {
+            let mut mi = method.into_inner();
+            let name = mi.next().map_or("", |p| p.as_str());
+            // The receiver is the implicit first argument; explicit args follow.
+            let mut args = vec![acc];
+            for arg in mi {
+                args.push(self.evaluate_arith_operand(arg, json.clone(), calc_data));
+            }
+            acc = eval_function(name, args, &mut calc_data.regex_cache);
```
gabsow, line 1981: **Correctness (medium) — `length`/`count` aren't arity-checked, so extra args are silently dropped.** *(Fix is in `eval_function`'s `arity_ok` at `json_path/src/json_path.rs:734`, just outside this hunk.)* `length`/`count` fall into `_ => true` (only `value` is guarded by `args.len() == 1`). Combined with this method dispatch (receiver = implicit first arg), `$.length($.arr)` computes `length(<document>)` and ignores `$.arr` → returns the key-count `4`; `$.arr.length(99)` silently drops `99`. Add `"length" | "count" => args.len() == 1` to `arity_ok`.

redis_json/src/manager.rs
```diff
@@ -102,6 +102,12 @@ pub fn err_invalid_path() -> RedisError {
     RedisError::Str("ERR Path does not exist")
 }
 
+pub fn err_projection_readonly() -> RedisError {
+    RedisError::Str(
+        "ERR computed/projection expressions are only supported by JSON.GET/JSON.MGET/JSON.RESP",
```
gabsow, line 107: **Correctness (low) — legacy-form projections return `nil` instead of this error from TYPE/OBJLEN/OBJKEYS.** A legacy path like `a + 1` normalizes to the projection `$.a + 1` and routes to the legacy branch, where `json_type_legacy` (`commands.rs:1216`, `.map_or(Null,…)`), `json_obj_keys_legacy` (`commands.rs:2765`, `_ => Null`), and `obj_len` (`key_value.rs:512`, `_ => Ok(NoneExisting)`) all swallow this `err_projection_readonly` → silent `nil`. The `$.a + 1` jsonpath form *is* rejected (which `testProjectionRejectedByAllPathCommands` checks), so the legacy form's gap is untested. No mutation/crash — a contract inconsistency.

redis_json/src/key_value.rs
```diff
@@ -188,7 +244,11 @@ impl<'a, V: SelectValue + 'a> KeyValue<'a, V> {
 
     pub fn to_resp3_path(&self, path: &Path, format: &ReplyFormatOptions) -> RedisValue {
         compile(path.get_path()).map_or(RedisValue::Array(vec![]), |q| {
```
gabsow: **Correctness (low) — RESP3 swallows malformed-projection parse errors.** `compile(...).map_or(RedisValue::Array(vec![]), …)` masks a compile failure, so under RESP3 `JSON.GET doc '$.a +'` returns `[]` while under RESP2 the same command errors (as `testProjectionEdgeCases` asserts). Pre-existing error-swallow for paths, now also hides malformed projection syntax from RESP3 clients.

redis_json/src/redisjson.rs
```diff
@@ -132,6 +130,36 @@ impl<'a> Path<'a> {
...
 
+    /// Whether `path` is a JSONPath (v2) expression rather than a legacy (1.x) path.
+    /// Rooted JSONPath is `$` alone or `$.`/`$[` — `$<name>` stays a legacy field (e.g. a
+    /// key literally named `$a`), preserving 1.x behavior. A top-level *projection*
+    /// expression is also JSONPath: a parenthesized group, a unary sign on a
+    /// `$`-rooted/parenthesized operand, or a prefix function call (`name(...)`). These
+    /// forms can only be computed expressions, never a legacy path (`(` is not a legacy
+    /// path char and a bare field is not followed by `(`), so legacy paths starting with
+    /// `.` or a bare field name are never misclassified.
+    fn is_jsonpath(path: &str) -> bool {
```
gabsow, line 141: **Altitude — `is_jsonpath` re-derives a classification the parser already owns.** This hand-rolled byte scan (rooted `$`, leading `(`, unary `±$`/`±(`, `name(`) duplicates what the pest grammar + `classify_query` decide, and the two drift apart — the root cause of the whitespace bug and the `-$…` reclassification flagged elsewhere, plus: `_f(…)` is accepted here (wider than `function_name = ASCII_ALPHA …`) then rejected by the grammar, and a top-level `@.a` (which the grammar accepts) is mangled to `$.@.a` → parse error. Consider classifying from the parse result (normalize only when raw `compile` fails or yields a non-path) instead of pre-guessing with bytes.

redis_json/src/key_value.rs
```diff
@@ -201,8 +261,12 @@ impl<'a, V: SelectValue + 'a> KeyValue<'a, V> {
         let res = if is_legacy {
             self.to_string_single(path, format)?.into()
         } else if format.is_resp3_reply() {
-            let values = self.get_values(path)?;
-            Self::values_to_resp3(&values, format)
+            let query = compile(path)?;
+            if query.is_projection() {
```
gabsow, line 272: **Simplification — dead RESP3 branches gained projection handling that can never run.** `to_json` routes *all* RESP3 replies through `to_resp3`, calling `to_json_single`/`to_json_multi` only on the non-RESP3 leg. So this `if query.is_projection()` block inside `to_json_single`'s `else if format.is_resp3_reply()` arm — and the equivalent `Values::Computed => projection_to_resp3` arm in `to_json_multi`'s `if format.is_resp3_reply()` block (~line 216) — are both unreachable. The live RESP3 single-path route is `to_resp3_path`.

redis_json/src/key_value.rs
```diff
@@ -101,8 +116,39 @@ impl<'a, V: SelectValue + 'a> KeyValue<'a, V> {
...
+        if query.is_projection() {
+            return Err(err_projection_readonly());
+        }
+        Ok(calc_once(query, self.val.as_ref()))
+    }
+
+    /// Serialize a projection result (0 or 1 value) as a RESP3 array, matching JSONPath
+    /// nodelist output. The value is a `serde_json::Value` (itself a `SelectValue`), reusing
+    /// the node serializer via `KeyValue::<Value>`.
+    fn projection_to_resp3(value: Option<Value>, format: &ReplyFormatOptions) -> RedisValue {
```
gabsow, line 129: **Reuse/simplification — projection plumbing is heavily duplicated.** `projection_to_resp3` is `values_to_resp3` over a 0-or-1 `Option<Value>`; `projection_to_string` is a one-line delegate to `serialize_object`. The `compile(path)?; if is_projection { calc_once_projection } else { calc_once }` dispatch is repeated across ~6 read sites, and the `if is_projection { return Err(err_projection_readonly()) }` guard across ~5 write/node sites. A `compile_node_path()` helper (compile + reject projections) plus one classified-read helper would centralize both and make the read-only invariant impossible to omit on a future command.

redis_json/src/key_value.rs
```diff
@@ -370,13 +440,30 @@ impl<'a, V: SelectValue + 'a> KeyValue<'a, V> {
...
+            .into_iter()
+            .next()
+            .ok_or_else(err_invalid_path)?;
         Self::serialize_object(&result, format)
     }
 
     pub fn to_string_multi(&self, path: &str, format: &ReplyFormatOptions) -> RedisResult<String> {
-        let results = self.get_values(path)?;
-        Self::serialize_object(&results, format)
+        let query = compile(path)?;
```
gabsow, line 468: **Efficiency — JSON.MGET recompiles the same path once per key.** `json_mget_command_impl` loops over N keys calling `to_string_multi`/`to_string_single`, each of which runs `compile(path)` on the identical path string → N full pest parses instead of one. Compile once in the command layer and pass the `Query` down (the value varies per key, the path is constant).

json_path/src/json_path.rs
```diff
@@ -771,21 +790,114 @@ fn eval_function<'i, 'j, S: SelectValue>(
...
+    Path(Pairs<'i, Rule>),
+    /// A computed projection; evaluated via `eval_projection`.
+    Projection,
+}
+
+/// Classify a top-level `arith_expr`. It is a plain *path* iff it is a single bare
+/// `from_root` term: no arithmetic operator, no unary sign, no method call, not a function
+/// call, not parenthesized, and not `@`-rooted. Anything else is a projection. `empty` is a
+/// guaranteed-empty `Pairs` (the EOI subtree) reused as the `root` for a bare `$`.
+fn classify_query<'i>(expr: &Pair<'i, Rule>, empty: &Pairs<'i, Rule>) -> QueryClass<'i> {
```
gabsow, line 833: **Altitude — `classify_query` re-walks the parse tree by position.** It unwinds `arith_term → arith_factor → arith_primary` via positional `into_inner().next()` to recover 'lone path vs projection.' Any future precedence/wrapping change to the arithmetic rules silently breaks classification with no compile error. A dedicated grammar alternative (e.g. `query = path_query | projection_query`) would let the parser tag the node directly and delete this re-walk.

tests/pytest/test.py
```diff
@@ -1764,6 +1764,141 @@ def testFilterArithmetic(env):
...
+    r.assertEqual(json.loads(r.execute_command('JSON.GET', 'doc', '$')), [doc])
+    r.expect('PING').equal(True)
+    r.expect('JSON.GET', 'doc', p).equal('[3]')
+
+def testProjectionResp(env):
+    # JSON.RESP is a value-returning read command, so it evaluates a projection (like GET/MGET),
+    # returning the computed value resp-encoded (or an empty reply for a Nothing projection).
+    r = env
+    r.expect('JSON.SET', 'doc', '$', json.dumps({"a": 2, "arr": [1, 2, 3]})).ok()
+    r.assertEqual(r.execute_command('JSON.RESP', 'doc', '$.a + 1'), [3])
```
gabsow, line 1859: **Test gap — the live RESP3 path and structured/float projection outputs are untested.** Every `testProjection*` test runs under RESP2 and asserts string replies; none enable RESP3 or `FORMAT EXPAND` (the only *live* RESP3 projection route, `to_resp3_path` → `projection_to_resp3`). `testProjectionResp` checks only scalar integers, so `JSON.RESP` over a projection that returns an array/object (`first()`/`value()`) or a float (`$.nums.min()` → `1.0`) is unverified — a bug in `projection_to_resp` → `resp_serialize_inner` over `Value::Array`/float would pass CI.

json_path/src/json_path.rs
AvivDavid23: check previous response to Cursor bugbot

json_path/src/json_path.rs
AvivDavid23: kmown issue, will need separate handling

json_path/src/json_path.rs
```diff
@@ -771,21 +790,137 @@ fn eval_function<'i, 'j, S: SelectValue>(
...
+        _ => QueryClass::Projection,
+    }
+}
+
 /// Compile the given string query into a query object.
 /// Returns error on compilation error.
 pub(crate) fn compile(path: &str) -> Result<Query<'_>, QueryCompilationError> {
+    const MAX_NESTING_DEPTH: usize = 128;
+    let (mut depth, mut max_depth) = (0usize, 0usize);
+    for b in path.bytes() {
```
gabsow: **Correctness — regression from the new nesting-depth guard: brackets inside string literals are counted, so valid queries are falsely rejected.**

This byte scan counts every `(`/`[` toward depth, including those inside **string literals** and **bracket-quoted field names** — it never skips quoted spans. A well-formed query with >128 net-unbalanced opening brackets inside a literal now errors. Verified on this commit (built `json_path`, called `compile()` directly):

```
compile  $.a[?@.s == "(((…200 '('…"]   -> Err  (should be Ok)
compile  $["(((…200 '('…"]             -> Err  (should be Ok)
```

Most realistic trigger: a regex with deep grouping in `=~`/`match()` (e.g. `@.s =~ "((((…))))"`), or a field name literally containing many parens. Balanced brackets in a literal (`"()()()"`) are fine — only a net imbalance >128 trips it.

Fix options: skip quoted spans while scanning (track an in-string flag for `'`/`"`), or enforce the limit via a parser recursion depth rather than a second byte heuristic (see the broader `is_jsonpath`-duplicates-the-grammar comment).

gabsow:
### Re: the two remaining review items (#4, #6) — non-blocking, your call

With the latest fixes (incl. the nesting-guard string-literal fix in `CR`), these are the only two correctness items I have left, and both are minor edge cases rather than bugs in normal use:

- **#4 — projection arithmetic on integers above `i64::MAX`** (`json_path/src/json_path.rs:800`, the `Float` arm of `term_to_output`; root typing is in `json_node.rs`). A `u64 > i64::MAX` is read as `f64`, so `$.big + 0` returns `1.84e19` while the plain path `$.big` returns the exact integer. Pre-existed in the filter evaluator; this PR just makes it user-visible in projection output.
- **#6 — `=~` with a string-*literal* left operand** (`json_path/src/json_path.rs:1470`, `re_match` via `term_as_str`). `$.items[?"abc" =~ @.pat]` now matches where it previously matched nothing. This is arguably a *fix* (no real query relies on a filter that always returns empty), so I'd lean toward leaving it.

I don't think either should block this PR. **If you'd prefer to leave them as known limitations (or track #4 as a follow-up), that's fine by me — no need to fix here, and I'm happy to approve once CI is green.** If you'd rather patch #4, the targeted fix is to carry the `u64` through the arithmetic instead of widening to `f64`.

gabsow (approved):

PR #1607, opened 2026-06-22
MOD-16274 | MOD-16275 Extended JSONPath — `keys()`/`~` get-keys and `append()`

AvivDavid23:
Adds the two remaining MOD-16275 extended-JSONPath features on top of the top-level
**projection** machinery (#1604): the object **get-keys** capability (`~` operator + `keys()`
alias) and **`append()`** reply-enrichment.

### Get-keys: `~` / `keys()`

Emit an object's member names (one capability, two syntaxes):

JSON.GET k '$.obj~'         -> ["x","y"]      (alias: '$.obj.keys()')
JSON.GET k '$~'             -> ["a","b"]      (root keys)

- A non-object / missing / empty receiver → `[]`.
- `~` is **terminal** and attaches only to a bare term: `$.obj~`/`$~` work, while
  `$.obj.keys()~`, `$.obj~.x`, `$.obj~~`, `$.obj~.length()` are all parse errors. Use the
  `keys()` function form to compose (e.g. `$.obj.keys().count()`).

### `append(x)`

Enrich the **reply** with `x` **without modifying the document**:

JSON.GET k '$.books[?@.price >= 10].append({"min_price":10})'  -> [ …matched books…, {"min_price":10} ]
JSON.GET k '$.arr.append(9)'                                    -> [1,2,3,9]

- **Multi-result receiver** (filter, `[*]`) → matched nodes + `x`.
- **Single array receiver** → appended *into* the array.
- **Single non-array node** → appended alongside (`[node, x]`); no match → `[x]`.
- `x` is appended as **one** element: a multi-value `x` (a multi-node path or a synthesized
  list) is wrapped — `$.arr.append($.obj.keys())` → `[1,2,3,["k1","k2"]]`.
- A **Nothing** `x` propagates → `$.arr.append($.missing)` → `[]`.

Both are **read-only**: mutating / node commands (`JSON.SET`/`DEL`/`TYPE`/`OBJLEN`/…) reject
them via the existing projection guard. Supported by the value-returning reads
(`JSON.GET` / `JSON.MGET` / `JSON.RESP`), including RESP3 `FORMAT EXPAND`.

## Compatibility (breaking)

`~` is now the get-keys operator and **no longer a field-name character**. A key containing `~`
must use bracket notation — migrate dot paths: `$.a~b` → `$["a~b"]`. After upgrade the dot form
(`$.a~b`) is a parse error. Verified no path in the test corpus relied on this.

---

> [!NOTE]
> **Medium Risk**
> Breaking change reserves `~` for get-keys (dot paths with `~` in field names fail); projection output shape changes from optional single value to vectors, affecting JSON.GET/MGET/RESP paths.
> 
> **Overview**
> Adds extended JSONPath **get-keys** and **`append()`** as read-only projections on top of the existing projection machinery.
> 
> **Get-keys:** Parses a terminal **`~`** on a bare term (`$.obj~`, `$~`) as an alias for **`keys()`**, emitting member names as a flat string list. **`~` is removed from the `literal` charset** — dot paths like `$.a~b` no longer work; use `$["a~b"]` instead. **`~` is terminal** (no chaining after it); **`keys()`** remains composable (e.g. `$.obj.keys().count()`).
> 
> **`append(x)`:** Builds a **synthesized reply** (matched nodes or array elements plus `x`) **without mutating the document**. Multi-value arguments are wrapped as a single appended element; a Nothing `x` yields an empty result.
> 
> **Evaluation model:** Introduces **`TermEvaluationResult::Results`** for synthesized multi-value outputs (distinct from **`NodeList`**, which still serializes as one JSON array). Projections now return **`Vec<Value>`** via **`term_to_outputs`** / **`calc_once_projection`**; **`redis_json`** **`Values::Computed`** and RESP/JSON serialization follow the flat nodelist shape. **`Results`** participates in filters (`in`, set ops, comparisons any-of), **`count`/`length`/`value`**, indexing, and aggregations.
> 
> Extensive unit and integration tests cover parse errors, Redis GET/MGET/RESP, and projection read-only guards.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ba8cd1c6905f384c59fca5fb7ebc8e5932716ba1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/json_path/src/grammar.pest
+++ b/json_path/src/grammar.pest
@@ -1,4 +1,4 @@
-literal = @{('a' .. 'z' | 'A' .. 'Z' | '0' .. '9' | "-" | "_" | "`" | "~" | "+" | "#" | "%" | "$" | "^" | "/" | ":")+}
+literal = @{('a' .. 'z' | 'A' .. 'Z' | '0' .. '9' | "-" | "_" | "`" | "+" | "#" | "%" | "$" | "^" | "/" | ":")+}
 
 string = _{("'" ~ (string_value) ~ "'") | ("\"" ~ (string_value) ~ "\"") | ("\"" ~ (string_value_escape_1) ~ "\"") | ("'" ~ (string_value_escape_2) ~ "'")}
     string_escape = @{"\\"}
@@ -84,9 +84,13 @@ unary_op = _{neg | pos}
 // `$.a.index(2)`. Disambiguated from a `.field` path segment by the immediately
 // following `(` (which is not a `literal` char). Args are full expressions.
 postfix_method = {"." ~ function_name ~ "(" ~ (arith_expr ~ ("," ~ arith_expr)*)? ~ ")"}
-// A term followed by >=1 method call. The `+` keeps a bare term out of this rule so
-// existing (non-method) parse trees are unchanged.
-method_chain = {term ~ postfix_method+}
+// Terminal get-keys operator: `$.obj~` / `$~` emit the object's member names. The function
+// alias `$.obj.keys()` is just a `postfix_method`. `~` was removed from the `literal`
+// charset, so a field literally named with `~` is reachable only via `["a~b"]`.
+get_keys_op = {"~"}
+// A term followed by >=1 postfix method or a trailing `~`. The `+` keeps a bare term out of
+// this rule so existing (non-method) parse trees are unchanged.
+method_chain = {term ~ (postfix_method | get_keys_op)+}
 arith_primary = _{"(" ~ arith_expr ~ ")" | method_chain | term}
 arith_factor = {unary_op? ~ arith_primary}
 arith_term = {arith_factor ~ (mul_op ~ arith_factor)*}
--- a/json_path/src/grammar.pest
+++ b/json_path/src/grammar.pest
@@ -88,12 +88,12 @@ postfix_method = {"." ~ function_name ~ "(" ~ (arith_expr ~ ("," ~ arith_expr)*)
 // alias `$.obj.keys()` is just a `postfix_method`. `~` was removed from the `literal`
 // charset, so a field literally named with `~` is reachable only via `["a~b"]`.
 get_keys_op = {"~"}
-// Either a term + >=0 postfix methods ending in a single terminal `~`, or a term + >=1
-// postfix method. `~` is genuinely terminal: it closes the chain, so `$.obj~.length()`,
-// `$.obj~~` and `$.obj~.x` are all rejected (the `keys()` function form stays composable).
-// Both alternatives require a postfix or the `~`, keeping a bare term out so existing
+// Either a bare term + a single terminal `~`, or a term + >=1 postfix method. `~` attaches
+// only directly to a term (not after a method), so `$.obj~`/`$~` work while `$.obj.keys()~`,
+// `$.obj~.length()`, `$.obj~~` and `$.obj~.x` are all parse errors (use `keys()` to compose).
+// Both alternatives require the `~` or a postfix, keeping a bare term out so existing
 // (non-method) parse trees are unchanged.
-method_chain = {term ~ postfix_method* ~ get_keys_op | term ~ postfix_method+}
+method_chain = {term ~ get_keys_op | term ~ postfix_method+}
 arith_primary = _{"(" ~ arith_expr ~ ")" | method_chain | term}
 arith_factor = {unary_op? ~ arith_primary}
 arith_term = {arith_factor ~ (mul_op ~ arith_factor)*}
--- a/json_path/src/grammar.pest
+++ b/json_path/src/grammar.pest
@@ -88,9 +88,12 @@ postfix_method = {"." ~ function_name ~ "(" ~ (arith_expr ~ ("," ~ arith_expr)*)
 // alias `$.obj.keys()` is just a `postfix_method`. `~` was removed from the `literal`
 // charset, so a field literally named with `~` is reachable only via `["a~b"]`.
 get_keys_op = {"~"}
-// A term followed by >=1 postfix method or a trailing `~`. The `+` keeps a bare term out of
-// this rule so existing (non-method) parse trees are unchanged.
-method_chain = {term ~ (postfix_method | get_keys_op)+}
+// Either a term + >=0 postfix methods ending in a single terminal `~`, or a term + >=1
+// postfix method. `~` is genuinely terminal: it closes the chain, so `$.obj~.length()`,
+// `$.obj~~` and `$.obj~.x` are all rejected (the `keys()` function form stays composable).
+// Both alternatives require a postfix or the `~`, keeping a bare term out so existing
+// (non-method) parse trees are unchanged.
+method_chain = {term ~ postfix_method* ~ get_keys_op | term ~ postfix_method+}
 arith_primary = _{"(" ~ arith_expr ~ ")" | method_chain | term}
 arith_factor = {unary_op? ~ arith_primary}
 arith_term = {arith_factor ~ (mul_op ~ arith_factor)*}
--- a/json_path/src/json_path.rs
+++ b/json_path/src/json_path.rs
@@ -768,6 +768,25 @@ fn function_append<'i, 'j, S: SelectValue>(
     out.extend(term_to_outputs(x));
     TermEvaluationResult::Results(out)
 }
+const FN_LENGTH: &str = "length";
+const FN_COUNT: &str = "count";
+const FN_VALUE: &str = "value";
+const FN_CEILING: &str = "ceiling";
+const FN_FLOOR: &str = "floor";
+const FN_ABS: &str = "abs";
+const FN_CONCAT: &str = "concat";
+const FN_SUM: &str = "sum";
+const FN_MIN: &str = "min";
+const FN_MAX: &str = "max";
+const FN_AVG: &str = "avg";
+const FN_STDDEV: &str = "stddev";
+const FN_KEYS: &str = "keys";
+const FN_APPEND: &str = "append";
+const FN_FIRST: &str = "first";
+const FN_LAST: &str = "last";
+const FN_INDEX: &str = "index";
+const FN_MATCH: &str = "match";
+const FN_SEARCH: &str = "search";
 
 fn eval_function<'i, 'j, S: SelectValue>(
     name: &str,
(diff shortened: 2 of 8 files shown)


json_path/src/grammar.pest
```diff
@@ -1,4 +1,4 @@
-literal = @{('a' .. 'z' | 'A' .. 'Z' | '0' .. '9' | "-" | "_" | "`" | "~" | "+" | "#" | "%" | "$" | "^" | "/" | ":")+}
+literal = @{('a' .. 'z' | 'A' .. 'Z' | '0' .. '9' | "-" | "_" | "`" | "+" | "#" | "%" | "$" | "^" | "/" | ":")+}
```
gabsow, line 1: **Breaking change: `~` removed from the `literal` charset**

Removing `~` from `literal` silently breaks any existing client that stores keys containing `~` and accesses them via dot notation. For example, a document `{"a~b": 5}` previously reachable as `$.a~b` will now produce a parse error instead of `[5]`.

The PR description notes this ("a field literally named with `~` is reachable only via `["a~b"]`"), but there is no migration note or version gate. Clients using dot-notation paths to `~`-containing fields will break silently — they'll receive an error response on queries that previously succeeded, with no indication of what changed.

Suggest at minimum:
1. Documenting the breaking change in the release notes with a clear before/after migration path (`$.a~b` → `$["a~b"]`).
2. Verifying with a query against the existing test corpus to confirm no pre-existing paths rely on this.

---

json_path/src/grammar.pest
```diff
@@ -84,9 +84,16 @@ unary_op = _{neg | pos}
...
+// Terminal get-keys operator: `$.obj~` / `$~` emit the object's member names. The function
+// alias `$.obj.keys()` is just a `postfix_method`. `~` was removed from the `literal`
+// charset, so a field literally named with `~` is reachable only via `["a~b"]`.
+get_keys_op = {"~"}
+// Either a term + >=0 postfix methods ending in a single terminal `~`, or a term + >=1
+// postfix method. `~` is genuinely terminal: it closes the chain, so `$.obj~.length()`,
+// `$.obj~~` and `$.obj~.x` are all rejected (the `keys()` function form stays composable).
+// Both alternatives require a postfix or the `~`, keeping a bare term out so existing
+// (non-method) parse trees are unchanged.
+method_chain = {term ~ postfix_method* ~ get_keys_op | term ~ postfix_method+}
```
gabsow: **`$.obj.keys()~` silently returns empty instead of a parse error**

The first alternative `term ~ postfix_method* ~ get_keys_op` allows zero or more `postfix_method`s before `~`. This means `$.obj.keys()~` is syntactically valid: `term=$.obj`, `postfix_method*=.keys()`, `get_keys_op=~`.

At runtime, the eval loop first applies `.keys()` (giving `Results(["x","y"])`), then applies `~` which calls `function_keys(Some(Results([...])))`. Since `Results` doesn't match either `Value(Object)` or `Literal(Object)` in `function_keys`, it falls to `_ => None` → `Invalid`. The result is silently `[]`.

The comment above says `~` is "genuinely terminal" and rejects `$.obj~~`, but `$.obj.keys()~` is not caught. A user who mistakenly writes `$.obj.keys()~` instead of `$.obj.keys()` gets an empty result with no diagnostic.

One fix: add a grammar-level guard so `~` cannot follow a `postfix_method`. For example:
```pest
// Only allow ~ directly after a bare term (no preceding postfix methods)
get_keys_chain = {term ~ get_keys_op}
method_chain   = {term ~ postfix_method+}
```
Alternatively, add a test case that asserts `$.obj.keys()~` is a parse error (it currently isn't).

---

json_path/src/json_path.rs
```diff
@@ -475,6 +475,7 @@ fn function_length<'i, 'j, S: SelectValue>(
         | TermEvaluationResult::Float(_)
         | TermEvaluationResult::Bool(_)
         | TermEvaluationResult::Null
+        | TermEvaluationResult::Results(_)
```
gabsow: **`function_length` returns `Invalid` for `Results`, asymmetric with `count`**

This line places `Results(_)` in the `None` arm, so `$.obj.keys().length()` returns Nothing (empty `[]`). Meanwhile `function_count` correctly adds `Results(vs) => vs.len()` so `$.obj.keys().count()` returns the key count.

The asymmetry is surprising: `keys()` returns string values and users will naturally try `.length()` on the result to get the list size (mirroring how `.count()` is documented to work on nodelists). The current behaviour silently returns `[]` rather than erroring, making it hard to diagnose.

Options:
1. Make `function_length` handle `Results(vs) => Some(vs.len() as i64)` (consistent with `count`).
2. Or keep the existing behaviour but add a test and doc note: "use `.count()` not `.length()` on a synthesized list".

---

json_path/src/json_path.rs
```diff
@@ -723,6 +733,81 @@ fn function_index<'i, 'j, S: SelectValue>(
...
+                    .collect()
+            })
+        }
+        Some(TermEvaluationResult::Literal(Value::Array(items))) => items,
+        Some(TermEvaluationResult::Results(vs)) => vs,
+        Some(TermEvaluationResult::Invalid) | None => Vec::new(),
+        // A single non-array node / scalar: appended alongside as one element.
+        Some(other) => term_to_outputs(other),
+    };
+    out.extend(term_to_outputs(x));
```
gabsow: **`function_append` silently drops `x` when the appended value is absent**

`function_append` is called via `eval_function` which always passes `Some(...)` for both args (the arity check ensures two args). This means `x` is always `Some(TermEvaluationResult)`, never `None`. When the append argument is a non-matching path (e.g. `$.missing_field`), `x` is `Some(Invalid)` — it passes the `let Some(x) = x else { ... }` guard.

`term_to_outputs(Invalid)` returns `vec[]`, so `out.extend([])` is a no-op. The result is `Results([...receiver_elements])` with nothing appended and no error.

**Concrete example:** `$.arr.append($.missing_key)` on `{"arr":[1,2,3]}` silently returns `[1,2,3]` instead of signalling that the append argument was absent.

This is undocumented. The PR description says "no match → [x]" for the receiver side, but says nothing about what happens when `x` itself doesn't exist. Suggest either:
1. Return `Invalid` (Nothing) when `x` is `Invalid`, to be consistent with how Nothing propagates elsewhere.
2. Or document this "append nothing" behaviour explicitly and add a test case for it.

---

json_path/src/json_path.rs
```diff
@@ -790,31 +881,38 @@ fn eval_function<'i, 'j, S: SelectValue>(
...
+        TermEvaluationResult::Null => vec![Value::Null],
+        TermEvaluationResult::Literal(v) => vec![v],
         // A real document node (e.g. `$.a.first()`, `value($.a)`): serialize it to a Value.
         TermEvaluationResult::Value(vref) => {
-            Some(serde_json::to_value(&vref).unwrap_or(Value::Null))
+            vec![serde_json::to_value(&vref).unwrap_or(Value::Null)]
         }
-        // Multiple nodes (e.g. `($..x)`): render as a JSON array.
-        TermEvaluationResult::NodeList(list) => Some(Value::Array(
+        // A multi-node value: render as one JSON array (a parenthesized path is classified as a
```
gabsow: **`NodeList` and `Results` are treated asymmetrically as the `x` argument to `append()`**

`term_to_outputs` wraps a `NodeList` as a single `Value::Array` element but spreads `Results` flat:

```rust
NodeList(list) => vec[Value::Array(list.iter().map(to_value).collect())],  // ONE element
Results(vs)    => vs,  // flat spread
```

This means the *same logical payload* produces structurally different output depending on how it arrived:

- `$.arr.append($.obj.keys())` — arg is `Results(["k1","k2"])` → flattened into output → `[...arr, "k1", "k2"]`
- `$.arr.append($..nodes)` (multi-node path) — arg is `NodeList([n1,n2])` → wrapped → `[...arr, [n1, n2]]`

Users who pass a multi-value path expression as the `x` argument will get a nested array rather than individual elements, which is inconsistent with the `keys()` case.

If the intent is that `append(multi_value_expr)` always appends the collection as a single unit, then `Results` should also be wrapped: `Results(vs) => vec[Value::Array(vs)]`. If the intent is to flatten both, then `NodeList` should be flattened too. Either way, the current asymmetry is likely unintentional and should be made consistent (and tested).

---

gabsow (changes requested):
## Review of MOD-16274/MOD-16275 — `keys()`/`~` and `append()` projections

Five confirmed findings, ranked by severity. All were verified against the actual diff.

### Summary

| # | File | Severity | Finding |
|---|------|----------|---------|
| 1 | `grammar.pest:1` | High | `~` removal from `literal` is a silent breaking change for field names containing `~` |
| 2 | `grammar.pest:96` | Medium | Grammar allows `$.obj.keys()~` to parse; silently returns `[]` instead of a parse error |
| 3 | `json_path.rs:478` | Medium | `function_length` returns `Invalid` for `Results` — asymmetric with `count` |
| 4 | `json_path.rs:788` | Low-Medium | `function_append` silently drops `x` when the appended-value path is absent |
| 5 | `json_path.rs:905` | Low | `NodeList` vs `Results` treated asymmetrically as `append()` argument |

### Notes

- The overall architecture (new `Results` term variant, flat `term_to_outputs`, `Values::Computed(Vec<…>)`) is clean and the wiring through JSON.GET / JSON.MGET / JSON.RESP / RESP3 EXPAND is consistent.
- Most of the logic is correct; the findings above are edge cases or inconsistencies rather than fundamental design problems.
- Finding #1 is the most user-facing: clients relying on dot notation for `~`-containing keys will break silently after upgrade.

---

gabsow:
## Review follow-up — status after latest CR commits

Thanks for the quick turnaround on the CR commits. Here's where things stand:

### ✅ Addressed

- **`$.obj.keys()~` parse ambiguity** — grammar now rejects `$.obj.keys()~`, `$.obj~~`, etc. as parse errors. Good.
- **`function_length` asymmetry with `Results`** — lines changed in the new commits, thread is outdated.
- **`function_append` silent drop on `Nothing`** — now documented explicitly: `$.arr.append($.missing)` → `[]`.
- **Breaking change migration** — the PR description now has a clear "Compatibility (breaking)" section explaining that `~` is no longer a valid field-name character and showing the `$.a~b` → `$["a~b"]` migration path.

### Still open

- **`NodeList` vs `Results` asymmetry in `term_to_outputs`** — the PR description documents this as intentional: a multi-node path argument to `append` is wrapped as a single element rather than spread. That's a reasonable design choice, but worth a short inline note in the code explaining *why* NodeList is wrapped (to preserve "one argument → one appended element" semantics). As written, a reader will likely read `vec[Value::Array(...)]` vs `vs` and assume it's a bug.

- **`in`/`nin` filter with `Results` RHS** (raised by Bugbot) — `Results` is not coerced to an array when used as the right-hand side of `in`/`nin`. This means `?@ in $.obj.keys()` silently fails instead of doing a membership test. Probably deserves a fix or at minimum a documented limitation before merging.

Overall the PR is in much better shape. The two remaining items are relatively small — a comment and a filter-operator edge case.

---

gabsow:
I went ahead and addressed the two remaining open items. The fixes are on branch `claude/redisjson-pr-1607-review-tbvdbd` (one commit: `ef6f726`), ready to cherry-pick or pull into `projection-cont`.

### What was fixed

**1. `in`/`nin` with a `Results` RHS** (`json_path/src/json_path.rs`)

`member_of` was falling through to `false` whenever the right-hand side was a `Results` (synthesized list from `keys()`/`~`/`append()`). Added a `Results(vs)` arm that iterates and checks membership, mirroring the existing `Literal(Value::Array)` arm:

```rust
// Before: Results fell through to the false catch-all
TermEvaluationResult::Results(_) | TermEvaluationResult::Invalid => false,

// After
TermEvaluationResult::Results(vs) => vs.iter().any(|v| self.equals_value(v)),
TermEvaluationResult::Invalid => false,
```

So `?("x" in @.obj.keys())` now correctly filters nodes where `"x"` is a key. Added `test_in_nin_with_results_rhs` covering both `in` and `nin`.

**2. NodeList/Results asymmetry comment** (`json_path/src/json_path.rs`)

Replaced the vague comment on the `NodeList` arm in `term_to_outputs` with one that explains the design intent: NodeList wraps as a single array to preserve the "one path argument → one output element" invariant (consistent with the documented `append` behavior), while Results emits flat because each synthesized value is an independent reply element.

All 174 unit tests pass.

---

gabsow:
## Review status update

Good progress — the `in`/`nin` Results fix is resolved. One new issue surfaced on the latest commit though.

### Still open

**1. Breaking change thread** (`grammar.pest` line 1) — the PR description now has the migration note, but the review thread itself hasn't been marked resolved. Please close it when you get a chance.

**2. NodeList/Results asymmetry** (`term_to_outputs`) — flagged as intentional per the PR description. Same ask: please mark the thread resolved so it's clear it's a conscious design decision and not overlooked.

**3. NEW — Bugbot: "Results lack filter comparison any-of"** (raised on `1a581220`)

Direct comparisons (`==`, `!=`, `>`, `<`, etc.) treat a `Results` operand as non-comparable and always yield false, while `NodeList` uses any-of semantics. So:

```
?(@.obj.keys() == "id")   // always false even when "id" is a key
?(@.obj.keys() > "a")     // always false
```

This is the same class of gap as the `in`/`nin` issue. The fix would be similar — in the comparison dispatch, add a `Results` arm that iterates and checks if **any** element satisfies the comparison (mirroring `NodeList` any-of semantics).

Once those three are addressed the PR should be in good shape.

---

gabsow (approved):
All review findings addressed. The `function_length` thread is outdated (code changed) so the underlying issue is fixed — just needs the conversation marked resolved. Good to merge.

---

PR #1609, opened 2026-06-23
MOD-13014: hide user data from trace logs by default (fail closed)

gabsow:
Follow-up to #1605 (already merged).

## Problem

The cached `hide-user-data-from-log` gate defaulted to `false` (show), matching Redis core's config default. That means in any context without a successfully-read config — a `CONFIG GET` error, an older Redis that lacks the config, the standalone `jsonpath` binary, or the unit tests — the JSONPath engine's trace logs would **show** user data.

## Change

Fail closed so user data is hidden unless the server is known to permit it:

- The cached flag now defaults to **`true` (hidden)**.
- `read_hide_user_data_from_log` returns `true` on any read failure, and shows data **only** when the server explicitly sets `hide-user-data-from-log no`.

A running server still honors an explicit `hide-user-data-from-log no` (shows data); everything else hides.

## Testing

- `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings` — clean
- All `json_path` tests pass (the gate-does-not-change-results regression test updated for the new default)

---

> [!NOTE]
> **Medium Risk**
> Changes default logging/privacy behavior for trace output (not query results); admins who expect verbose traces without explicit CONFIG may see less detail until they set hide-user-data-from-log no.
> 
> **Overview**
> **JSONPath trace logging now hides user data by default** instead of showing it when Redis config cannot be read or is ambiguous.
> 
> The cached `hide-user-data-from-log` flag in `json_path` defaults to **`true`**, and `read_hide_user_data_from_log` in the RedisJSON module returns **`true` on any `CONFIG GET` failure** and only allows verbose `trace_user_data!` output when the server explicitly sets **`hide-user-data-from-log no`** (previously it defaulted open and treated **`yes`** as the signal to hide). Standalone `jsonpath`, tests, and older Redis builds without the config therefore **no longer emit document values in trace logs** unless a running server opts out. A unit test was updated to restore the hidden-by-default flag after exercising both modes; query results are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2819dbb24dcefb6c4a26a73e82f06d7c6cb89a8d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/json_path/src/json_path.rs
+++ b/json_path/src/json_path.rs
@@ -24,13 +24,15 @@ use std::sync::atomic::{AtomicBool, Ordering as AtomicOrdering};
 
 /// Cached mirror of Redis' `hide-user-data-from-log` server config.
 ///
-/// Defaults to `false`, which is Redis core's own default for the config. The
-/// RedisJSON module keeps it in sync — once at load time and again on every
-/// `CONFIG SET` (see `sync_hide_user_data_from_log` in the `redis_json`
-/// crate). When this crate is used outside the module (the standalone
-/// `jsonpath` binary or the unit tests) there is no server to read from, so
-/// the default preserves the previous, fully verbose tracing behaviour.
-static HIDE_USER_DATA_FROM_LOG: AtomicBool = AtomicBool::new(false);
+/// Defaults to `true` (hidden): user data is kept out of the logs unless the
+/// server is known to permit it. The RedisJSON module keeps this in sync —
+/// once at load time and again on every `CONFIG SET` (see
+/// `sync_hide_user_data_from_log` in the `redis_json` crate), where it shows
+/// user data only when the server explicitly sets `hide-user-data-from-log no`.
+/// Failing closed means contexts without a readable config — a config read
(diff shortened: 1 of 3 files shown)


gabsow:
Closing — not needed. The merged work (#1605) already satisfies the requirement:

- RedisJSON mirrors core's `hide-user-data-from-log`, so **Redis Enterprise hides user data** by setting the config to `yes` and blocking tenant `CONFIG SET` (the same mechanism that secures core's own log redaction). Verified that core does *not* scrub module-emitted log strings, so RedisJSON honoring the flag itself is the load-bearing piece — and that's merged.
- The merged code defaults the gate to `false`, which intentionally lets unit tests and the standalone `jsonpath` binary show user data when debugging.

This follow-up would have flipped that default to hidden (also hiding it in tests), which isn't desired. The only other thing it added was failing closed on an unreadable config at load — which can't happen in a correctly-configured Enterprise deploy. So: closing.

PR #1610, opened 2026-06-23
MOD-13014: hide user data from trace logs by default (fail closed) [8.10]

gabsow:
Cherry-pick of #1609 onto `8.10`. Follow-up to #1608 (already merged).

Flips the JSONPath trace-log redaction gate to fail closed:

- the cached `hide-user-data-from-log` flag now defaults to **`true` (hidden)**
- `read_hide_user_data_from_log` returns `true` on any read failure, and shows data **only** when the server explicitly sets `hide-user-data-from-log no`

So contexts without a readable config (read errors, older Redis lacking the config, the standalone `jsonpath` binary, unit tests) keep user data hidden rather than leaking it. A running server still honors an explicit `hide-user-data-from-log no`.

## Testing
- `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings` — clean
- All `json_path` tests pass

---

> [!NOTE]
> **Medium Risk**
> Changes logging/redaction defaults and CONFIG interpretation; query results are unchanged but operators who relied on verbose traces without an explicit `hide-user-data-from-log no` will see less detail.
> 
> **Overview**
> **JSONPath trace logs now redact user data by default** instead of only when Redis says to hide it. The cached `hide-user-data-from-log` flag starts as **`true`**, and `trace_user_data!` still skips those traces while the flag is set.
> 
> In the RedisJSON module, **`read_hide_user_data_from_log` fails closed**: `CONFIG GET` errors or missing/unparseable values keep redaction on. User data appears in traces **only** when the server sets `hide-user-data-from-log` to **`no`** (the previous logic treated **`yes`** as the signal to hide and defaulted to showing data on read failure).
> 
> Standalone `jsonpath`, tests, and older Redis without the config therefore **no longer emit verbose user-data traces** unless something explicitly opts out. A unit test now restores the hidden default after toggling the flag.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 321966a7c24b53369651e64f43d3e4b60d57fd55. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/json_path/src/json_path.rs
+++ b/json_path/src/json_path.rs
@@ -24,13 +24,15 @@ use std::sync::atomic::{AtomicBool, Ordering as AtomicOrdering};
 
 /// Cached mirror of Redis' `hide-user-data-from-log` server config.
 ///
-/// Defaults to `false`, which is Redis core's own default for the config. The
-/// RedisJSON module keeps it in sync — once at load time and again on every
-/// `CONFIG SET` (see `sync_hide_user_data_from_log` in the `redis_json`
-/// crate). When this crate is used outside the module (the standalone
-/// `jsonpath` binary or the unit tests) there is no server to read from, so
-/// the default preserves the previous, fully verbose tracing behaviour.
-static HIDE_USER_DATA_FROM_LOG: AtomicBool = AtomicBool::new(false);
+/// Defaults to `true` (hidden): user data is kept out of the logs unless the
+/// server is known to permit it. The RedisJSON module keeps this in sync —
+/// once at load time and again on every `CONFIG SET` (see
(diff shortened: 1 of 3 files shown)


gabsow:
Closing along with #1609 — not needed. The merged 8.10 work (#1608) already does the job: RedisJSON mirrors `hide-user-data-from-log`, so Enterprise hides user data via config (`yes` + tenant CONFIG SET blocked), while the gate's `false` default intentionally lets tests and the standalone `jsonpath` binary show user data when debugging. The fail-closed flip would have hidden it in tests too, which isn't wanted.

PR #1613, opened 2026-06-30
remove `#[cfg(not(feature = "as-library"))]` from `sync_hide_user_dat…

AvivDavid23:
…a_from_log`

---

> [!NOTE]
> **Low Risk**
> Small visibility/cfg refactor around log redaction sync; no change to JSON command logic or security-sensitive paths beyond making the same sync API available in library builds.
> 
> **Overview**
> **`hide-user-data-from-log` plumbing is no longer limited to non–`as-library` builds.** `read_hide_user_data_from_log`, `sync_hide_user_data_from_log` (now **`pub`**), and `hide_user_data_config_changed` lose their `#[cfg(not(feature = "as-library"))]` attributes so they compile for every feature set.
> 
> Imports are split so **`Context` and `RedisValue` are always available** for those helpers, while module-only types like **`RedisResult`** stay behind the `as-library` gate. Runtime behavior for the loaded module is unchanged: init still calls `sync_hide_user_data_from_log`, and the config-changed handler still refreshes `json_path` when Redis updates the setting.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bbc27c25399a13079af9c634d5bb72deef3913ae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis_json/src/lib.rs
+++ b/redis_json/src/lib.rs
@@ -18,7 +18,7 @@ use redis_module::InfoContext;
 
 #[cfg(not(feature = "as-library"))]
 use redis_module::Status;
-#[cfg(not(feature = "as-library"))]
+
 use redis_module::{Context, RedisResult, RedisValue};
 
 #[cfg(not(feature = "as-library"))]
@@ -457,7 +457,6 @@ pub fn setup_panic_handler() {
 /// Read Redis' `hide-user-data-from-log` server config via `CONFIG GET`.
 /// Returns `false` (the server's own default) when the config cannot be read
 /// or parsed.
-#[cfg(not(feature = "as-library"))]
 fn read_hide_user_data_from_log(ctx: &Context) -> bool {
     let Ok(reply) = ctx.call("CONFIG", &["GET", "hide-user-data-from-log"]) else {
         return false;
@@ -477,14 +476,12 @@ fn read_hide_user_data_from_log(ctx: &Context) -> bool {
 /// Refresh the json path engine's cached copy of Redis' `hide-user-data-from-log`
 /// server config, so its trace logs redact user data exactly when Redis core
 /// does. Called once at module load and again on every relevant `CONFIG SET`.
-#[cfg(not(feature = "as-library"))]
 pub fn sync_hide_user_data_from_log(ctx: &Context) {
     json_path::set_hide_user_data_from_log(read_hide_user_data_from_log(ctx));
 }
 
 /// Keep the cached `hide-user-data-from-log` value in sync when it is changed at
 /// runtime via `CONFIG SET`.
-#[cfg(not(feature = "as-library"))]
 #[::redis_module_macros::config_changed_event_handler]
 fn hide_user_data_config_changed(ctx: &Context, changed_configs: &[&str]) {
     if changed_configs.contains(&"hide-user-data-from-log") {
--- a/redis_json/src/lib.rs
+++ b/redis_json/src/lib.rs
@@ -19,7 +19,9 @@ use redis_module::InfoContext;
 #[cfg(not(feature = "as-library"))]
 use redis_module::Status;
 
-use redis_module::{Context, RedisResult, RedisValue};
+#[cfg(not(feature = "as-library"))]
+use redis_module::RedisResult;
+use redis_module::{Context, RedisValue};
 
 #[cfg(not(feature = "as-library"))]
 use redis_module::key::KeyFlags;


redis_json/src/lib.rs
```diff
@@ -18,7 +18,7 @@ use redis_module::InfoContext;
 
 #[cfg(not(feature = "as-library"))]
 use redis_module::Status;
-#[cfg(not(feature = "as-library"))]
+
 use redis_module::{Context, RedisResult, RedisValue};
```
gabsow: Ungating the whole import line introduces a **new `unused import: RedisResult` warning** in `as-library` builds. The three functions you ungated use only `Context` and `RedisValue`; `RedisResult` is referenced solely inside the `redis_json_module_create!` macro body, whose only invocation stays `#[cfg(not(feature = "as-library"))]`-gated. So under `--features as-library` `RedisResult` is now imported but never used.

Verified with `cargo check -p redis_json --features as-library` on this branch: it emits `warning: unused import: `RedisResult`` at this line (this warning is not present on master). It's non-fatal today since as-library builds already carry other warnings, but under `-D warnings` it would re-break the build this PR is fixing. Keep `RedisResult` gated:

```suggestion
#[cfg(not(feature = "as-library"))]
use redis_module::RedisResult;
use redis_module::{Context, RedisValue};
```

redis_json/src/lib.rs
```diff
@@ -477,14 +478,12 @@ fn read_hide_user_data_from_log(ctx: &Context) -> bool {
...
 /// does. Called once at module load and again on every relevant `CONFIG SET`.
-#[cfg(not(feature = "as-library"))]
 pub fn sync_hide_user_data_from_log(ctx: &Context) {
     json_path::set_hide_user_data_from_log(read_hide_user_data_from_log(ctx));
 }
 
 /// Keep the cached `hide-user-data-from-log` value in sync when it is changed at
 /// runtime via `CONFIG SET`.
-#[cfg(not(feature = "as-library"))]
 #[::redis_module_macros::config_changed_event_handler]
```
gabsow, line 487: **Asymmetric wiring in `as-library` builds: handler auto-registers but the load-time sync does not run.**

Removing `#[cfg(not(feature = "as-library"))]` here makes `hide_user_data_config_changed` register itself into the global `linkme` slice `CONFIG_CHANGED_SERVER_EVENTS_LIST` in *every* build, including `as-library`. When a host module embeds redis_json as a library, that handler is collected by the host's `redis_module!` and will fire on `CONFIG SET hide-user-data-from-log`.

However, the *initial* sync — `sync_hide_user_data_from_log(ctx)` — lives inside the `redis_json_module_create!` macro's `initialize` fn (lib.rs:382), which is still gated `#[cfg(not(feature = "as-library"))]` (lib.rs:525) and is **not** invoked in library builds.

Net effect for an `as-library` consumer: the cached value starts at its default `false` and is only corrected on the *first runtime* `CONFIG SET`. A server started with `hide-user-data-from-log yes` that never toggles it at runtime will have JSONPath trace logs **not redact user data** — a fail-open privacy gap. The embedder must remember to call the now-`pub` `sync_hide_user_data_from_log` at its own init; consider documenting that requirement on the function, or gating the handler registration on the same condition as the init sync so the two stay symmetric.

AvivDavid23: `redis_json_module_create` is the start point of every module, other external consumers which will crate a module based on this repo will still go though it...

gabsow: You're right, and I'll retract my comment above. I traced it through the actual `as-library` consumer (RedisJsonHdt), and the asymmetry I described doesn't exist:

The consumer invokes `redis_json_module_create!` itself to build its module. That `#[macro_export]` macro's generated `initialize` fn — registered as a module init — calls `$crate::sync_hide_user_data_from_log(ctx)` at load, and its inner `redis_module!` consumes the `linkme` slice that registers `hide_user_data_config_changed`. So the load-time sync **and** the runtime handler both run for the embedding module — symmetrically. The `#[cfg(not(feature = "as-library"))]` on line 525 only suppresses redis_json's *own* duplicate module entrypoint, not the macro body where the sync call lives.

Net: ungating these three fns is exactly what's needed — without it the consumer's `redis_json_module_create!` expansion would reference `redis_json::sync_hide_user_data_from_log`, which wouldn't exist under `--features as-library`, and the build would fail to compile. No fail-open, no missing init. LGTM.

gabsow (approved):
Verified the `as-library` build compiles cleanly (no new warnings) and traced the change through the actual consumer (RedisJsonHdt): it invokes `redis_json_module_create!` itself, so both the load-time `sync_hide_user_data_from_log` and the runtime config-changed handler run for the embedding module — symmetric, no fail-open. Ungating these fns is in fact required for the consumer's macro expansion to compile under `--features as-library`. LGTM.

PR #1614, opened 2026-06-30
MOD-16366 json add flow tests to check llapi c api

AvivDavid23:
Adds a pytest flow tests which are testing the LLAPI the json module registers at `redis_json/src/c_api.rs`.

- Creates a dummy C redis module which registers functions that uses the API and returns results based on it
- Register the module when creating a new env in these tests
- Add a pytest tests on top on it, which calles it and validates the results

---

> [!NOTE]
> **Low Risk**
> Test-only infrastructure and fixtures; no changes to the JSON module’s production LLAPI registration or runtime behavior.
> 
> **Overview**
> Adds **end-to-end flow coverage** for the RedisJSON LLAPI (`RedisModule_GetSharedAPI` / `RedisJSON_V<n>`), which client commands cannot reach directly.
> 
> A small C **consumer module** (`llapi_test_module`) binds the latest shared API and exposes thin **`LLAPI.*`** commands that mirror `RedisJSONAPI` (open/get, iterators, scalars, arrays, key-values, path parse, `isJSON`, etc.). **`test_llapi.py`** loads RedisJSON plus that module, seeds a fixture document, and asserts happy paths and error behavior (missing keys, type mismatches, bad paths).
> 
> **`tests.sh`** now auto-builds `llapi_test.so` (initializing **`deps/RedisModulesSDK`** when needed), sets **`LLAPI_TEST_MODULE`**, and fails the run if the build breaks; direct pytest runs skip when the `.so` is absent. **`.gitignore`** ignores built artifacts; **`llapi_doc.json`** is the shared test payload.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 322cde0253c72ff083bd8c436f32c3f4cb4315ea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/pytest/llapi_test_module/module.c
+++ b/tests/pytest/llapi_test_module/module.c
@@ -1,0 +1,509 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of (a) the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+ */
+
+/*
+ * Test consumer module for the RedisJSON shared C API (LLAPI).
+ *
+ * This module fetches the RedisJSON shared API via RedisModule_GetSharedAPI and
+ * exposes each RedisJSONAPI function as a thin `LLAPI.*` command, so the Python
+ * flow tests can exercise the C API end-to-end against a live Redis.
+ *
+ * It binds the API *by name* ("RedisJSON_V<n>") and never depends on RedisJSON
+ * internals, so the same tests can run against any module that exports the same
+ * shared API (e.g. JsonHDT in the OSS test suite, MOD-14113).
+ *
+ * It must be loaded *after* the JSON module under test, e.g.:
+ *   redis-server --loadmodule rejson.so --loadmodule llapi_test.so
+ */
+
+#define REDISMODULE_MAIN
+#include "redismodule.h"
+#include "rejson_api.h"
+
+#include <string.h>
+#include <stdlib.h>
+
+static const RedisJSONAPI *japi = NULL;
+static int japi_ver = 0;
+
+/* Map a JSONType to a stable lowercase name used in replies. */
+static const char *json_type_name(JSONType t) {
+    switch (t) {
+        case JSONType_String: return "string";
+        case JSONType_Int:    return "int";
+        case JSONType_Double: return "double";
+        case JSONType_Bool:   return "bool";
+        case JSONType_Object: return "object";
+        case JSONType_Array:  return "array";
+        case JSONType_Null:   return "null";
+        default:              return "unknown";
+    }
+}
+
+/* Map a JSONArrayType to a stable name used in replies. */
+static const char *json_array_type_name(JSONArrayType t) {
+    switch (t) {
+        case JSONArrayType_Heterogeneous: return "heterogeneous";
+        case JSONArrayType_I8:   return "i8";
+        case JSONArrayType_U8:   return "u8";
+        case JSONArrayType_I16:  return "i16";
+        case JSONArrayType_U16:  return "u16";
+        case JSONArrayType_F16:  return "f16";
+        case JSONArrayType_BF16: return "bf16";
+        case JSONArrayType_I32:  return "i32";
+        case JSONArrayType_U32:  return "u32";
+        case JSONArrayType_F32:  return "f32";
+        case JSONArrayType_I64:  return "i64";
+        case JSONArrayType_U64:  return "u64";
+        case JSONArrayType_F64:  return "f64";
+        default:                 return "unknown";
+    }
+}
+
+/* Open the key named by argv[1] and run get(path=argv[2]).
+ * On failure, replies with an error and returns NULL (caller should return). */
+static JSONResultsIterator open_and_get(RedisModuleCtx *ctx, RedisModuleString **argv) {
+    RedisJSON json = japi->openKey(ctx, argv[1]);
+    if (!json) {
+        RedisModule_ReplyWithError(ctx, "ERR key does not exist or is not JSON");
+        return NULL;
+    }
+    const char *path = RedisModule_StringPtrLen(argv[2], NULL);
+    JSONResultsIterator it = japi->get(json, path);
+    if (!it) {
+        RedisModule_ReplyWithError(ctx, "ERR path could not be evaluated");
+        return NULL;
+    }
+    return it;
+}
+
+/* LLAPI.VERSION -> the bound shared-API version (integer). */
+static int VersionCmd(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
+    REDISMODULE_NOT_USED(argv);
+    if (argc != 1) return RedisModule_WrongArity(ctx);
+    RedisModule_ReplyWithLongLong(ctx, japi_ver);
+    return REDISMODULE_OK;
+}
+
+/* LLAPI.OPEN_GET key path -> array of the JSON string of every matched node. */
+static int OpenGetCmd(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
+    if (argc != 3) return RedisModule_WrongArity(ctx);
+    JSONResultsIterator it = open_and_get(ctx, argv);
+    if (!it) return REDISMODULE_OK;
+
+    RedisModule_ReplyWithArray(ctx, REDISMODULE_POSTPONED_LEN);
+    long n = 0;
+    RedisJSON node;
+    while ((node = japi->next(it)) != NULL) {
+        RedisModuleString *s = NULL;
+        if (japi->getJSON(node, ctx, &s) == REDISMODULE_OK) {
+            RedisModule_ReplyWithString(ctx, s);
+            RedisModule_FreeString(ctx, s);
+        } else {
+            RedisModule_ReplyWithError(ctx, "ERR getJSON failed");
+        }
+        n++;
+    }
+    RedisModule_ReplySetArrayLength(ctx, n);
+    japi->freeIter(it);
+    return REDISMODULE_OK;
+}
+
+/* LLAPI.ITER_JSON key path -> getJSONFromIter (whole result set as one JSON string). */
+static int IterJsonCmd(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
+    if (argc != 3) return RedisModule_WrongArity(ctx);
+    JSONResultsIterator it = open_and_get(ctx, argv);
+    if (!it) return REDISMODULE_OK;
+
+    RedisModuleString *s = NULL;
+    if (japi->getJSONFromIter(it, ctx, &s) == REDISMODULE_OK) {
+        RedisModule_ReplyWithString(ctx, s);
+        RedisModule_FreeString(ctx, s);
+    } else {
+        RedisModule_ReplyWithError(ctx, "ERR getJSONFromIter failed");
+    }
+    japi->freeIter(it);
+    return REDISMODULE_OK;
+}
+
+/* LLAPI.ITER_LEN key path -> number of matched nodes (iterator len). */
+static int IterLenCmd(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
+    if (argc != 3) return RedisModule_WrongArity(ctx);
+    JSONResultsIterator it = open_and_get(ctx, argv);
+    if (!it) return REDISMODULE_OK;
+    RedisModule_ReplyWithLongLong(ctx, (long long)japi->len(it));
+    japi->freeIter(it);
+    return REDISMODULE_OK;
+}
+
+/* LLAPI.RESET key path -> [count_first_pass, count_after_reset]. Proves resetIter. */
+static int ResetCmd(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
+    if (argc != 3) return RedisModule_WrongArity(ctx);
+    JSONResultsIterator it = open_and_get(ctx, argv);
+    if (!it) return REDISMODULE_OK;
+
+    long long first = 0, second = 0;
+    while (japi->next(it) != NULL) first++;
+    japi->resetIter(it);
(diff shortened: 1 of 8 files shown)


tests/pytest/tests.sh
```diff
@@ -504,6 +504,25 @@ if [[ $RLEC != 1 ]]; then
...
+# test_llapi.py loads a small C consumer module that fetches the JSON shared API
+# (RedisJSON_V<n>) and exposes it as LLAPI.* commands.
+LLAPI_TEST_DIR="$HERE/llapi_test_module"
+if [[ -f "$LLAPI_TEST_DIR/module.c" ]]; then
+	case "$(uname -s)" in
+		Darwin) LLAPI_TEST_LDFLAGS="-bundle -undefined dynamic_lookup" ;;
+		*)      LLAPI_TEST_LDFLAGS="-shared" ;;
+	esac
+	if "${LLAPI_TEST_CC:-cc}" -fPIC $LLAPI_TEST_LDFLAGS -O2 -Wall \
+		-I"$ROOT/deps/RedisModulesSDK" -I"$ROOT/redis_json/src/include" \
```
gabsow, line 521: **Submodule not initialized here, and `redismodule.h` is not vendored in the repo.** This `-I"$ROOT/deps/RedisModulesSDK"` is the only source of `redismodule.h` (the repo only ships `redis_json/src/include/rejson_api.h`), and `deps/RedisModulesSDK` is a brand-new submodule. `tests.sh` never runs `git submodule update --init`. CI is fine because the workflows check out with `submodules: recursive`, but on any non-recursive checkout (fresh clone, local dev, an alternate CI lane) the include dir is empty, the build fails, and you fall through to the soft warning below — surfacing as an opaque `RuntimeError` in every test rather than "submodule missing". Consider `git submodule update --init deps/RedisModulesSDK` here (or a clear precondition check).

tests/pytest/tests.sh
```diff
@@ -504,6 +504,25 @@ if [[ $RLEC != 1 ]]; then
...
+		-o "$LLAPI_TEST_DIR/llapi_test.so" "$LLAPI_TEST_DIR/module.c"; then
+		export LLAPI_TEST_MODULE="$LLAPI_TEST_DIR/llapi_test.so"
+	else
+		echo "warning: failed to build LLAPI test module; test_llapi.py will fail"
```
gabsow: **Build failure is swallowed.** On a compile error you `echo` a warning and continue without exporting `LLAPI_TEST_MODULE`. Two failure modes: (1) if `test_llapi.py` is collected, every test fails with a confusing `RuntimeError` far from the real cause (the compile error scrolled past); (2) if a given run filters `test_llapi.py` out, a broken `module.c` goes completely undetected — there is no build-success invariant. Recommend failing the build step loudly on compile error so the cause is reported where it happens.

tests/pytest/llapi_test_module/module.c
```diff
@@ -0,0 +1,509 @@
...
+        return REDISMODULE_ERR;
+
+    if (fetch_japi(ctx) != REDISMODULE_OK) {
+        RedisModule_Log(ctx, "warning",
+            "llapi_test: RedisJSON shared API not found; load a JSON module first");
+        return REDISMODULE_ERR;
+    }
+    RedisModule_Log(ctx, "notice", "llapi_test: bound RedisJSON shared API V%d", japi_ver);
+
+    REGISTER("LLAPI.VERSION", VersionCmd);
```
gabsow, line 487: **V4/V6/V7 commands are registered (and called) unconditionally, regardless of the bound API version.** `fetch_japi()` binds the highest available `RedisJSON_V<n>` down to V1, and `testLLAPIVersion` explicitly accepts `ver >= 1`, matching the file header's stated goal of running against *any* module exporting this shared API (e.g. JsonHDT, MOD-14113). But `GETKEYVALUES` (V4), `GETAT`/`allocJson`/`freeJson` (V6) and `GETARRAY` (V7) are always registered and dereference those struct fields. Against RedisJSON this is safe (it exports the *same* full struct under every version name — see `c_api.rs:858-864`), but against a genuinely lower-version provider those fields lie past the provider's struct → garbage/NULL function pointer → crash. If cross-module reuse is a real goal, gate the `REGISTER(...)` calls (or the command bodies) on `japi_ver`.

tests/pytest/test_llapi.py
```diff
@@ -0,0 +1,207 @@
...
+    m = Defaults.module
+    if not m:
+        return []
+    return list(m) if isinstance(m, (list, tuple)) else [m]
+
+
+def _new_env():
+    """A fresh Env loading the JSON module under test plus the LLAPI consumer."""
+    if not LLAPI_MODULE or not os.path.exists(LLAPI_MODULE):
+        raise RuntimeError(
```
gabsow: **Running `pytest` outside `tests.sh` errors the whole file instead of skipping.** `_new_env` raises `RuntimeError` when `LLAPI_TEST_MODULE` is unset/missing, so a direct `pytest`/IDE run (the common debug loop) fails all ~17 tests rather than skipping them. If skip-on-missing is intended, use `unittest.SkipTest` / `pytest.skip`; if fail-loud is intended (so a missing module can't hide), that's defensible but worth a comment stating it.

tests/pytest/llapi_test_module/module.c
```diff
@@ -0,0 +1,509 @@
...
+    if (!node) {
+        RedisModule_ReplyWithError(ctx, "ERR no value at path");
+        japi->freeIter(it);
+        return REDISMODULE_OK;
+    }
+
+    size_t len = 0;
+    JSONArrayType atype = JSONArrayType_Heterogeneous;
+    const void *arr = japi->getArray(node, &len, &atype);
+    if (!arr && len == 0 && japi->getType(node) != JSONType_Array) {
```
gabsow: **Fragile not-an-array detection.** `if (!arr && len == 0 && japi->getType(node) != JSONType_Array)` reconstructs "not an array" from three coupled signals because `getArray` returns `(NULL, 0)` for non-arrays *and* potentially for empty arrays. It works today only because an empty array reports `getType == Array`. The `TYPE`/`KEYVALUES`/`SCALAR` handlers already gate cleanly on `getType(node)` up front — doing the same here (`if (getType(node) != JSONType_Array) error;` before calling `getArray`) is unambiguous and not coupled to undocumented NULL-vs-empty pointer semantics.

tests/pytest/llapi_test_module/module.c
```diff
@@ -0,0 +1,509 @@
...
+}
+
+/* LLAPI.ITER_JSON key path -> getJSONFromIter (whole result set as one JSON string). */
+static int IterJsonCmd(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
+    if (argc != 3) return RedisModule_WrongArity(ctx);
+    JSONResultsIterator it = open_and_get(ctx, argv);
+    if (!it) return REDISMODULE_OK;
+
+    RedisModuleString *s = NULL;
+    if (japi->getJSONFromIter(it, ctx, &s) == REDISMODULE_OK) {
```
gabsow, line 129: **`LLAPI.ITER_JSON` returns an error for a valid path that matches zero nodes.** `getJSONFromIter` returns `REDISMODULE_ERR` when the result set is empty (`c_api.rs:205`, `pos >= len`), so `LLAPI.ITER_JSON doc $.does_not_exist` replies `ERR getJSONFromIter failed`, while `LLAPI.ITER_LEN` returns `0` and `LLAPI.OPEN_GET` returns `[]` for the same input. Not covered by a test today, but the inconsistent error surface is surprising — consider replying an empty result for the zero-match case to match the other iterators.

tests/pytest/llapi_test_module/module.c
```diff
@@ -0,0 +1,509 @@
...
+    RedisModule_ReplyWithLongLong(ctx, second);
+    japi->freeIter(it);
+    return REDISMODULE_OK;
+}
+
+/* LLAPI.OPENFROMSTR keyname path -> number of matched nodes (uses openKeyFromStr). */
+static int OpenFromStrCmd(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
+    if (argc != 3) return RedisModule_WrongArity(ctx);
+    const char *keyname = RedisModule_StringPtrLen(argv[1], NULL);
+    RedisJSON json = japi->openKeyFromStr(ctx, keyname);
```
gabsow: _(cleanup)_ `OpenFromStrCmd` and `OpenFlagsCmd` re-implement the open-key + `get(path)` + error-reply body of `open_and_get` (lines 122-135), differing only in the open call (`openKeyFromStr` / `openKeyWithFlags`). Three near-identical sequences with duplicated error strings; a helper parameterized by the opened `RedisJSON` would collapse them and keep the error wording in one place.

tests/pytest/llapi_test_module/module.c
```diff
@@ -0,0 +1,509 @@
...
+    japi->freeKeyValuesIter(kv);
+    japi->freeIter(it);
+    return REDISMODULE_OK;
+}
+
+/* LLAPI.ISJSON key -> 1 if the key holds JSON, else 0. */
+static int IsJsonCmd(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
+    if (argc != 2) return RedisModule_WrongArity(ctx);
+    RedisModuleKey *key = RedisModule_OpenKey(ctx, argv[1], REDISMODULE_READ);
+    int res = japi->isJSON(key);
```
gabsow, line 422: _(minor)_ `IsJsonCmd` calls `japi->isJSON(key)` before the `if (key)` null check on the next line. `RedisModule_OpenKey(..., REDISMODULE_READ)` can return NULL, and the trailing `if (key) RedisModule_CloseKey(key)` shows NULL is anticipated — but `isJSON` already received the (possibly NULL) handle. Reorder to null-check before the `isJSON` call (untested today since only existing keys are exercised).

gabsow (commented):
Automated code review (recall-biased, high effort). Test-only PR — no production JSON behavior changes. Findings below are latent/robustness/quality issues in the new test harness and consumer module; none break the current RedisJSON test run, but several affect reliability and the module's stated cross-module reuse goal. Refuted candidates (e.g. `num_array` typed-packing assertion, `allocJson` NULL check, Linux `-shared` symbol resolution) are not reported.

tests/pytest/test_llapi.py
gabsow: Still open after the CR commits (cursor flagged this one too): a direct `pytest tests/pytest/test_llapi.py` / IDE run leaves `LLAPI_TEST_MODULE` unset, so `_new_env()` raises `RuntimeError` and **every** test in the file errors instead of skipping. Since the consumer `.so` is only built and exported by `tests.sh`, consider `raise unittest.SkipTest(...)` (or `pytest.skip(..., allow_module_level=True)`) here so out-of-harness runs skip cleanly rather than fail the file.

tests/pytest/llapi_test_module/module.c
```diff
@@ -0,0 +1,505 @@
...
+}
+
+/* LLAPI.ITER_JSON key path -> getJSONFromIter (whole result set as one JSON string). */
+static int IterJsonCmd(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
+    if (argc != 3) return RedisModule_WrongArity(ctx);
+    JSONResultsIterator it = open_and_get(ctx, argv);
+    if (!it) return REDISMODULE_OK;
+
+    RedisModuleString *s = NULL;
+    if (japi->getJSONFromIter(it, ctx, &s) == REDISMODULE_OK) {
```
gabsow, line 129: Still open after the CR commits (cursor flagged this too): for a valid path that matches **zero** nodes, `getJSONFromIter` returns non-OK, so `LLAPI.ITER_JSON` replies with an error — whereas `LLAPI.ITER_LEN` returns `0` and `LLAPI.OPEN_GET` returns an empty array for the same input. That inconsistency bites anyone using ITER_JSON as the "serialize the whole result set" entry point. Consider treating an empty iterator as an empty (`[]` / null) result rather than an error, to mirror ITER_LEN/OPEN_GET.

gabsow (commented):
Re-reviewed after your CR commits — thanks, most of my comments are addressed (submodule init before the LLAPI build, hard build-failure now `exit 1`, latest-only shared-API bind so no out-of-bounds struct read, not-an-array detection, `OpenFromStr`/`OpenFlags` dedup, `IsJson` null-check order). Two items are still open (both also flagged by cursor) — re-flagged inline. Holding approval until those are addressed or consciously waived.

tests/pytest/llapi_test_module/module.c
AvivDavid23: Looked into this — the inconsistency mirrors the real C API contract, so I'd keep ITER_JSON as-is. In c_api.rs, getJSONFromIter returns Status::Err (and writes no string) when the iterator is empty:
```
▎
▎ if iter.pos >= iter.results.len() {
▎     Status::Err as c_int
▎ } else {
▎     // serialize_object(...) -> JSON array
▎ }
▎
```

gabsow: Confirmed against `redis_json/src/c_api.rs` — `json_api_get_json_from_iter` returns `Status::Err` when `iter.pos >= iter.results.len()`, so ITER_JSON erroring on an empty result set faithfully mirrors the real LLAPI contract. Agreed it should stay as-is; thanks for adding the explicit assertion in `testLLAPIOpenGetAndIter` documenting the intentional difference. 👍

gabsow (approved):
Re-reviewed the latest CR (322cde0) — both of my remaining items are addressed:

- `test_llapi.py` now raises `unittest.SkipTest` instead of `RuntimeError`, so a direct `pytest`/IDE run (with `LLAPI_TEST_MODULE` unset) skips cleanly instead of erroring the whole file.
- `LLAPI.ITER_JSON` on a valid zero-match path: confirmed against `redis_json/src/c_api.rs` that `getJSONFromIter` returns `Status::Err` on an empty iterator, so the test faithfully mirrors the real C API contract — keeping it as-is is correct, and the new assertion documents the intent.

Everything from my earlier review (submodule init before the LLAPI build, hard build-failure, latest-only shared-API bind, dedup, null-check order) is in. LGTM — approving. Thanks for the thorough follow-ups!

PR #1615, opened 2026-06-30
MOD-16300 perf(json_path): materialize filter literals once per query

AvivDavid23:
> [!NOTE]
> **Medium Risk**
> Touches core JSONPath filter evaluation and subquery wiring; behavior should be unchanged but mistakes in caching or Rc sharing could affect correctness on complex filters.
> 
> **Overview**
> Filter evaluation no longer rebuilds constant array/object literals on every matched element. **`PathCalculatorData`** now holds a per-query **`literal_cache`** (keyed by the literal’s parse span) and shares it—along with the regex cache—across **`@`/`$` subqueries** via **`Rc<RefCell<_>>`**, so nested filters like `$.a[?@.b[?@ in [1,2,3]]]` pay one **`build_literal`** / one compile for fixed operands regardless of document size.
> 
> **`TermEvaluationResult::Literal`** is **`Rc<Value>`** instead of an owned **`Value`**, with call sites updated to use **`.as_ref()`** and explicit handling when a literal is a scalar (not only an array/object). Test-only thread-local counters assert literal build and **`Regex::new`** counts stay flat across large arrays and nested subqueries; additional tests cover scalar-literal edge paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c7d670471c12ed609817d627ef7b81f9ce4051d8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/json_path/src/json_path.rs
+++ b/json_path/src/json_path.rs
@@ -1772,6 +1772,17 @@ struct PathCalculatorData<'i, S: SelectValue, UPT: UserPathTracker> {
     literal_cache: HashMap<usize, Rc<Value>>,
 }
 
+impl<'i, S: SelectValue, UPT: UserPathTracker> PathCalculatorData<'i, S, UPT> {
+    fn new(root: ValueRef<'i, S>) -> Self {
+        PathCalculatorData {
+            results: Vec::new(),
+            root,
+            regex_cache: HashMap::new(),
+            literal_cache: HashMap::new(),
+        }
+    }
+}
+
 // The following block of code is used to create a unified iterator for arrays and objects.
 // This can be used in places where we need to iterate over both arrays and objects, create a path tracker from them.
 enum Item<'a, S: SelectValue> {
@@ -2102,12 +2113,7 @@ impl<'i, UPTG: UserPathTrackerGenerator> PathCalculator<'i, UPTG> {
             }
             Rule::from_current => match term.into_inner().next() {
                 Some(term) => {
-                    let mut calc_data = PathCalculatorData {
-                        results: Vec::new(),
-                        root: json.clone(),
-                        regex_cache: HashMap::new(),
-                        literal_cache: HashMap::new(),
-                    };
+                    let mut calc_data = PathCalculatorData::new(json.clone());
                     self.calc_internal(term.into_inner(), json, None, &mut calc_data);
                     Self::results_to_term(calc_data.results)
                 }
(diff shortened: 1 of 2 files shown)


json_path/src/json_path.rs
```diff
@@ -285,14 +286,22 @@ fn unescape_string_value<'a>(pair: Pair<'a, Rule>) -> Cow<'a, str> {
...
+thread_local! {
+    pub(crate) static BUILD_LITERAL_CALLS: std::cell::Cell<usize> = const { std::cell::Cell::new(0) };
+}
+
 /// Recursively build an owned JSON `Value` from an array/object literal parse subtree
 /// (e.g. `[1,{"a":2}]`). Used for structured filter operands; compared against the
 /// document value via the cross-type `is_equal`.
 ///
-/// PERF/TODO: this runs per element during filter evaluation, so a constant literal
-/// (e.g. the array in `$.a[?@ in [1,2,...]]`) is rebuilt for every node. A follow-up
```
TalBarYakar, line 293: remove this comments

json_path/src/json_path.rs
```diff
@@ -2062,23 +2113,15 @@ impl<'i, UPTG: UserPathTrackerGenerator> PathCalculator<'i, UPTG> {
             }
             Rule::from_current => match term.into_inner().next() {
                 Some(term) => {
-                    let mut calc_data = PathCalculatorData {
-                        results: Vec::new(),
-                        root: json.clone(),
-                        regex_cache: HashMap::new(),
-                    };
+                    let mut calc_data = PathCalculatorData::new(json.clone());
```
TalBarYakar: cache is per PathCalculatorData. @/$ subqueries make their own, so a literal in a
nested filter ($.a[?@.b[?@ in [1,2,3]]]) gets rebuilt per outer element, not once.

TalBarYakar:
patch coverage is red — new branches aren't tested. can you add some tests to cover them?

json_path/src/json_path.rs
AvivDavid23: ?
It is removed

json_path/src/json_path.rs
AvivDavid23: good catch, will check this(same issue also with the Regex cache probably

TalBarYakar (approved):

PR #1621, opened 2026-07-14
MOD-16872 make `apply_updates` return the current status of updates

AvivDavid23:
…as succeeded

This fixes issue when multiple paths are being updated. Previously, if one of the updates failed, the command will not be replicated to slaves. User will still get the original None response if one of the updates failed. Changing the same response for MSET

---

> [!NOTE]
> **Medium Risk**
> Changes write-command success, notification, and replication semantics for JSON.SET/MSET on partial failures; behavior is well covered by new tests but clients relying on old OK-on-partial-write behavior may see different replies.
> 
> **Overview**
> **JSON.SET** and **JSON.MSET** now distinguish *whether any matched path/key was written* from *whether every update succeeded*. `apply_updates` returns an `ApplyUpdatesResult` (`NoneUpdated` / `SomeUpdated` / `AllUpdated`) and aggregates multi-path results with `combine` instead of folding booleans with `||`, so a later success is not treated as full success after an earlier failure.
> 
> When only some paths or keys apply, the reply stays **nil** (not OK), but keys that did change still get keyspace notifications and `apply_changes` so replication matches applied data. **JSON.MSET** tracks per-key success and only returns OK when every triplet fully succeeds.
> 
> Pytests cover partial **MSET**, keyspace events, a deep **JSON.SET** multi-match case, and depth-limit **MSET** overall failure. Linux/macOS CI and `tests.sh` enable parallel RLTest (`PARALLEL=1`), including removing the macOS-only serial override.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec4a0ef8bdae326c51ebd703f7b7a32d68a36594. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/flow-macos.yml
+++ b/.github/workflows/flow-macos.yml
@@ -138,7 +138,7 @@ jobs:
             export BINDGEN_EXTRA_CLANG_ARGS="-isysroot ${_SDK}"
             export CC=/usr/bin/clang CXX=/usr/bin/clang++
           fi
-          make test 2>&1 | tee test_output.log
+          PARALLEL=1 make test 2>&1 | tee test_output.log
       - name: Capture test results
         if: always()
         uses: ./.github/actions/capture-test-results
--- a/redis_json/src/commands.rs
+++ b/redis_json/src/commands.rs
@@ -439,12 +439,19 @@ pub fn json_set_command_impl<M: Manager>(
                 }
             } else {
                 let update_info = KeyValue::new(doc).find_paths(path.get_path(), op)?;
-                if !update_info.is_empty() && apply_updates::<M>(&mut redis_key, val, update_info) {
-                    redis_key.notify_keyspace_event(ctx, "json.set")?;
-                    manager.apply_changes(ctx);
-                    REDIS_OK
-                } else {
+                if update_info.is_empty() {
                     Ok(RedisValue::Null)
+                } else {
+                    let result = apply_updates::<M>(&mut redis_key, val, update_info);
+                    if result.any_updated() {
+                        redis_key.notify_keyspace_event(ctx, "json.set")?;
+                        manager.apply_changes(ctx);
+                    }
+                    if result.all_updated() {
+                        REDIS_OK
+                    } else {
+                        Ok(RedisValue::Null)
+                    }
                 }
             }
         }
@@ -714,7 +721,8 @@ pub fn json_mset_command_impl<M: Manager>(
             let value = manager.from_str(&value_str, Format::JSON, true, None)?;
 
             let updated = if let Some(update_info) = update_info {
-                !update_info.is_empty() && apply_updates::<M>(&mut redis_key, value, update_info)
+                !update_info.is_empty()
+                    && apply_updates::<M>(&mut redis_key, value, update_info).any_updated()
             } else {
                 // In case it is a root path
                 redis_key.set_value(Vec::new(), value)?
@@ -729,25 +737,60 @@ pub fn json_mset_command_impl<M: Manager>(
     res
 }
 
+enum ApplyUpdatesResult {
+    NoneUpdated,
+    SomeUpdated,
+    AllUpdated,
+}
+
+impl ApplyUpdatesResult {
+    fn combine(self, updated: bool) -> Self {
+        match (self, updated) {
+            (Self::NoneUpdated, false) => Self::NoneUpdated,
+            (Self::NoneUpdated | Self::AllUpdated, true) => Self::AllUpdated,
+            (Self::AllUpdated, false) | (Self::SomeUpdated, _) => Self::SomeUpdated,
+        }
+    }
+
+    fn any_updated(&self) -> bool {
+        !matches!(self, Self::NoneUpdated)
+    }
+
+    fn all_updated(&self) -> bool {
+        matches!(self, Self::AllUpdated)
+    }
+}
+
+impl From<bool> for ApplyUpdatesResult {
+    fn from(value: bool) -> Self {
+        if value {
+            Self::AllUpdated
+        } else {
+            Self::NoneUpdated
+        }
+    }
+}
+
 fn apply_updates<M: Manager>(
     redis_key: &mut M::WriteHolder,
     value: M::O,
     mut update_info: Vec<UpdateInfo>,
-) -> bool {
+) -> ApplyUpdatesResult {
     // If there is only one update info, we can avoid cloning the value
     if update_info.len() == 1 {
         match update_info.pop().unwrap() {
             UpdateInfo::SUI(sui) => redis_key.set_value(sui.path, value),
             UpdateInfo::AUI(aui) => redis_key.dict_add(aui.path, &aui.key, value),
         }
-        .unwrap_or(false)
+        .unwrap_or(false).into()
     } else {
-        update_info.into_iter().fold(false, |updated, ui| {
-            match ui {
+        update_info.into_iter().fold(ApplyUpdatesResult::NoneUpdated, |acc, ui| {
+            let updated = match ui {
                 UpdateInfo::SUI(sui) => redis_key.set_value(sui.path, value.clone()),
                 UpdateInfo::AUI(aui) => redis_key.dict_add(aui.path, &aui.key, value.clone()),
             }
-            .map_or(updated, |v| v || updated)
+            .unwrap_or(false);
+            acc.combine(updated)
         })
     }
 }
--- a/redis_json/src/commands.rs
+++ b/redis_json/src/commands.rs
@@ -442,7 +442,10 @@ pub fn json_set_command_impl<M: Manager>(
                 if update_info.is_empty() {
                     Ok(RedisValue::Null)
                 } else {
-                    let result = apply_updates::<M>(&mut redis_key, val, update_info);
+                    let result: ApplyUpdatesResult =
+                        apply_updates::<M>(&mut redis_key, val, update_info);
+                    // If any path is updated, notify the keyspace event
+                    // But only return OK if all paths are updated, otherwise return null
                     if result.any_updated() {
                         redis_key.notify_keyspace_event(ctx, "json.set")?;
                         manager.apply_changes(ctx);
--- a/redis_json/src/commands.rs
+++ b/redis_json/src/commands.rs
@@ -746,6 +746,7 @@ pub fn json_mset_command_impl<M: Manager>(
     }
 }
 
+#[allow(clippy::enum_variant_names)]
 enum ApplyUpdatesResult {
     NoneUpdated,
     SomeUpdated,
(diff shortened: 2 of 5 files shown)


redis_json/src/commands.rs
```diff
@@ -439,12 +439,19 @@ pub fn json_set_command_impl<M: Manager>(
...
-                } else {
+                if update_info.is_empty() {
                     Ok(RedisValue::Null)
+                } else {
+                    let result = apply_updates::<M>(&mut redis_key, val, update_info);
+                    if result.any_updated() {
+                        redis_key.notify_keyspace_event(ctx, "json.set")?;
+                        manager.apply_changes(ctx);
+                    }
+                    if result.all_updated() {
```
TalBarYakar, line 453: write some comments to explain the return value paths

redis_json/src/commands.rs
```diff
@@ -706,49 +713,110 @@ pub fn json_mset_command_impl<M: Manager>(
...
+        } else {
+            // In case it is a root path
+            let updated = redis_key.set_value(Vec::new(), value)?;
+            (updated, updated)
+        };
+
+        if any_updated {
+            redis_key.notify_keyspace_event(ctx, "json.mset")?;
+        }
+        all_updated &= key_all_updated;
```
TalBarYakar, line 741: Why do we need the all_updated logic ?
if you fail to do something , isn't it good enough to just return a mid loop error and thats it ?
if we finish all the loop we can retun 'REDIS_OK'

AvivDavid23: This would be a breaking change.. though I agree with you

redis_json/src/commands.rs
AvivDavid23: Its literarily what the code says.. If all updates were to all paths Ok, return Ok, else RedisValue::Null

redis_json/src/commands.rs
TalBarYakar: @gabsow WDYT ?

redis_json/src/commands.rs
TalBarYakar: I prefer having a comment at the function header so its purpose is clear without needing to jump into the implementation.

redis_json/src/commands.rs
AvivDavid23: @gabsow for context, let say we have:
`JSON.MSET a $ '{"a":"b"}' b $ wada c $ '{"a":"b"}'`
currently, setting a will work, b will fail but we will continue to c,
Tal is suggesting stop after b fail 
(in practice, *in this example*. nothing will run because we validate the input first, including the json given but lets assume we are not for this example)

redis_json/src/commands.rs
```diff
@@ -706,49 +716,110 @@ pub fn json_mset_command_impl<M: Manager>(
...
+    }
 
     manager.apply_changes(ctx);
```
gabsow, line 744: **Correctness — partial writes may not be replicated on a mid-loop error.** The loop now uses `?` directly (`open_key_write`/`from_str` on 721-723, and the root-path `set_value?` on 734), so any error early-returns from the whole command. But by then earlier keys in `parsed` may already have been mutated in place and had `notify_keyspace_event` fired, while `manager.apply_changes(ctx)` (`replicate_verbatim`) is only reached *after* the loop. An error on key N>1 therefore leaves the master mutated but the replica/AOF never receives those writes → divergence. The old `fold` always ran `apply_changes` afterwards, so this is a regression. Validating/opening all keys up front (so the apply loop is infallible) — or running `apply_changes` before propagating the error — would close the gap.

redis_json/src/commands.rs
```diff
@@ -706,49 +716,110 @@ pub fn json_mset_command_impl<M: Manager>(
...
+                    UpdateInfo::SUI(sui) => redis_key.set_value(sui.path, value.clone()),
+                    UpdateInfo::AUI(aui) => redis_key.dict_add(aui.path, &aui.key, value.clone()),
+                }
+                .unwrap_or(false)
+                .into()
+            })
+            .reduce(ApplyUpdatesResult::combine)
+            // SAFETY: `reduce` is guaranteed to return a value
+            // because the iterator is not empty
+            .unwrap()
```
gabsow: **Robustness — `apply_updates` now panics on empty input.** The `else` branch handles `update_info.len() != 1`, which includes `len() == 0`, and `.reduce(…).unwrap()` panics on an empty iterator. It's safe *today* only because both call sites guard with `is_empty()` first; the old `fold(false, …)` was total and returned `false` for the empty case. A future caller that forgets the guard gets a panic instead of `NoneUpdated`. Prefer making the function total:

```suggestion
            .reduce(ApplyUpdatesResult::combine)
            .unwrap_or(ApplyUpdatesResult::NoneUpdated)
```

redis_json/src/commands.rs
```diff
@@ -706,49 +716,110 @@ pub fn json_mset_command_impl<M: Manager>(
...
+
+        if any_updated {
+            redis_key.notify_keyspace_event(ctx, "json.mset")?;
+        }
+        all_updated &= key_all_updated;
+    }
 
     manager.apply_changes(ctx);
-    res
+    if all_updated {
```
gabsow, line 745: **Semantics — a `nil` reply here no longer means "nothing changed."** Re: the `all_updated` discussion above — worth deciding two consequences deliberately, since the tests assert that MSET (and SET, see line 453) can return `nil` while having already mutated **and replicated** a subset of keys:

- `nil` has historically meant "no write happened" (e.g. NX/XX miss). Overloading it to also mean "partial write happened" is ambiguous for clients.
- A client that retries on `nil` re-applies the keys that already succeeded — not idempotent for append/merge-style paths.

If the intent is all-or-nothing, the failing command's writes shouldn't be committed/replicated at all; if it's "best-effort + report," an *error* (rather than `nil`) would disambiguate success-with-partial-failure from a plain miss. Either way it's a client-visible contract change worth documenting.

redis_json/src/commands.rs
```diff
@@ -439,12 +439,22 @@ pub fn json_set_command_impl<M: Manager>(
...
+                } else {
+                    let result: ApplyUpdatesResult =
+                        apply_updates::<M>(&mut redis_key, val, update_info);
+                    // If any path is updated, notify the keyspace event
+                    // But only return OK if all paths are updated, otherwise return null
+                    if result.any_updated() {
+                        redis_key.notify_keyspace_event(ctx, "json.set")?;
+                        manager.apply_changes(ctx);
+                    }
+                    if result.all_updated() {
```
gabsow, line 453: **Behavior change — this narrows `JSON.SET`'s reply too, not only `JSON.MSET`.** The active thread is on the MSET side, but the same `all_updated` change applies here to `JSON.SET` with a multi-match path (e.g. `$..x`): previously it replied `OK` if at least one match was written; now it replies `nil` unless *every* matched path succeeds (while still writing + replicating the ones that did — see `testApplyUpdatesFoldDoesNotLoseEarlierSuccess`). That's a contract change for a very common command, and it isn't reflected in the PR title/description (scoped to `apply_updates`/MSET). Please confirm it's intended and document it.

.github/workflows/flow-macos.yml
```diff
@@ -138,7 +138,7 @@ jobs:
             export BINDGEN_EXTRA_CLANG_ARGS="-isysroot ${_SDK}"
             export CC=/usr/bin/clang CXX=/usr/bin/clang++
           fi
-          make test 2>&1 | tee test_output.log
+          PARALLEL=1 make test 2>&1 | tee test_output.log
```
gabsow, line 141: **CI risk — this runs the *entire* suite in parallel on macOS.** Together with dropping `[[ $OS == macos ]] && PARALLEL=0` in `tests.sh`, this forces `PARALLEL=1` on macOS. That guard's comment documented a real failure ("Can't pickle local object" in RLTest), and it was global — so *any* test in the suite that defines a local/nested function or lambda (not just the ones added here) can now break macOS CI under multiprocessing. If the underlying pickling issue is fixed, great; otherwise this re-introduces the flake the guard prevented. Worth confirming a full macOS run is green before merge.

tests/pytest/test.py
```diff
@@ -1375,18 +1375,87 @@ def testMSET(env):
...
+    # key a{s}: succeeds. key b{s}: path doesn't match anything -> fails.
+    # key c{s}: succeeds. Overall reply must be nil, but a{s}/c{s} must still
+    # be updated and notified/replicated -- only the reply reflects the
+    # all-or-nothing contract, not the underlying per-key writes.
+    env.expect("JSON.SET", "a{s}", '$', '{"x": 1}').ok()
+    env.expect("JSON.SET", "b{s}", '$', '{"x": 1}').ok()
+    env.expect("JSON.SET", "c{s}", '$', '{"x": 1}').ok()
+
+    with env.getClusterConnectionIfNeeded() as r:
+        r.execute_command('config', 'set', 'notify-keyspace-events', 'KEA')
```
gabsow, line 1402: **Test hygiene — `notify-keyspace-events` is set on the shared `env` and never restored.** `test_mset_replication_in_aof` does the same `config set … KEA` but on an isolated `Env(useAof=True)`; this test mutates the shared server's config (and leaves the `pubsub` subscribed) without resetting it, so later tests in the module run with keyspace notifications forced on. Consider restoring the previous value and closing/unsubscribing the pubsub in a `finally`, or using an isolated `Env`.

tests/pytest/test.py
```diff
@@ -1375,18 +1375,87 @@ def testMSET(env):
...
+        env.assertEqual(pubsub.get_message(timeout=1), None)
+
+def testApplyUpdatesFoldDoesNotLoseEarlierSuccess(env):
+    chain_depth = 100
+    doc = '{"a":{"x":0},"b":' + ('{"n":' * chain_depth) + '{"x":0}' + ('}' * chain_depth) + '}'
+    env.expect("JSON.SET", "foldbug", "$", doc).ok()
+
+    # A value nested 60 arrays deep (calculate_value_depth() == 60).
+    patch_nesting = 60
+    patch = ('[' * patch_nesting) + '0' + (']' * patch_nesting)
```
gabsow, line 1440: **Test brittleness — depth arithmetic is pinned to internal limit constants.** The 60/102/128 math (shallow `2+60=62 < 128` succeeds, deep `102+60=162 ≥ 128` fails) hard-codes the current recursion limit and depth counting. If either changes, both `$..x` matches will *both* succeed or *both* fail — the test then fails at the first `assertEqual(res, None)` or at the `$.a.x` GET rather than clearly signaling "the limit moved." Deriving the patch depth from the limit constant (or a comment naming it) would make the failure mode legible.

TalBarYakar (approved):

PR #1624, opened 2026-07-22
MOD-16952 - enhance bootstrap (add list and dry-run arg)

TalBarYakar:
> [!NOTE]
> **Medium Risk**
> Broad changes to multi-OS bootstrap and package install primitives could affect real installs via new idempotency paths and RHEL Python wiring, though list/dry-run are isolated read-only paths.
> 
> **Overview**
> Adds **read-only bootstrap modes** wired through the Makefile: `make bootstrap list` runs `install_script.sh` with `CHECK_DEPS=1` (reports required vs missing deps, optional deps separately, non-zero exit if required gaps remain) and **does not** remove `venv` or install anything; `make bootstrap dry-run` prints the exact commands that would run for still-missing pieces, also without mutating the host.
> 
> The install stack is refactored around **`CHECK_DEPS` / `DRY_RUN` in `pm.sh`**: package helpers record presence via `_check_pkgs` instead of installing; `_run` / `_sh` centralize execute vs print vs skip; `SUDO` is neutralized in check/dry-run so repo enables, symlinks, and profile edits cannot run. **`OPTIONAL_PKGS`** and sparse **`MIN_VERSIONS`** classify test/debug tooling vs hard requirements. **`install_script.sh`** skips `git safe.directory` in these modes, aggregates results to **`DEPS_REPORT_FILE`** when set, and handles Rust/cargo as check-or-print instead of invoking `getrust.sh`.
> 
> Per-OS scripts gain **idempotency guards** (skip PPA/cmake source builds, devtoolset/openssl/cmake symlinks, compiler alternatives when already satisfied) and route mutating steps through `_run`/`_sh`. Notable install-path tweaks: **RHEL python** is installed via `rhel_default_install` with an **EL8 `RHEL_PYTHON` override** so list/dry-run don’t false-flag `python3`; **Amazon Linux 2** gates x86_64-only SCL/devtoolset; **apt** gets **Acquire::Retries**; **root without sudo** gets a sudo pass-through stub.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2dc37d78bdd81880fd9e1448ce8ce47dbe55d4c0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.install/lib/pm.sh
+++ b/.install/lib/pm.sh
@@ -65,7 +65,7 @@ _min_for() { for _e in $MIN_VERSIONS; do case "$_e" in "$1:"*) echo "${_e#*:}"; 
 _get_installed_version() {
     case "$1" in
         gcc|g++) "$1" -dumpversion 2>/dev/null ;;
-        *)       "$1" --version 2>/dev/null | grep -oE '[0-9]+\.[0-9]+(\.[0-9]+)?' | head -1 ;;
+        *)       "$1" --version 2>/dev/null | grep -oE '[0-9]+\.[0-9]+(\.[0-9]+)?' | head -1 || true ;;
     esac
 }
 
--- a/.install/lib/pm.sh
+++ b/.install/lib/pm.sh
@@ -153,9 +153,9 @@ apt_install() {
     # connections mid-build; without retries a single dropped fetch fails the
     # whole docker build (exit 100). Retry each download before giving up.
     local apt_retry="-o Acquire::Retries=5"
-    if [ "$DRY_RUN" != 1 ] && [ "$_pm_apt_updated" = 0 ]; then
+    if [ "$_pm_apt_updated" = 0 ]; then
         export DEBIAN_FRONTEND=noninteractive
-        $SUDO apt-get update -qq $apt_retry
+        _run apt-get update -qq $apt_retry
         _pm_apt_updated=1
     fi
     # env goes THROUGH sudo: sudo's env_reset strips exported variables, so a
--- a/.install/lib/pm.sh
+++ b/.install/lib/pm.sh
@@ -54,6 +54,28 @@ DEPS_OPT_MISSING=""
 OPTIONAL_PKGS="${OPTIONAL_PKGS:-}"
 _is_optional() { case " $OPTIONAL_PKGS " in *" $1 "*) return 0 ;; *) return 1 ;; esac; }
 
+# MIN_VERSIONS (lib/packages.sh): sparse "pkg:minversion" list -- only deps
+# with a real floor. Everything else is presence-only. Default empty.
+MIN_VERSIONS="${MIN_VERSIONS:-}"
+_min_for() { for _e in $MIN_VERSIONS; do case "$_e" in "$1:"*) echo "${_e#*:}"; return ;; esac; done; }
+
+# Read a tool's installed version from the tool itself (not the package DB --
+# e.g. cmake is overlaid into /usr/local by install_cmake.sh, so dpkg would
+# report the stale apt version).
+_get_installed_version() {
+    case "$1" in
+        gcc|g++) "$1" -dumpversion 2>/dev/null ;;
+        *)       "$1" --version 2>/dev/null | grep -oE '[0-9]+\.[0-9]+(\.[0-9]+)?' | head -1 ;;
+    esac
+}
+
+# version_ge HAVE WANT -> 0 if HAVE >= WANT (strips -rev/+build suffixes).
+version_ge() {
+    _have="${1%%[-+]*}"; _want="${2%%[-+]*}"
+    _s=sort; sort -V </dev/null >/dev/null 2>&1 || { command -v gsort >/dev/null 2>&1 && _s=gsort; }
+    [ "$(printf '%s\n%s\n' "$_want" "$_have" | "$_s" -V | head -1)" = "$_want" ]
+}
+
 if [ "$CHECK_DEPS" = 1 ]; then
     SUDO=":"   # ':' is the shell no-op builtin; it ignores its arguments
 fi
@@ -77,7 +99,14 @@ _check_pkgs() {
         if _is_optional "$_p"; then
             if _pkg_installed "$_p"; then DEPS_OPT_OK="$DEPS_OPT_OK $_p"; else DEPS_OPT_MISSING="$DEPS_OPT_MISSING $_p"; fi
         else
-            if _pkg_installed "$_p"; then DEPS_OK="$DEPS_OK $_p"; else DEPS_MISSING="$DEPS_MISSING $_p"; fi
+            _min=$(_min_for "$_p")
+            if ! _pkg_installed "$_p"; then
+                if [ -n "$_min" ]; then DEPS_MISSING="$DEPS_MISSING $_p:$_min"; else DEPS_MISSING="$DEPS_MISSING $_p"; fi
+            elif [ -n "$_min" ] && _have=$(_get_installed_version "$_p") && [ -n "$_have" ] && ! version_ge "$_have" "$_min"; then
+                DEPS_MISSING="$DEPS_MISSING $_p:$_min"
+            else
+                DEPS_OK="$DEPS_OK $_p"
+            fi
         fi
     done
 }
--- a/.install/lib/pm.sh
+++ b/.install/lib/pm.sh
@@ -76,9 +76,33 @@ version_ge() {
     [ "$(printf '%s\n%s\n' "$_want" "$_have" | "$_s" -V | head -1)" = "$_want" ]
 }
 
-if [ "$CHECK_DEPS" = 1 ]; then
-    SUDO=":"   # ':' is the shell no-op builtin; it ignores its arguments
+# DRY_RUN=1: run the bootstrap flow but install nothing — for each MISSING
+# package, print the exact install command that WOULD run (the same single
+# line the primitive executes normally, via _run — no duplicated command).
+DRY_RUN="${DRY_RUN:-0}"
+
+if [ "$CHECK_DEPS" = 1 ] || [ "$DRY_RUN" = 1 ]; then
+    _SUDO_DISPLAY="$SUDO"   # remember the real sudo prefix for dry-run printing
+    SUDO=":"                # neutralize privileged side-commands (no mutation)
 fi
+
+# dry-run output is blue on a real terminal, plain when piped (CI logs).
+if [ "$DRY_RUN" = 1 ] && [ -t 1 ]; then _DRY_C="$(printf '\033[1;34m')"; _DRY_R="$(printf '\033[0m')"; else _DRY_C=""; _DRY_R=""; fi
+_dry_line() { printf '%s%s%s\n' "$_DRY_C" "$*" "$_DRY_R"; }
+
+# _run CMD... — one wrapper for every "would-install" command, so callers
+# never branch on the mode:
+#   install  -> execute it (with the real sudo prefix)
+#   dry-run  -> print it (blue), don't execute
+#   list     -> skip it (a check neither installs nor prints)
+_run() {
+    if [ "$CHECK_DEPS" = 1 ]; then return 0
+    elif [ "$DRY_RUN" = 1 ]; then _dry_line "${_SUDO_DISPLAY:+$_SUDO_DISPLAY }$*"
+    else ${_SUDO_DISPLAY:-$SUDO} "$@"; fi
+}
+
+# Echo (space-separated) only the packages from "$@" that are NOT installed.
+_missing_only() { for _p in "$@"; do _pkg_installed "$_p" || printf '%s ' "$_p"; done; }
 
 # Read-only "is this package installed?" probe, per package manager.
 _pkg_installed() {
@@ -121,16 +145,17 @@ _pm_apt_updated=0
 apt_install() {
     [ "$#" -gt 0 ] || return 0
     if [ "$CHECK_DEPS" = 1 ]; then _check_pkgs "$@"; return 0; fi
-    if [ "$_pm_apt_updated" = 0 ]; then
+    if [ "$DRY_RUN" = 1 ]; then set -- $(_missing_only "$@"); [ "$#" -gt 0 ] || return 0; fi
+    if [ "$DRY_RUN" != 1 ] && [ "$_pm_apt_updated" = 0 ]; then
         export DEBIAN_FRONTEND=noninteractive
         $SUDO apt-get update -qq
         _pm_apt_updated=1
     fi
-    # env goes THROUGH $SUDO: sudo's env_reset strips exported variables, so a
+    # env goes THROUGH sudo: sudo's env_reset strips exported variables, so a
     # plain export upstream never reaches dpkg — debconf (e.g. tzdata on focal,
     # which the base image doesn't preinstall) then blocks on an interactive
     # prompt and the bootstrap hangs.
-    $SUDO env DEBIAN_FRONTEND=noninteractive apt-get install -yqq --no-install-recommends "$@"
+    _run env DEBIAN_FRONTEND=noninteractive apt-get install -yqq --no-install-recommends "$@"
 }
 
 # `--allowerasing` lets dnf pick our `curl` over the slimmer `curl-minimal`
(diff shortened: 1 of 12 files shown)


.install/lib/pm.sh
```diff
@@ -34,6 +34,110 @@ else
...
+}
+
+# DRY_RUN=1: run the bootstrap flow but install nothing — for each MISSING
+# package, print the exact install command that WOULD run (the same single
+# line the primitive executes normally, via _run — no duplicated command).
+DRY_RUN="${DRY_RUN:-0}"
+
+if [ "$CHECK_DEPS" = 1 ] || [ "$DRY_RUN" = 1 ]; then
+    _SUDO_DISPLAY="$SUDO"   # remember the real sudo prefix for dry-run printing
+    SUDO=":"                # neutralize privileged side-commands (no mutation)
```
gabsow, line 95: **`list`/`dry-run` are not actually read-only, and both under-report — the safety model misses the per-OS scripts.**

The mode machinery neutralizes only two things: `$SUDO` → `:` and the `*_install` primitives. Any command in an `os/<osnick>.sh` file (or `lib/packages.sh`) that touches the network/disk/pip **directly** bypasses both and runs for real under `make bootstrap list` / `dry-run`:

- `install_aws_cli` (`lib/packages.sh`) — `curl … -o /tmp/awscliv2.zip` + `unzip` execute; only the trailing `$SUDO …/aws/install` is neutralized. Reached from `os/rocky8.sh`, `rocky9.sh`, `mariner2.sh`, `azurelinux3.sh`, `bionic.sh`.
- `os/bionic.sh` — builds CMake from source (`wget` + `tar` + `./configure` + `make -j`) and runs `pip3 install dataclasses` (redis `scripts/bootstrap.sh` exports `PIP_BREAK_SYSTEM_PACKAGES=1`), plus two keyserver `wget`s.

Two more consequences of the same design:
- **Abort on a fresh host:** `os/focal.sh:12` / `os/bionic.sh:27` run `_cur=$(gcc -dumpversion | cut -d. -f1)` ungated. In check mode gcc was only *recorded*, not installed, so on a clean `ubuntu:20.04`/`18.04` the pipeline exits 127 and `set -euo pipefail` kills the script **before any report prints** — the exact fresh-host case `list` targets. (Same shape: `_pm_enable_el8_extras` does `return 1` on subscription-less RHEL 8.)
- **Under-reporting:** because `SUDO=":"` swallows every `$SUDO`-prefixed non-primitive, `list` never records `epel-release`, the `Development Tools` groupinstall, or the gcc-toolset symlinks/`update-alternatives`, and `dry-run`'s "commands above are what bootstrap would run" is false for those lines.

Probably needs a mode-aware wrapper (a `_run`-style guard, or gating) around the raw `curl`/`wget`/`pip`/`make`/tool-probe calls in `os/*.sh`, not just `$SUDO`.

.install/os/macos.sh
```diff
@@ -11,8 +11,7 @@ fi
 brew_default_install
 
 if ! command -v python3 >/dev/null 2>&1; then
-    echo "==> [redisjson] python3 not on PATH; installing brew python@3.11"
-    HOMEBREW_NO_AUTO_UPDATE=1 brew install python@3.11
+    HOMEBREW_NO_AUTO_UPDATE=1 _run brew install python@3.11
```
gabsow: **macOS `list` never reports a missing `python3` (dry-run does) — the two modes disagree.**

`python` is deliberately absent from `BREW_BASE`, and this conditional routes through `_run`, which is a no-op under `CHECK_DEPS` that records nothing into `DEPS_*`. So on a Mac with no `python3`, `make bootstrap list` reports everything present and exits 0, while `make bootstrap dry-run` *does* print `brew install python@3.11`. `list` should record python3 as a (missing) dependency so it matches dry-run and the real install. (Minor, same line: the dry-run print drops the `HOMEBREW_NO_AUTO_UPDATE=1` prefix, so the copy-paste command differs from what bootstrap runs.)

.install/lib/pm.sh
```diff
@@ -43,28 +147,34 @@ fi
...
         $SUDO apt-get update -qq
         _pm_apt_updated=1
     fi
-    # env goes THROUGH $SUDO: sudo's env_reset strips exported variables, so a
+    # env goes THROUGH sudo: sudo's env_reset strips exported variables, so a
     # plain export upstream never reaches dpkg — debconf (e.g. tzdata on focal,
     # which the base image doesn't preinstall) then blocks on an interactive
     # prompt and the bootstrap hangs.
-    $SUDO env DEBIAN_FRONTEND=noninteractive apt-get install -yqq --no-install-recommends "$@"
+    _run env DEBIAN_FRONTEND=noninteractive apt-get install -yqq --no-install-recommends "$@"
```
gabsow: **JSON's `apt_install` drops the `Acquire::Retries=5` that RedisTimeSeries/RedisBloom carry.** Their `apt_install` passes `-o Acquire::Retries=5` on both `apt-get update` and `apt-get install` (comment there: ports.ubuntu.com's arm64 mirror intermittently drops connections and fails the whole docker build, exit 100). JSON passes neither, so JSON's bootstrap is more prone to transient apt failures than its siblings — and the `dry-run` output prints the retry-less command too. Suggest matching the other two modules.

.install/lib/pm.sh
TalBarYakar: fixed

.install/os/macos.sh
TalBarYakar: fixed

.install/lib/pm.sh
TalBarYakar: done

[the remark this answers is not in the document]
TalBarYakar: fixed

.install/os/bionic.sh
```diff
@@ -35,17 +38,23 @@ if [ "$_cur" -lt 10 ]; then
...
+        make -j"$(nproc)"
+        $SUDO make install
+        cd /
+        rm -rf /tmp/cmake-3.28.0 /tmp/cmake-3.28.0.tar.gz
+        $SUDO ln -sf /usr/local/bin/cmake /usr/bin/cmake
+    fi
+    pip3 install dataclasses
+elif [ "$DRY_RUN" = 1 ]; then
+    _dry_line "build cmake 3.28 from source (if cmake < 3.28), then \$SUDO make install"
+    _dry_line "pip3 install dataclasses"
```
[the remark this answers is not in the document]
TalBarYakar: fixed

.install/lib/pm.sh
```diff
@@ -34,6 +34,110 @@ else
...
+        else
+            _min=$(_min_for "$_p")
+            if ! _pkg_installed "$_p"; then
+                if [ -n "$_min" ]; then DEPS_MISSING="$DEPS_MISSING $_p:$_min"; else DEPS_MISSING="$DEPS_MISSING $_p"; fi
+            elif [ -n "$_min" ] && _have=$(_get_installed_version "$_p") && [ -n "$_have" ] && ! version_ge "$_have" "$_min"; then
+                DEPS_MISSING="$DEPS_MISSING $_p:$_min"
+            else
+                DEPS_OK="$DEPS_OK $_p"
+            fi
+        fi
```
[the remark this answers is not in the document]
TalBarYakar: Intentional, not a bug. These names are in the *_BASE install superset but aren't actual build/link requirements — the module builds and runs without them (verified), so they're classified optional on purpose: list should only fail on deps that genuinely break build/run. Marking them required would make list fail on machines that build fine. OPTIONAL_PKGS deliberately holds only non-build-critical names (test/coverage/debug + install-superset extras).

.install/lib/setup-python.sh
```diff
@@ -12,8 +12,31 @@
...
+_have_uv() { command -v uv >/dev/null 2>&1 || [ -x "$HOME/.local/bin/uv" ] || [ -x "$HOME/.cargo/bin/uv" ]; }
+
+if [ "${CHECK_DEPS:-0}" = 1 ]; then
+    # uv presence, routed through OPTIONAL_PKGS like any other dep.
+    if _have_uv; then _uv=ok; else _uv=missing; fi
+    if _is_optional uv; then
+        [ "$_uv" = ok ] && DEPS_OPT_OK="$DEPS_OPT_OK uv" || DEPS_OPT_MISSING="$DEPS_OPT_MISSING uv"
+    else
+        [ "$_uv" = ok ] && DEPS_OK="$DEPS_OK uv" || DEPS_MISSING="$DEPS_MISSING uv"
+    fi
```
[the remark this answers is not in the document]
TalBarYakar: uv isn't a pre-requisite — a real make bootstrap installs it itself (curl … astral.sh/uv/install.sh), same as it provisions cmake/rust/aws-cli. The hard-fail in setup-python.sh only fires if that install fails, not when uv is simply absent. So list correctly treats uv as optional: a missing uv doesn't block bootstrap, and flagging it required would fail the check on every clean host even though bootstrap proceeds fine.

.install/lib/pm.sh
gabsow: Main items here are fixed in the latest commit ✅ — the `gcc -dumpversion` probe no longer aborts (`… 2>/dev/null | … || echo 0`), and the raw `install_aws_cli` / cmake-from-source / `pip3 install` / keyserver-`wget` mutations are now gated (dry-run prints them; aws-cli is recorded as optional).

One minor, non-blocking gap remains: `list`/`dry-run` still don't record the `$SUDO`-neutralized side-commands on EL8/9 — `epel-release`, the `Development Tools` groupinstall, and the gcc-toolset symlinks — so they won't appear in the report. Fine to leave as a follow-up.

.install/os/macos.sh
gabsow: Fixed ✅ — `macos.sh` now records `python@3.11` into `DEPS_OK`/`DEPS_MISSING` under list and prints the install command under dry-run, so the two modes agree.

.install/lib/pm.sh
gabsow: Fixed ✅ — `apt_install` now sets and passes `-o Acquire::Retries=5`, matching RedisTimeSeries/RedisBloom.

gabsow (dismissed):
Re-reviewed after the latest commit — all my earlier review points are addressed. LGTM 🚀

.install/lib/pm.sh
```diff
@@ -34,6 +34,143 @@ else
...
+        esac
+    else ${_SUDO_DISPLAY:-$SUDO} "$@"; fi
+}
+
+# _sh 'RAW SHELL' — a mutating step that doesn't fit _run (pipes, redirects,
+# cd/&&-chains, source builds). dry-run: print it verbatim so the output stays a
+# copy-pasteable script; list: skip; else: execute via eval. Write the string
+# with a literal `sudo` (these raw steps are Linux-only, where bootstrap ensures
+# sudo) so the printed line runs as-is.
+_sh() {
```
gabsow, line 136: Minor / non-blocking: `_sh` hardcodes a literal `sudo` in its command strings, whereas `_run` uses the MODE-aware prefix (`${_SUDO_DISPLAY:-$SUDO}`). These `_sh 'sudo …'` steps replace former `$SUDO …` calls, so on a Linux-root environment run with `INSTALL_SCRIPT_MODE=""` and no `sudo` binary they'd fail (`sudo: command not found`) where the old `$SUDO`-based code ran the command directly. Not a blocker (the redis dispatcher installs sudo when root on Linux, and MODE defaults to `sudo`) — but threading the real prefix through `_sh` like `_run` does would make it MODE-consistent.

.install/lib/pm.sh
```diff
@@ -143,24 +301,30 @@ _pm_enable_el8_extras() {
...
+            # before the diagnostic below can run.
+            local rid=$(dnf repolist --all 2>/dev/null \
                 | grep -i 'codeready-builder-for-rhel-8' \
                 | grep -vi source \
                 | head -1 \
                 | awk '{print $1}')
             if [ -n "$rid" ]; then
-                $SUDO dnf config-manager --set-enabled "$rid"
+                dnf repolist --enabled 2>/dev/null | grep -q "$rid" \
+                    || _run dnf config-manager --set-enabled "$rid"
```
gabsow, line 326: Minor / non-blocking (niche): a few lines below, the `return 1` when no CodeReady repo is found makes `_pm_enable_el8_extras` exit non-zero, and `el8_default_install` calls it unguarded — so under `set -euo pipefail` that aborts `make bootstrap list` before the dependency report prints. Only reachable on a *subscription-less real RHEL 8* (Alma/Rocky report `ID=almalinux`/`rocky` and skip this branch, so CI is unaffected). A `|| true` at the call site would let `list` degrade gracefully.

gabsow (dismissed):
Re-reviewed the latest revision. The read-only routing via `_sh`/`_run`, the idempotency guards, and the arm64 Amazon Linux 2 fix all look good, and the earlier follow-up is resolved — `list`/`dry-run` now report (and print) the side-commands that were previously swallowed by `SUDO=":"`, with nothing mutating. LGTM 🚀

Left two minor, non-blocking notes inline (`_sh` hardcoded `sudo` vs MODE-aware prefix; `_pm_enable_el8_extras` `return 1` on subscription-less RHEL 8) — neither needs to hold the merge.

.install/lib/pm.sh
TalBarYakar: Acknowledged — very minor. Only reachable on subscription-less real RHEL 8 (Alma/Rocky skip the ID=rhel branch, so CI is unaffected), where CodeReady Builder is genuinely required for gcc-toolset-11, so the hard abort with the CRB diagnostic is acceptable. Leaving as-is.

.install/lib/pm.sh
TalBarYakar: Acknowledged — very minor / niche. Only reachable on subscription-less real RHEL 8 (Alma/Rocky skip the ID=rhel branch, so CI is unaffected), where CodeReady Builder is genuinely required for gcc-toolset-11. The hard abort with the "fix RHUI/subscription repos" diagnostic is acceptable there, so leaving as-is rather than adding || true.

.install/lib/pm.sh
```diff
@@ -34,6 +34,143 @@ else
...
+# with a real floor. Everything else is presence-only. Default empty.
+MIN_VERSIONS="${MIN_VERSIONS:-}"
+_min_for() { for _e in $MIN_VERSIONS; do case "$_e" in "$1:"*) echo "${_e#*:}"; return ;; esac; done; }
+
+# Read a tool's installed version from the tool itself (not the package DB --
+# e.g. cmake is overlaid into /usr/local by install_cmake.sh, so dpkg would
+# report the stale apt version).
+_get_installed_version() {
+    case "$1" in
+        gcc|g++) "$1" -dumpversion 2>/dev/null ;;
```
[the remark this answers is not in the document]
TalBarYakar: Acknowledged — low severity, leaving as-is. Latent here since MIN_VERSIONS="", so the gcc|g++ return value is never version-compared. It would only bite on a broken/unrunnable gcc that the package DB still reports as installed — a degenerate case where make bootstrap list exiting non-zero is acceptable. Not worth the change for that edge.

.install/lib/pm.sh
```diff
@@ -172,18 +336,35 @@ _pm_enable_el8_extras() {
 # download its own 3.12, then psutil's wheel-less aarch64 source build would
 # fail looking for Python.h that matches the wrong interpreter.
 el8_default_install() {
-    $SUDO dnf -y install epel-release
+    rpm -q epel-release >/dev/null 2>&1 || _run dnf -y install epel-release
```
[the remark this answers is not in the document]
TalBarYakar: Acknowledged — very minor, same as the related el8 threads. Only reachable on subscription-less real RHEL 8 (Alma/Rocky skip the ID=rhel branch, so CI is unaffected), where CodeReady Builder is required for gcc-toolset-11 regardless. The hard abort with the CRB diagnostic is acceptable there, so leaving as-is.

.install/lib/pm.sh
gabsow, line 136: (Re-posting after the rebase orphaned it — non-blocking.) `_sh` hardcodes a literal `sudo`, whereas `_run` uses the MODE-aware prefix (`${_SUDO_DISPLAY:-$SUDO}`). On a Linux-root env run with `INSTALL_SCRIPT_MODE=""` and no `sudo` binary, these `_sh 'sudo …'` steps would fail where the old `$SUDO`-based code ran directly. Not a blocker (bootstrap ensures sudo when root; MODE defaults to `sudo`) — threading the real prefix through `_sh` would make it MODE-consistent.

.install/lib/pm.sh
gabsow, line 326: (Re-posting after the rebase orphaned it — non-blocking, niche.) The `return 1` when no CodeReady repo is found makes `_pm_enable_el8_extras` exit non-zero, and `el8_default_install` calls it unguarded, so under `set -euo pipefail` that aborts `make bootstrap list` before the report. Only on subscription-less real RHEL 8 (Alma/Rocky skip this branch). A `|| true` at the call site would let `list` degrade gracefully.

gabsow (dismissed):
Re-approving after the rebase — I diffed every file each PR touches at my previously-approved commit vs this rebased head and they are byte-identical, so this is a pure rebase with no content change (which is what dismissed the approval). Nothing new to review. LGTM 🚀

The two minor, non-blocking notes from before still apply (re-posted inline): `_sh` literal `sudo` vs MODE-aware prefix; `_pm_enable_el8_extras` `return 1` on subscription-less RHEL 8.

gabsow (approved):

PR #1630, opened 2026-08-20
Index the backing slice in IValue::get_index instead of iter().nth()

gabsow:
Follow-up to RED-213492, split out of #1629.

## Please read this first — it is not the fix the ticket asks for

The ticket hypothesis was that `getAt(index)` on packed typed arrays became linear, turning a per-element vector read into O(dim²). **I measured it and that is not happening in the shipped module.** Server-side `usec_per_call` on a release build of `librejson.so`, 65536-element packed f32 array:

| | usec_per_call |
| --- | --- |
| `LLAPI.GETAT` index 0 | 8.59 |
| `LLAPI.GETAT` index 65535 | 8.59 |

Identical. Per-element cost across a full scan is flat too — 66ns/element at dim 1280, 92ns at 8192, 58ns at 65536 — so total cost is linear in elements read, not quadratic. The optimizer collapses the `nth` skip loop, because `next()` on the inner `std::slice::Iter` is a pointer bump and the discarded item construction is side-effect-free.

So: this PR does not speed up production, and the slow vector loads in the ticket are not caused by indexed-access complexity. That still needs a separate answer.

## What this PR is for

`arr.iter().nth(index)` is O(index) *by contract* — `ijson::array::ArrayIter` implements only `next()`. Today it is fast because of an optimization we do not control, and that optimization already fails in another real build of this workspace. Same source, `cargo test --release -p json_path`, single read of the last element:

| n | index 0 | last index | ratio |
| --- | --- | --- | --- |
| 1280 | 18.67ns | 19,167.69ns | ~1,000x |
| 65536 | 17.70ns | 966,498.40ns | ~54,000x |

A full index-by-index scan there is genuinely quadratic. Whether that cliff shows up depends on inlining, so any future change to codegen settings, crate boundaries, or how `get_index` is reached could surface it in the module too.

## The change

Index the backing slice directly via `as_slice()`: borrow the element for heterogeneous arrays, build the number for the 12 packed tags. Cost is index-independent by construction instead of by optimizer behaviour.

Measured after the change:

- the slow build above: 17.74ns for the last element at n=65536 (was 966,498ns)
- the shipped module: unchanged within noise — scan of 1280 elements 84.93µs before, 85.62µs after; `GETAT` 8.59µs → 8.38/8.74µs

Same `get_index` also backs JSONPath index/slice evaluation and `JSON.ARRINDEX`, so those get the same guarantee.

## Tests

- `json_node.rs`: indexing over all 12 packed array tags; heterogeneous arrays return a borrow rather than a clone (asserted by pointer identity); out-of-bounds and non-array cases; and a guard that per-element cost does not grow with array length (512 vs 8192 elements, best-of-5 passes, asserts under 4x where linear would be ~16x). That guard fails on the previous implementation in this build.
- `c_api.rs`: `JSONAPI_getAt` over every packed array tag, covering the LLAPI entry point RediSearch uses.

`cargo test -p json_path` (299 tests) and `-p redis_json` green; `cargo clippy --all-targets` clean on both crates.

## Reviewer judgement needed

1. **Is a no-op-in-production change worth taking?** The argument is robustness, not speed. If you would rather not carry a timing-sensitive test for a cliff that is currently latent, closing this is a reasonable call.
2. **The timing guard.** It is best-of-5 with a 4x threshold against an expected 16x, so it has headroom, but it is still wall-clock in a unit test and CI runners are noisy. Happy to drop it and keep only the deterministic per-type tests.
3. **`half` is added as a dev-dependency** of `json_path` (`Cargo.toml` + `Cargo.lock`) so the tests can construct f16/bf16 arrays.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AZUWRHkrd9XtKdKqENRs9r)_

---

> [!NOTE]
> **Medium Risk**
> Touches the shared array-index path used by the LLAPI and JSONPath. Behavior should be equivalent, but packed-array element wrapping and a timing-sensitive unit test are the main review points.
> 
> **Overview**
> Makes `IValue::get_index` index the packed/heterogeneous backing slice directly instead of `iter().nth(index)`, so cost is O(1) by construction rather than depending on the optimizer collapsing a linear skip.
> 
> That path backs LLAPI `JSONAPI_getAt` (RediSearch vector reads), JSONPath index/slice, and `JSON.ARRINDEX`. Heterogeneous elements stay borrowed; packed tags wrap a primitive into an owned `IValue`.
> 
> Adds coverage for all 12 packed tags, borrow identity on mixed arrays, out-of-bounds/non-array, a wall-clock guard that per-element cost stays flat (512 vs 8192), and matching `json_api_get_at` tests. `half` is a `json_path` dev-dependency for f16/bf16 fixtures.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9212a3475a2727641c5a227f24b78df21f6eda30. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -602,6 +602,7 @@ version = "0.1.0"
 dependencies = [
  "bson",
  "env_logger",
+ "half",
  "ijson",
  "itertools",
  "log",
--- a/json_path/Cargo.toml
+++ b/json_path/Cargo.toml
@@ -18,6 +18,7 @@ regex = "1"
 itertools = "0.13"
 
 [dev-dependencies]
+half = "2.0.0"
 env_logger = "0.10" # do not change this version without running a full ci cycle
 
 [[bin]]
--- a/json_path/src/json_node.rs
+++ b/json_path/src/json_node.rs
@@ -232,8 +232,23 @@ impl SelectValue for IValue {
     }
 
     fn get_index<'a>(&'a self, index: usize) -> Option<ValueRef<'a, Self>> {
-        self.as_array()
-            .and_then(|arr| arr.iter().nth(index).map(Into::into))
+        use ijson::array::ArraySliceRef;
+        let arr = self.as_array()?;
+        // Index straight into the backing slice. `ArrayIter` does not override
+        // `Iterator::nth`, so `iter().nth(index)` walks the array element by element,
+        // which makes an index-by-index pass over an N-element array O(N^2) --
+        // painfully visible for high-dimension vectors read through the LLAPI.
+        macro_rules! get_indexed {
+            ($($variant:ident),*) => {
+                match arr.as_slice() {
+                    ArraySliceRef::Heterogeneous(s) => s.get(index).map(ValueRef::Borrowed),
+                    $(ArraySliceRef::$variant(s) => {
+                        s.get(index).map(|&v| ValueRef::Owned(IValue::from(v)))
+                    })*
+                }
+            };
+        }
+        get_indexed!(I8, U8, I16, U16, F16, BF16, I32, U32, F32, I64, U64, F64)
     }
 
     fn is_array(&self) -> bool {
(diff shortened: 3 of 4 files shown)


AvivDavid23:
Solved in https://github.com/RedisJSON/RedisJSON/pull/1631, closing

PR #1638, opened 2026-09-15
MOD-18184 Gate benchmarks on committed KPI baselines

LiranAbir:
[MOD-18184](https://redislabs.atlassian.net/browse/MOD-18184) — follow-up to [MOD-17912](https://redislabs.atlassian.net/browse/MOD-17912) (`get_index` regression), which we had no CI gate to catch.

## What this does

Lets each `tests/benchmarks/<test>.yml` carry its own performance baseline as a redisbench-admin `kpis` floor:

```yaml
kpis:
  - ge:
      "$.Tests.Overall.rps": 194750.47
```

No comparison code was needed — redisbench-admin already checks those floors at the end of every run (`results_dict_kpi_check` → `return_code |= 1` → `run_remote.py` `exit(return_code)`). Verified against the installed package: a run above the floor returns 0, a run below returns 1.

That means every lane which already runs the benchmarks now gates on the baseline: the `run-benchmark` PR label, pushes to `master` / `feature-*` / `X.Y`, and the nightly.

Worth noting for reviewers: `redisbench-admin compare`, which this flow already ran, **cannot** serve as the gate — `cli.py` discards its return value, so it never fails a job regardless of how many regressions it prints.

## Changes

- **`benchmark-flow.yml`** — dropped `continue-on-error: true` from the run step. It was swallowing the exit code, which would have left the gate decorative. Added an upload of each group's result json files as a `benchmark-results-<group>` artifact (30 day retention).
- **`event-nightly.yml`** — added the `benchmark` job, and wired it into `test-summary`'s `needs` and `job-failures` string so a red baseline gate is named in the Slack summary instead of being visible only by drilling into the Actions run. (`memory-nightly-report` has the same gap today; left alone as out of scope.)
- **`tests/benchmarks/update_kpis.py`** — new. Reads the result json files a run leaves behind and raises each floor to `measured * (1 - margin)` (5% default). It can only **raise**: a slower run proposes nothing, so a baseline cannot drift downwards on its own, and lowering one is always a hand edit in a reviewed PR. `--self-test` covers raise / never-lower / no-duplicate-block.
- **`tests/benchmarks/Readme.md`** — documents the floors and the promotion flow.

## Baseline promotion

Baselines are never written by CI. A faster-than-baseline nightly is picked up by the nightly analysis, which runs `update_kpis.py` against that run's own numbers and opens a PR with the raised floors. A human merges it; leaving it unmerged keeps the current baselines.

## Not in this PR

Floors are intentionally **not seeded** — that needs measured values from a real run, and inventing numbers would give a meaningless gate. Until the first seeding PR merges, the gate is inert but harmless. The 5% margin is a starting point borrowed from `redisbench-admin compare`'s own regression waterline; it should be revisited against the real run-to-run spread on the [CI benchmarks dashboard](https://benchmarksrediscom.grafana.net/d/UErSC0jGk/redisjson-ci-benchmarks) once floors exist.

## Risk

Dropping `continue-on-error` also means remote EC2/provisioning flake now reddens the job rather than passing silently. That is the intent of a gate, but it is a visible behavior change on three lanes.

[MOD-18184]: https://redislabs.atlassian.net/browse/MOD-18184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MOD-17912]: https://redislabs.atlassian.net/browse/MOD-17912?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Benchmark jobs can now fail merges and nightlies on KPI breaches or EC2 flakes; nightly Slack will report benchmark failures, and the known MOD-18654 benchmark may mask regression signal in job status alone.
> 
> **Overview**
> **Enforces committed benchmark KPI floors in CI** by letting `redisbench-admin run-remote` fail the workflow when throughput drops below each test’s `kpis` floor, instead of swallowing errors with `continue-on-error`.
> 
> In **`benchmark-flow.yml`**, remote runs now fail the job on KPI breach; benchmark result JSON is uploaded as artifacts (for baseline bumps); and the PR performance comment runs with `always()` so it still posts when the run step fails. **`benchmark-runner.yml`** sets matrix **`fail-fast: false`** so one shard’s failure does not cancel the rest of the suite.
> 
> **Nightly** (`event-nightly.yml`) adds a **`benchmark`** job (skipped on push to avoid duplicate EC2 runs) and includes benchmark failures in the Slack **`test-summary`** `job-failures` string.
> 
> Adds **`tests/benchmarks/update_kpis.py`** to raise YAML floors from result JSON (`measured × (1 − margin)`, never lowering) and replaces **`Readme.md`** with **`README.md`** documenting baselines, promotion flow, and that **`json_nummultby_num_2`** (MOD-18654) keeps the job red until fixed—regressions vs that error must be read from logs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c669f889dfac6a4fb7f9fa19eba90e1f79cc301c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/benchmarks/README.md
```diff
@@ -17,3 +17,60 @@ pip3 install redisbench_admin>=0.1.74
 
 Each benchmark requires a benchmark definition yaml file to present on the current directory. The benchmark spec file is fully explained on the following link: https://github.com/RedisLabsModules/redisbench-admin/tree/master/docs
 
+## Performance baselines
```
AvivDavid23, line 20: nitpick: change file name to README.md

.github/workflows/benchmark-flow.yml
```diff
@@ -112,8 +123,25 @@ jobs:
...
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          # triggering_env distinguishes the profiler matrix, which reuses the
+          # same member ids in the same workflow run -- upload-artifact@v4
+          # requires unique names.
+          name: benchmark-results-${{ inputs.triggering_env }}-${{ inputs.benchmark_runner_group_member_id }}
+          path: tests/benchmarks/*.json
+          if-no-files-found: ignore
+          retention-days: 30
```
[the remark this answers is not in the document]
AvivDavid23: change to v6

tests/benchmarks/README.md
LiranAbir: Done in c669f88 — renamed with `git mv` so history follows, and updated the reference to it in `event-nightly.yml`.

AvivDavid23 (approved):

PR #1640, opened 2026-09-15
MOD-17687 Support Auto-Create missing paths

AvivDavid23:
Adds opt-in creation of missing object paths (`json-auto-create-deep-paths yes`, disabled by default).

  - Supported: `JSON.SET`, `JSON.MSET`, `JSON.MERGE`.
  - Also supported: `JSON.ARRAPPEND`, `JSON.ARRINSERT` (index `0` only),
  `JSON.STRAPPEND`, `JSON.NUMINCRBY`.
  - Prepares missing subtrees before attachment using shared logic.
  - Enforces depth limits; tests cover overlapping paths, failures, and replication.
  - Example when enabled: `JSON.SET k $.a.b 5` on `{}` produces `{"a":{"b":5}}`.

---

> [!NOTE]
> **High Risk**
> Changes core JSON mutation, replication, and MSET partial-success semantics; cluster nodes must use the same immutable config or replicas can diverge.
> 
> **Overview**
> Adds an **opt-in, load-time** module setting `json-auto-create-deep-paths` (default off) so JSON write commands can materialize **missing object key chains** along a path instead of returning nil or failing when intermediates are absent.
> 
> **Write behavior:** `JSON.SET`, `JSON.MERGE`, and `JSON.MSET` plan detached subtrees (resolve prefix matches once, group overlaps, enforce `MAX_DEPTH`), attach via `dict_add`, then apply updates. Missing keys can be bootstrapped from a plain object-key path. Seeded commands create leaves then run unchanged: `JSON.NUMINCRBY` (`0`), `JSON.STRAPPEND` (`""`), `JSON.ARRAPPEND` / `JSON.ARRINSERT` at index `0` (`[]`). Array indexes and wildcards in the creatable suffix are not invented; disabled mode keeps legacy static-path / final-key rules.
> 
> **json_path support:** `Query::pop_last_object_key` peels trailing plain object segments; `visit_once_with_paths` streams matches in query order for preparation without a full result vector.
> 
> **Supporting changes:** `Manager` gains `clone_value` / `create_object`; shared `number_op_result` for increment validation; coverage CI pins Redis via `get-redis-ref.sh` and treats test failures as fatal in `make coverage`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 83ca885a5957cccabfca7f42097cf3cd178348d7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


json_path/src/json_path.rs
```diff
@@ -146,7 +146,7 @@ pub enum JsonPathToken {
 }
 
 /* Struct that represent a compiled json path query. */
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
TalBarYakar, line 149: Could we document why Query now needs Clone

json_path/src/json_path.rs
```diff
@@ -2694,6 +2729,109 @@ mod json_path_compiler_tests {
...
+    #[test]
+    fn test_pop_last_object_key_accepts_dot_field() {
+        let mut query = compile("$.foo").unwrap();
+        assert_eq!(query.pop_last_object_key(), Some("foo".to_string()));
+        assert_eq!(query.pop_last_object_key(), None);
+    }
+
+    #[test]
+    fn test_pop_last_object_key_accepts_bracket_string() {
+        for path in [r#"$["foo"]"#, "$['foo']"] {
```
TalBarYakar, line 2741: could we extend this test to cover $"" and $"0"? Both are valid object keys , no ?

TalBarYakar:
I think we should consider the recursive version, where we validate the path while traversing forward, reach the last element, create it, and link everything back into the existing path as the recursion unwinds.
I believe this would allow us to remove all the planning functions and make the code less complex. As mentioned, a single traversal could handle creation, planning, and validation together.
Aviv, could you look into this approach?

AvivDavid23:
> I think we should consider the recursive version, where we validate the path while traversing forward, reach the last element, create it, and link everything back into the existing path as the recursion unwinds. I believe this would allow us to remove all the planning functions and make the code less complex. As mentioned, a single traversal could handle creation, planning, and validation together. Aviv, could you look into this approach?

good idea, check now

